PR #7, opened 2017-05-11
Implements SPI lookup for execution planners

bartoszmajsak:


bartoszmajsak:
> LGTM Let's see how this https://github.com/arquillian/smart-testing/pull/7/files#diff-bade6fa86288fa76e28a4353b873db1dR9 evolve in the future but for now it is ok.

The original idea I discussed with @matous was to pass a map of some sort to make it flexible. But for the moment I don't really see the use case yet, so I concluded it would be too early to speculate. If we see there is a need for more flexible way of creating instances of test execution planners we will definitely refine this.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
```diff
@@ -44,32 +39,14 @@ public TestsToRun apply(List<String> orderStrategy) {
...
-            final Set<File> mainClasses =
-                new NewFilesDetector(projectDir, previousCommit, commit, "**/src/main/java/**/*.java").getFiles();
-
-            mainClasses.addAll(
-                new ChangedFilesDetector(projectDir, previousCommit, commit, "**/src/main/java/**/*.java").getFiles());
-
-            return new AffectedChangesDetector(new FileSystemTestClassDetector(projectDir), mainClasses);
-        } else if(Objects.equals(orderStrategy, "failed")) {
-            return new FailedTestsDetector();
+        for (final TestExecutionPlannerFactory testExecutionPlanner : ServiceLoader.load(TestExecutionPlannerFactory.class)) {
```
MatousJobanek: the code `ServiceLoader.load(TestExecutionPlannerFactory.class)` will be called for every strategy:
https://github.com/arquillian/smart-testing/pull/7/files#diff-356cfd962197d56c7e69c381ba349281R26
but the list of the implementations won't change during the loop. I'm not sure how big is the load (if there is any), but I believe that it would be good to get the list of implementations before the loop and use it as a local variable.

surefire-provider/pom.xml
```diff
@@ -27,6 +27,7 @@
     <dependency>
       <groupId>org.arquillian.smart.testing</groupId>
       <artifactId>smart-testing-core</artifactId>
+      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>org.arquillian.smart.testing</groupId>
```
MatousJobanek: I think that the following strategy dependencies are not necessary now - as it should be completely isolated. 
They should be moved to the functional tests - to have them on classpath. It would be good to have some depchain not to have 4 dependencies but only one...

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
MatousJobanek: One more thing - what if the required planner is not found - an exception should be thrown or at least some warning printed - this should contain the list of strategies that are available...

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
```diff
@@ -44,32 +39,14 @@ public TestsToRun apply(List<String> orderStrategy) {
...
+            if (testExecutionPlanner.isFor(orderStrategy)) {
+                return testExecutionPlanner.create(projectDir, globPatternsAsArray);
```
bartoszmajsak: I am not sure if that should be there - my thinking is that the "client" (in this case surefire plugin) will decide when and how to instantiate it. At this point, we have a bit of coupling here too. Let's open an issue or document this concern in the wiki so that we don't forget to come back it.

surefire-provider/pom.xml
bartoszmajsak: > They should be moved to the functional tests - to have them on classpath. It would be good to have some depchain not to have 4 dependencies but only one...

It's not even about having 1 dep instead of X, but that we can offer all strategies to the users at once without them asking  which dependencies to pull. Good spot  @MatousJobanek, can you create issue here?

core/src/main/java/org/arquillian/smart/testing/spi/TestExecutionPlannerFactory.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.io.File;
+
+public interface TestExecutionPlannerFactory {
+
+    String alias();
+
+    boolean isFor(String name);
+
+    TestExecutionPlanner create(File projectDir, String[] globPatterns);
```
bartoszmajsak: It's not optimal that we pass `globPatterns` around, but there is potentially bigger improvement needed in this area, so let's do refactor it as a part of #8

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestExecutionPlannerLoader.java
```diff
@@ -0,0 +1,42 @@
...
+        if (availableStrategies.isEmpty()) {
+            loadStrategies();
+        }
+
+        if (availableStrategies.containsKey(strategy)) {
+            return availableStrategies.get(strategy);
+        }
+
+        throw new IllegalArgumentException("No strategy found for [" + strategy
+            + "]. Please make sure you have corresponding dependency defined.");
```
MatousJobanek: How about printing the list of available strategies? For this purpose, we could use the aliases stored as keys in the map

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestExecutionPlannerLoader.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    private void loadStrategies() {
+        for (final TestExecutionPlannerFactory testExecutionPlannerFactory : ServiceLoader.load(TestExecutionPlannerFactory.class)) {
+            final File projectDir = new File(System.getProperty("user.dir"));
+            final TestExecutionPlanner testExecutionPlanner =
+                testExecutionPlannerFactory.create(projectDir, this.globPatterns);
```
MatousJobanek: Do you really want to create the instance of the planner? You cannot know if there is some logic in the factory method/constructor already that could be an unnecessary load. Instead of this, store only the class and create the instance lazily.

bartoszmajsak: Not sure if we need it yet. Can we live with it for the time being and reiterate soon (see also #8)

bartoszmajsak: Also storing the class would imply involving reflection later on to actually instantiate factory itself. I would prefer avoiding reflection when it's not really needed.

MatousJobanek: How about this solution: https://gist.github.com/MatousJobanek/650a4bf6126dfdb9d84e8879910f2ea4

bartoszmajsak: I think we are overdoing it. If you look at initial for-loop it was doing exactly what we are aiming for - loading and instantiating strategy only when it's needed. Except of the question how expensive is `ServiceLoader.load` call. Maybe we simply roll back to the original solution instead?

MatousJobanek: in the original solution, I have the comment that we shouldn't look for all available implementations of the factory method for every strategy defined. which means, for example, 4 times when I have "new,changed,failed,affected". 
This was fixed by having the HashMap, but here I don't like that you create all available instances of the strategy planner immediately. Which means that when I use only "new" all planners will be created.
My proposal is a combination of both - having HashMap and creating only when necessary.

MatousJobanek: It's just an optimization detail. Choose the solution you like...

bartoszmajsak: > I have the comment that we shouldn't look for all available implementations of the factory method for every strategy defined. which means, for example, 4 times when I have "new,changed,failed,affected".

That's what I mean by saying "Except of the question how expensive is ServiceLoader.load call." At the end of the day these are very simple objects. 

> It's just an optimization detail

Detail or not, maybe premature?

> My proposal is a combination of both - having HashMap and creating only when necessary.

Sure, but we trade 3 lines of code for extra complexity which I don't know we really benefit from. Do we know how faster this is comparing to original for-loop?

MatousJobanek: I'm giving this conversation up.
My comments just represent how I would write the code. I don't compare a time of execution for every part of the code to have the statistics to choose the best solution... 
Take any of them. It's just a matter of point of view.

bartoszmajsak: > I'm giving this conversation up.

I think you confused discussion with defending the code. I'm not defending mine. In fact I don't like it at all. I just wanted to discuss it and understand the point of view.

bartoszmajsak: Thanks for all the suggestions. As we want to give user also information about available strategies I pushed similar approach with a map as you suggested. I will write some tests and merge afterward.

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/TestExecutionPlannerLoaderTest.java
```diff
@@ -0,0 +1,84 @@
...
+    public void should_find_matching_strategy() throws Exception {
+        // given
+        final JavaSPILoader mockedSpiLoader = mock(JavaSPILoader.class);
+        when(mockedSpiLoader.load(eq(TestExecutionPlannerFactory.class))).thenAnswer(i -> Collections.singletonList(
+            new DummyTestExecutionPlannerFactory()));
+        final TestExecutionPlannerLoader testExecutionPlannerLoader =
+            new TestExecutionPlannerLoader(mockedSpiLoader, new String[] {});
+
+        // when
+        final TestExecutionPlanner exceptionShouldBeThrown = testExecutionPlannerLoader.getPlannerForStrategy("dummy");
```
MatousJobanek: why the name of the variable is `exceptionShouldBeThrown` when there is no exception expected? 
Isn't there missing the `// then` part?

bartoszmajsak: Isn't it obvious? I'm following best practices such as Copy-Paste Design Pattern! Will fix it soon. Thank you

Issue #11, opened 2017-05-19
Ability to fetch all git changes at once


lordofthejars:
Currently when fetching git changes for affected strategy we have to call both new and changed git detections. This results in calling the same git logic twice. We should introduce lower level component letting us get this information all at once.

lordofthejars:
In partial solution we have added a composition of both strategies into one _util_ class: https://github.com/arquillian/smart-testing/blob/master/strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/AllChangedFilesDetector.java Now we still have the problem that git repository is opened twice, but from the client perspective it is only one call. We need to profile that and decide if we want a full refactor of the solution or we can leave it as is.

bartoszmajsak:
Related to #44

lordofthejars:
I think this can be closed as maven extension work is on master. cc/ @bartoszmajsak

bartoszmajsak:
Fixed with refactorings in #57

PR #12, opened 2017-05-19
feat(#11): Adds class that get all changes from git commits

lordofthejars:


strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/AllChanges.java
```diff
@@ -0,0 +1,39 @@
+package org.arquillian.smart.testing.vcs.git;
+
+import java.io.File;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * The intention of this class is not to provide a final strategy to final user but to be used by other strategies that requires all changes of Git commits.
+ */
+public class AllChanges {
```
lordofthejars: Don't like so much the name

strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/AllChanges.java
```diff
@@ -0,0 +1,39 @@
...
+    public AllChanges(File currentDir, String previous, String head, String ... globPatterns) {
+        newFilesDetector = new NewFilesDetector(currentDir, previous, head, globPatterns);
+        changedFilesDetector = new ChangedFilesDetector(currentDir, previous, head, globPatterns);
+    }
+
+    public Set<String> getTests() {
+        final Set<String> tests = new HashSet<>();
+
+        tests.addAll(newFilesDetector.getTests());
+        tests.addAll(changedFilesDetector.getTests());
```
bartoszmajsak: This way we only hide the implementation of calling `new` and `changed`, but not really fixing the problem though, as we still call `gitChangeResolver.diff` underneath twice, right? I think the idea here was to avoid this?

strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/AllChanges.java
bartoszmajsak: Me neither, but then maybe we call it `AllChangedFilesDetector`... humm....

strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/AllChanges.java
lordofthejars: Yes you are right, the problem is that then we get some duplication of code on child classes and moving some classes from parent to child. So for now and since this should be called only once, I decided to do it easy for now (also easy to understand) and when we have this integrated with real projects, then profile it and see if we are loosing time there or not.

bartoszmajsak: But then it's fine to have the same code as originally, this encapsulation just adds a layer which I'm not sure we benefit from.

bartoszmajsak: I thought the main driver to work on this issue was to not do the repo and I/O work twice.

lordofthejars: Yes this was also mine but I notice that if we do this refactor maybe the code becomes less clear, so we can do two things the first one accept it and if we need to do in the future we have the class, or just drop the PR or do a full refactor which I am not totally sure now.

bartoszmajsak: I would suggest to let those changes in but also create new issue.

lordofthejars: Ok, then I refactor the class and squash when you accept :)

Issue #13, opened 2017-05-22
Ordering tests vs Run only important tests


lordofthejars:
Currently only tests that matches some criteria (based on strategies) are run. Let's add a new configuration parameter that allows you to choose if you want this mode (so not all tests are executed) or just want to execute all tests but ordered in terms of priority (first the ones that matches criteria (based on strategies) and then the rest.

PR #14, opened 2017-05-22
feat(#13): Adds possibility to configure order vs smart execution

lordofthejars:


surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplierTest.java
```diff
@@ -0,0 +1,88 @@
...
+        strategyTests.add(TestExecutionPlannerLoaderTest.class.getName());
+
+        when(testExecutionPlanner.getTests()).thenReturn(strategyTests);
+
+        // when
+
+        TestStrategyApplier testStrategyApplier = new TestStrategyApplier(testsToRun, providerParametersParser, testExecutionPlannerLoader);
+        final TestsToRun realTestPlanning = testStrategyApplier.apply(Arrays.asList("static"));
+
+        Assertions.assertThat(realTestPlanning.getLocatedClasses())
```
bartoszmajsak: static import + `// then` part?

functional-tests/affected-ftest/pom.xml
```diff
@@ -42,6 +42,9 @@
           <properties>
             <orderStrategy>affected</orderStrategy>
           </properties>
+          <includes>
+            <include>**/src/test/java/**/Test*.java</include>
```
bartoszmajsak: Let's make it more generic, e.g. `*Test*.java`, but then we should avoid having `TestUtils` or something similar ;)

lordofthejars: Changed to two includes to follow surefire notation and only allows `Test` at start or at end but not in the middle.

lordofthejars:
By default smart testing will just use strategies configured using `orderStrategy` to select which tests are the ones to be executed ignoring the other ones. For example if you choose `failed` strategy only tests that failed in previous run will be executed, and the other tests will be just ignored.

This might be the desired behaviour in some circumstances but you might also want to not ignore the other tests but just run them at last.

For this purpose a new attribute has been created called `order` that when it is set to true then all tests are executed but instead of using default surefire ordering, smart testing will first run the ones detected by the defined strategies, and finally the rest of them.

bartoszmajsak:
Thanks for explaining @lordofthejars. We can use it for documents. However I would suggest following, instead of `<order>true</order>` we maybe can have `<executeAllTests>`. I believe it reveals meaning a bit better. Also can we have its default value to `true`?

lordofthejars:
+1 to change of name, so it is an attribute without value, if present then it is enabled. What I am not sure is which should be the default

bartoszmajsak:
> What I am not sure is which should be the default

My thinking here is - I would expect surefire to order tests, but execute them all anyway. Otherwise we would call it "filtering strategy" I believe. I think it's a good convention. For multiple build steps we anyway will have to fine tune behaviour so this is small thing to enable/disable if needed.

@MatousJobanek WDYT?

lordofthejars:
So if default is execute all tests, then maybe the attribute should be called `filtering`

bartoszmajsak:
> So if default is execute all tests, then maybe the attribute should be called filtering

that's also a possibility :) let's agree first on default behaviour and then based on it we decide how to call the flag.

lordofthejars:
My point is that maybe as a developer what I want to do is `mvn test` and not run everything but just what it has been modified/affected/previously failed. Then in continuous delivery pipeline, during commit stage, you don't want to run the whole suite so I still want to relay on filtering, and it is only in acceptance stage where you run everything that you want to execute everything (so you do `mvn test -P acceptancetest`).

Saying that but if we want to focus on microservices which probably there are less tests to execute then the ordering mode could be the default one.

Opinions?

bartoszmajsak:
>  in continuous delivery pipeline, during commit stage, you don't want to run the whole suite so I still want to relay on filtering, and it is only in acceptance stage where you run everything that you want to execute everything (so you do mvn test -P acceptancetest)

Pipelines are anyway quite heavily configured - I am opting for my reasoning as convention over configuration. I typically want to run all tests and if I can schedule them to not waste time waiting for the whole build, but capture errors earlier is just awesome.

lordofthejars:
Sure I don't strong opinion on any of both modes

dipak-pawar:
I would go with default behavior as run all tests with provided `orderStrategy`, Call flag as `filtering`

hemanik:
I have a opinion similar to Alex's here. If I make a minor change in just a single file and want to test it, I probably wouldn't want to spend time running all the tests I did not even touch. 
Probably, I could for instance want to check just my file for linting errors. 
So, by default I would probably like to just run the changed/modified tests and at the last step execute all the tests at the acceptance stage with `<executeAllTests>` just to be sure everything works.

bartoszmajsak:
So we have 2:2 :) I think it's all on @MatousJobanek now :D

bartoszmajsak:
My reasoning here is following the principle of **least surprise** - if I run `mvn clean test` it will behave without surprises (convention over configuration) but better.

lordofthejars:
> without surprises

We can always use colour asciiart so user knows that we have took the control of execution :)

lordofthejars:
Ok the attribute is going to be called `filtering` and by default its value is false

functional-tests/affected-ftest/pom.xml
```diff
@@ -42,6 +42,10 @@
           <properties>
             <orderStrategy>affected</orderStrategy>
```
MatousJobanek: if we introduce two approaches how to apply the strategies:
* ordering
* filtering 
then the name `orderStrategy` can be quite confusing for the filter approach... I would change it to just `strategy` or `strategies`

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
```diff
@@ -9,15 +9,37 @@
 
 class TestStrategyApplier {
 
+    static final String FILTERING = "filtering";
```
MatousJobanek: Have you considered `selecting`? In most of the documents and studies related to "smart testing", there is mainly used the word "selection" in the context of running some subset of test classes.

functional-tests/affected-ftest/pom.xml
lordofthejars: +1 to this change

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
lordofthejars: cool, then let's change to this.

lordofthejars:
Then you mean adding `ordering` property two, and in case of being both then `ordering` takes precedence.

MatousJobanek:
> Then you mean adding ordering property two,

yes, either `ordering` or `prioritizing`. I don't have a strong opinion on which one is the better.

> and in case of being both then ordering takes precedence.

I don't think that there will be a case where would be both of them in one pom.xml file. I was more likely thinking about overriding it in a submodule. Then I have in parental pom `ordering` and in a submodule in need to change it to `selecting` then having `ordering` as the default one wouldn't help.

MatousJobanek:
I see your point now. as both of them are standalone elements, it won't be overridden in the resulting effective pom, right? It would be probably worth trying if it is possible somehow override the config or if it is even merged (the surefire config itself with the one from parental pom) etc. Or have you already tried it?

lordofthejars:
Maybe what I would do is calling it `selecting` which takes a boolean. If false (default) then is ordering if true then selecting, so we avoid having more configuration attributes.

MatousJobanek:
I tried it and it works as I had expected - if we use the standalone elements then both of them are available and there is no simple way to figure out which one is used in the "nearest" pom file. 
When we use the same element, then the value is overridden...
So I would vote for having one element with several possible values. But instead of using true/false values, I would use for example an element `usage` with corresponding values. Using this approach, we won't be limited only to two options, but we can create some new usages...

lordofthejars:
I thought the same, my only concern is if we are not anticipating the future :) But let's move as you suggest, at the end it doesn't hurt doing in this way.

PR #18, opened 2017-05-24
Adds support for git.lastcommits system property.

### LINKED ISSUE #17: Add support to provide option `-Dgit.lastcommits=no_of_commits` to run tests.
Currently we have to run tests using `mvn -Dgit.previous.commit=$PREV_COMMIT -Dgit.commit=$COMMIT clean test`

along with this add support to run tests using `mvn -Dgit.lastcommits=3 clean test`

So calculate all required commit using current HEAD with number of commits given in `git.lastcommits` property from HEAD.

dipak-pawar:
Fixes #17

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/GitRunnerProperties.java
+++ b/core/src/main/java/org/arquillian/smart/testing/GitRunnerProperties.java
@@ -1,0 +1,15 @@
+package org.arquillian.smart.testing;
+
+public class GitRunnerProperties {
+
+    public static final String PREVIOUS_COMMIT = "git.previous.commit";
+    public static final String COMMIT = "git.commit";
+    static final String LAST_COMMITS = "git.last.commits";
+    public static final String HEAD = "HEAD";
+
+
+    public static String getPrevCommitDefaultValue() {
+        return HEAD + "~" + System.getProperty(LAST_COMMITS, "0");
+    }
+}
+
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -20,6 +20,10 @@
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
+      <artifactId>system-rules</artifactId>
+    </dependency>
   </dependencies>
 
 </project>
--- a/core/src/test/java/org/arquillian/smart/testing/GitRunnerPropertiesTest.java
+++ b/core/src/test/java/org/arquillian/smart/testing/GitRunnerPropertiesTest.java
@@ -1,0 +1,60 @@
+package org.arquillian.smart.testing;
+
+import java.io.IOException;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.RestoreSystemProperties;
+
+import static org.arquillian.smart.testing.GitRunnerProperties.COMMIT;
+import static org.arquillian.smart.testing.GitRunnerProperties.HEAD;
+import static org.arquillian.smart.testing.GitRunnerProperties.LAST_COMMITS;
+import static org.arquillian.smart.testing.GitRunnerProperties.PREVIOUS_COMMIT;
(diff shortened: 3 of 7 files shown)


core/src/main/java/org/arquillian/smart/testing/SystemPropertyResolver.java
```diff
@@ -0,0 +1,33 @@
+package org.arquillian.smart.testing;
+
+public class SystemPropertyResolver {
```
lordofthejars: It is called SystemPropertyResolver but it just resolves git properties, so maybe another name such as Git `SystemPropertiesResolver` ? It is just an idea I am really bad giving names to things.

dipak-pawar: yeah currently it's resolving only git properties, but in future it might be possible we can add more system properties which is not for git. So we can easily add here. instead of creating different class for System property using category it's good to have everything at one place.

bartoszmajsak: I'm not sure if wrapping it all up in the Java Bean makes that much sense. I understand that idea here was to encapsulate property names. The challenge I see here is that it will grow quite a bit but also by having one central place will force it to end up in `core` as it will be needed by all modules. Maybe instead of having this class, we can just keep git-related property keys in place and still use good old `System.getProperty`? We are anyway using system properties rule for tests so the argument of improving testability with having this delegate lookup class does not really hold here.

dipak-pawar: If we are using `System.getProperty("foo.bar")` at multiple places. So instead of using it, isn't it a good idea to have it at one place & use it from there to follow DRY, and for less maintenance.

If we are using at only one place then we can use `System.getProperty`.

bartoszmajsak: `System.getProperty(GitRunnerProperties.FOO_BAR)` is effectively the same as `SystemPropertyResolver.getGitFooBar`, but latter has disadvantages which I shared above.

core/src/main/java/org/arquillian/smart/testing/SystemPropertyResolver.java
```diff
@@ -0,0 +1,33 @@
...
+    public String getPreviousCommit() {
+        return previousCommit;
+    }
+
+    public String getCommit() {
+        return commit;
+    }
+
+    private String getPrevDefaultValue() {
+        String noOfCommits = System.getProperty("git.lastcommits");
```
bartoszmajsak: You can also set it as default "0" then `HEAD~0` will work the same way, except of being in the detached state.

bartoszmajsak: And I would still suggest to consider an alternative solution I suggested in the other comment.

core/src/main/java/org/arquillian/smart/testing/GitRunnerProperties.java
```diff
@@ -0,0 +1,17 @@
+package org.arquillian.smart.testing;
+
+public class GitRunnerProperties {
+
+    public static final String PREVIOUS_COMMIT = "git.previous.commit";
```
bartoszmajsak: Can we follow the convention like in `PREVIOUS_COMMIT` here (`git.last.commits`)?

core/src/main/java/org/arquillian/smart/testing/GitRunnerProperties.java
```diff
@@ -0,0 +1,17 @@
...
+    public static final String COMMIT = "git.commit";
+    static final String LAST_COMMITS = "git.lastcommits";
+    public static final String HEAD = "HEAD";
+
+
+    public static String getPrevCommitDefaultValue() {
+        String noOfCommits = System.getProperty(LAST_COMMITS, "0");
+
+        return HEAD + "~" + noOfCommits;
```
bartoszmajsak: I think we can inline it.

PR #20, opened 2017-05-24
Now excluding classes from src/main/ which contains test* 

### LINKED ISSUE #19: Classes under src/main with name contains `test` in not excluding due to which tests are failing with exception.
We are filtering our glassed with globePattern which contains following matchingPattern
* `**/Test*.java`
* `**/*Test.java`
* `**/*TestCase.java`
 * `**/*$*`

If user has some class inside src/main/java/ `org/arquillian/smart/testing/surefire/provider/TestExecutionPlannerLoader.java` which is not test class, our smart testing should filter it, as it's not a test class.

But currently it is not filtering it & we are getting exception as `initializationError  java.lang.Exception: No runnable methods`.

dipak-pawar:
Fixes #19

### DIFF
--- a/strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/GitChangesDetector.java
+++ b/strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/GitChangesDetector.java
@@ -98,6 +98,10 @@ abstract class GitChangesDetector implements TestExecutionPlanner {
         return false;
     }
 
+    boolean matchMainClassesPattern(String path) {
+        return matchPattern(path, "**/src/main/java/**/*.java");
+    }
+
     private boolean matchPattern(String path, String pattern) {
         final PathMatcher pathMatcher = FileSystems.getDefault().getPathMatcher("glob:" + pattern);
         return pathMatcher.matches(Paths.get(path));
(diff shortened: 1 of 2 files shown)


strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/GitChangesDetector.java
```diff
@@ -98,6 +98,10 @@ boolean matchPatterns(String path) {
         return false;
     }
 
+    boolean matchMainClassesPattern(String path) {
+        return matchPattern(path, "**/src/main/java/**/*.java");
```
bartoszmajsak: This way we lock it for maven defaults, not sure if this is what we want here.

dipak-pawar: then I think it would be good to explicitly add it in `<includes>` of surefire configuration, Do we have any other way to handle it for other tools as well?

dipak-pawar: I think, we should improve this in future - https://github.com/arquillian/smart-testing/blob/master/strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedChangesDetectorFactory.java#L31

bartoszmajsak:
Is this PR relevant or should we take a closer look on how to deal with glob patterns? Asking because I am not sure if we should spend time on dealing with the conflicts

dipak-pawar:
We should  take a closer look on how to deal with glob patterns, Hence closing this.

Issue #8, opened 2017-05-12
Include / exclude refactoring


bartoszmajsak:
We have to review and refactor usage of `globPatterns` - it's strange that we keep both exclusions and inclusions in [one place](https://github.com/arquillian/smart-testing/blob/6bac712e6822898a8cca6ca76e2ea567bc115496/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java#L78), but also that we pass `globPatterns` around when creating / loading test execution strategies.

MatousJobanek:
Instead of using include/exclude, the Planner implementations could take a list of test classes that should be run. This list [is retrieved ](https://github.com/arquillian/smart-testing/blob/master/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java#L29) from an original Surefire provider, so it takes into account also the includes/excludes.
Currently, this list is used only for [checking](https://github.com/arquillian/smart-testing/blob/master/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java#L81) whether all returned test classes are intended to run. 
As a benefit would be that the planner can limit its logic only for specified test classes - not for all files that match the regex.

bartoszmajsak:
Closed through several refactoring rounds #109

PR #22, opened 2017-05-25
Validating ClassPath.

### LINKED ISSUE #21: Test should run successfully, if we are running tests from one module, there are some changed files in another module
If we have multi-module project 
- module A
- module B
maven surefire is configured to module B as follows:
```xml
<plugin>
        <version>${version.surefire}</version>
        <configuration>
          <properties>
            <orderStrategy>new,changed</orderStrategy>
          </properties>
        </configuration>
        <dependencies>
          <dependency>
            <artifactId>smart-testing-surefire-provider</artifactId>
            <version>${project.version}</version>
          </dependency>
        </dependencies>
      </plugin>
```

If I ran tests from module B using `mvn -Dgit.previous.commit=$PREV_COMMIT -Dgit.commit=$COMMIT clean test -pl module-B` 

Now after changing some tests from module A, If I try to run tests using above command again. Our smart-testing-tool should not worry about module-A change.

Currently getting exception ClassNotFound which is right as this class is from different module. But out tests should continue instead of throwing exception. You can look code [here](https://github.com/arquillian/smart-testing/blob/master/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java#L28)

dipak-pawar:
* Filtering classes beyond scope of current context. Fixes #21

### DIFF
--- a/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
+++ b/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
@@ -4,6 +4,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.apache.maven.surefire.providerapi.ProviderParameters;
 import org.apache.maven.surefire.util.TestsToRun;
 import org.arquillian.smart.testing.spi.TestExecutionPlanner;
 
@@ -14,11 +15,14 @@ class TestStrategyApplier {
     private final TestExecutionPlannerLoader testExecutionPlannerLoader;
     private TestsToRun testsToRun;
     private ProviderParametersParser paramsProvider;
+    private final ClassLoader testClassLoader;
 
-    TestStrategyApplier(TestsToRun testsToRun, ProviderParametersParser paramsProvider, TestExecutionPlannerLoader testExecutionPlannerLoader) {
+    TestStrategyApplier(TestsToRun testsToRun, ProviderParametersParser paramsProvider,
+        TestExecutionPlannerLoader testExecutionPlannerLoader, ProviderParameters bootParams) {
         this.testsToRun = testsToRun;
         this.testExecutionPlannerLoader = testExecutionPlannerLoader;
         this.paramsProvider = paramsProvider;
+        testClassLoader = bootParams.getTestClassLoader();
     }
 
     TestsToRun apply(List<String> strategies) {
@@ -32,7 +36,6 @@ class TestStrategyApplier {
 
             return new TestsToRun(orderedTests);
         }
-
     }
 
     private boolean isSelectingMode() {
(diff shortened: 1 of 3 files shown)


bartoszmajsak:
> Validating TestClass for @Test using Junit4Checker, Safe from all classes which doesn't have @Test method which contains Test* in there name. This includes classes from main & test scopes. Fixes #20

Does it mean that if we merge this, #20 is obsolete?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
```diff
@@ -48,15 +54,28 @@ private boolean isUsageSet() {
...
+
+    private boolean validateClassPathAndTestClass(String testClass) {
+        Class<?> aClass;
+        try {
+            aClass = testClassLoader.loadClass(testClass);
+        } catch (ClassNotFoundException e) {
+            return false;
+        }
+
+        return jUnit4TestChecker.accept(aClass);
```
bartoszmajsak: What if we are using TestNG instead?

dipak-pawar: Good point. didn't think about this. Need to figure out

bartoszmajsak: I think we should just base it on naming conventions (so includes/excludes). Does the classpath bit fixes module issue? If so maybe we can copy the code to our own class to be free from this dependency?

dipak-pawar: but from naming convention you didn't know it this class has method annotated with `@Test` we need some verification about class internals.

Yes classpath fixes module issue.

bartoszmajsak: I'm not sure we should really care about it. Notice that for surefire it doesn't matter too. It just provides list of files and the rest is taken care by underlying test runner. It will validates and reports if the class has a structure according to the framework of choice or not.

bartoszmajsak: We just manipulate the order, the rest we should just delegate (in principle).

dipak-pawar: We should really care about it.
e.g. 
We have FirstTest, SecondTest already committed in last commit. We have noncommitted file called `TestPlan` in your working. Consider strategy as `new,changed`

In this case we have FirstTest, secondTest given by surefireProvider, but we have new file called TestPlan which we are adding it to list from respective detector later. Look here https://github.com/arquillian/smart-testing/blob/master/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java#L46

Before adding we should look that it's test file or something different to make out tool smart.

bartoszmajsak: But this can be solved with includes/excludes as done by surefire itself, right?

bartoszmajsak: Otherwise we have to support every possible framework surefire does. And I believe this is not our responsibility as we simply should delegate it to the underlying "real test runner". We are just "test order manipulator" here.

bartoszmajsak: It all boils down to this issue: https://github.com/arquillian/smart-testing/issues/8

dipak-pawar: yeah but in this case we need to explicitly mention for surefire. In this case consider the situation you are adding multiple classes in main with name as `Test*` then we are also adding it as tests. So each time user has to add it in surefire about excludes for main classes as well.

bartoszmajsak: http://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html

bartoszmajsak: I checked in the bare project and by default surefire will ignore `Test*` from src/main. For us it is failing, but I believe the fix is not to check if a class is eg. JUnit4 test, because this should be done by the test runner we delegate to. So this means we are having an issue with the way how we instrument test runner that the check is not invoked.

bartoszmajsak: I would suggest to limit this PR to validating the module part. Using `ProviderParams` test classloader is enough for the moment. The other issue we are discussing here is much broader. I created ticket for it #23

dipak-pawar:
As discussed, limiting this PR only to validate classpath using `ProviderParameters`.

bartoszmajsak:
Upstream @13fce01d0abd0cd2eb6604bd7ee583bc2385446e, thanks!

Issue #25, opened 2017-05-29
Take versions of used dependencies from manifest files


MatousJobanek:
Instead of taking versions from parameter list it finds a corresponding manifest file and takes version from this file. 
This is the only way to figure out the used surefire-plugin version. In the parameter list is only surefire-api and this one will be always the same as we use in our provider

PR #26, opened 2017-05-29
 feat(#25) Implemented loader version extractor

MatousJobanek:
Instead of taking versions from parameter list it finds a corresponding manifest file (according to groupId and artifactId) and takes version from the path of this file.
I had also a solution that was taking versions directly from the file (manifest) but for example JUnit 4.11 doesn't contain any information about the version there. This is why I'm parsing the file path.

### DIFF
--- a/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractor.java
+++ b/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractor.java
@@ -20,13 +20,11 @@ import static java.lang.Thread.currentThread;
  */
 public class LoaderVersionExtractor {
 
-    private static final Logger logger = Logger.getLogger(LoaderVersionExtractor.class.getName());
-
     public static final MavenLibrary LIBRARY_SUREFIRE_BOOTER =
         new MavenLibrary("org.apache.maven.surefire", "surefire-booter");
     public static final MavenLibrary LIBRARY_JUNIT = new MavenLibrary("junit", "junit");
     public static final MavenLibrary LIBRARY_TEST_NG = new MavenLibrary("org.testng", "testng");
-
+    private static final Logger logger = Logger.getLogger(LoaderVersionExtractor.class.getName());
     private static Map<ClassLoader, Map<MavenLibrary, String>> loaderWithLibraryVersions = new HashMap<>();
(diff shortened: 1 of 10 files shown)


surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractor.java
```diff
@@ -0,0 +1,154 @@
...
+            return ".*"
+                + File.separator
+                + groupId.replaceAll("\\.", File.separator)
+                + File.separator
+                + artifactId
+                + File.separator;
+        }
+
+        @Override
+        public boolean equals(Object o) {
```
bartoszmajsak: Can you run autoformatting on that?

MatousJobanek: I ran and this part is still same... What don't you like?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractor.java
```diff
@@ -0,0 +1,154 @@
...
+        return getVersionFromClassLoader(LIBRARY_JUNIT, currentThread().getContextClassLoader());
+    }
+
+    public static String getTestNgVersion() {
+        return getVersionFromClassLoader(LIBRARY_TEST_NG, currentThread().getContextClassLoader());
+    }
+
+    /**
+     * In the given classloader finds manifest file on a path matching the given groupId and artifactId;
+     * when it the file is matched, then it retrieves and returns a version.
```
MatousJobanek: sure

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractor.java
bartoszmajsak: It's autogenerated from IDE and doesn't have `{` `}` for `if`s. Also  last two `if`s are rather cryptic for my taste.

bartoszmajsak: On top of that I prefer to use `instanceof` instead of direct class comparison.

Issue #23, opened 2017-05-25
Passing correct classes to underlying test runner


bartoszmajsak:
Our integration with Surefire is based on delegation to available test runners. We simply manipulate the order of test classes which the intended test runner should execute. 

The challenge in the instrumentation is that Surefire comes with several ordering strategies:

   * `alphabetical` 
   * `reversealphabetical`
   * `random`
   * `filesystem`
   * `hourly` (alphabetical on even hours, reverse alphabetical on odd hours)
   * `failedfirst`
   * `balanced`

but these are hardcoded - in other words we don’t have a way to provide our own implementation of `RunOrder` (and `RunOrderCalculator`), as these are hardcoded in the plugin code.

Our [current solution](https://github.com/arquillian/smart-testing/blob/41ade53a907ea284e8da3dee11b909be4aac6fd3/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java#L58-L67) relies on reflection and apparently does not use the underlying test runner properly. 

 ## Faulty Scenario

  * Take [`functional-tests/changed-ftest`](https://github.com/arquillian/smart-testing/tree/41ade53a907ea284e8da3dee11b909be4aac6fd3/functional-tests/changed-ftest) project. 
  * Add a POJO class matching the pattern for tests, e.g. `TestBean` with a field and constructor for it (https://gist.github.com/1947860368e0de35329cdf2323a641b3)
  * Run the build `mvn test`
  * You will see [following error](https://gist.github.com/8d34be3ae899c58da6d60e5a29d9abdf
)
  * If you disable [`smart-testing` plugin](https://github.com/arquillian/smart-testing/blob/41ade53a907ea284e8da3dee11b909be4aac6fd3/functional-tests/changed-ftest/pom.xml#L48-L53) and re-run the tests build passes, as the class is not a valid test based on its structure, not just name.

## Considerations

We should reach out to Surefire team and discuss opening up `RunOrder` logic through an SPI. This way we would be able to safely hook into the execution of the test runner and have a guarantee that we feed ordered test runs properly.

MatousJobanek:
The attached fix solves the problem. A question is if the provider shouldn't pass the list of the tests to be run to the TestExecutionPlanners. Then the planner could apply his strategy only on the limited set of test classes (not whole project) and then return a resulting set of tests classes which would be a subset of the original set. 
[...]

PR #28, opened 2017-05-29
fix(#23) Applier checks if the returned test classes should be really run

MatousJobanek:
It checks if it is in the list (`TestToRun`) that was previously retrieved when the classpath was scanned.

### DIFF
--- a/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/TestNgProviderInfo.java
+++ b/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/TestNgProviderInfo.java
@@ -12,6 +12,7 @@ public class TestNgProviderInfo implements ProviderInfo {
 
     public String getProviderClassName() {
         return "org.apache.maven.surefire.testng.TestNGProvider";
+
     }
 
     public boolean isApplicable() {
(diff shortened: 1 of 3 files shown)


surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/TestNgProviderInfo.java
```diff
@@ -12,6 +12,7 @@ public TestNgProviderInfo(ProviderParametersParser paramParser) {
 
     public String getProviderClassName() {
         return "org.apache.maven.surefire.testng.TestNGProvider";
+
```
bartoszmajsak: No need for that I guess :)

bartoszmajsak:
What is the corner case demonstrating this issue?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/TestNgProviderInfo.java
MatousJobanek: Ah, yes - I needed some change to verify the fix :-D

MatousJobanek:
I changed the `TestNgProviderInfo.java` class and the `changed-ftest` failed as it returns also this class.

PR #29, opened 2017-05-29
feat(failsafe) Added configuration for failsafe plugin into ftests

MatousJobanek:
To execute the "integration tests" use command: mvn verify
It invokes the same tests as standard surefire, but with ordering usage


functional-tests/changed-ftest/src/test/java/org/arquillian/smart/testing/ftest/ITIntegration.java
```diff
@@ -0,0 +1,11 @@
+package org.arquillian.smart.testing.ftest;
+
+import org.junit.Test;
+
+public class ITIntegration {
```
bartoszmajsak: Why are we adding it if we overwrite the defaults anyway?

MatousJobanek: I'm totally st...d/blind or my `git diff` command doesn't work correctly - this shouldn't be part of this commit of course. I'll remove it in a sec.

PR #31, opened 2017-05-30
feat(print info) Changed content&formating of info printed on the beginning of test execution

MatousJobanek:
An example of the printed info:
~~~
=============  Smart testing  ==============
===                                      ===
===  Applied strategies: [new, changed]  ===
===      Applied usage: [selecting]      ===
===                                      ===
============================================
~~~


bartoszmajsak:
Do we really need such a big "banner"?

lordofthejars:
I would make it even bigger and use some asciiart :D

MatousJobanek:
:-D I don't know, I wanted to add the "applied usage" line there and then I realized that it would be good to have also the title there...
So I could remove the empty lines:
~~~
=============  Smart testing  ==============
===  Applied strategies: [new, changed]  ===
===      Applied usage: [selecting]      ===
============================================
~~~

MatousJobanek:
Or maybe try to incorporate some gif :-)?

bartoszmajsak:
But why do we want it in the first place?

MatousJobanek:
What? information what has been set? I see it as a transparency - information that smart-testing tool is used, so the resulting test execution can be different than the "normal" one...

lordofthejars:
In fact in SoS review  they asked to see why executed tests are executed.

MatousJobanek:
> But why do we want it in the first place?

The same question you could have asked about the very first info: `Applied strategies: [new, changed]` that had been there before

bartoszmajsak:
It's already there, but as a simple `INFO`. I'm not really sure we have to scream with a banner.

> In fact in SoS review they asked to see why executed tests are executed.

I know about it, but this is entirely different thing.

bartoszmajsak:
> The same question you could have asked about the very first info: Applied strategies: [new, changed] that had been there before

I'm asking different question :)

MatousJobanek:
I wanted to make it more readable and separated... You know, to put there the information that it is related to smart-testing toll. But it is just my opinion, if there are more guys against it, we don't have to add it there...

bartoszmajsak:
The reason why I'm not convinced of it is:

  * no one is sitting idle reading the logs to see a banner
  * normally you use grep to find relevant bits (so we will find it anyway - just need to think about pattern unique enough to search for)
  * if you run tests over and over again you consume more space with the logs which are not really giving any extra info.

Just my personal feeling. But if we are excited about visual aspect of it then sure, why not - let's make asciiart too :)

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/InfoPrinter.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.logging.Logger;
+import java.util.stream.IntStream;
+
+import static java.lang.String.format;
+
+public class InfoPrinter {
+
+    private final Logger log = Logger.getLogger(InfoPrinter.class.getName());
```
bartoszmajsak: I think we can make it `static`

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/InfoPrinter.java
```diff
@@ -0,0 +1,62 @@
...
+    private boolean isSelecting;
+    private String title = "Smart testing";
+    private String surroundingCharacter = "=";
+
+    public InfoPrinter(List<String> strategies, boolean isSelecting) {
+        this.strategies = strategies;
+        this.isSelecting = isSelecting;
+    }
+
+    public void print() {
```
bartoszmajsak: Can we have better name for it?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/InfoPrinter.java
```diff
@@ -0,0 +1,62 @@
...
+        String usageLine = format("Applied usage: [%s]", isSelecting ? "selecting" : "ordering");
+        length = getLongestSize(title, strategiesLine, usageLine);
+        log.info("\n" + getStringToPrint(strategiesLine, usageLine));
+    }
+
+    private int getLongestSize(String... lines) {
+        return Arrays.stream(lines).mapToInt(line -> line.length()).max().getAsInt();
+    }
+
+    private String getStringToPrint(String... lines) {
```
bartoszmajsak: Maybe more meaningful name here too? e.g. `createInfoBanner` or sth

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/InfoPrinter.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class InfoPrinter {
+
+    private final Logger log = Logger.getLogger(InfoPrinter.class.getName());
+
+    int length;
+    private List<String> strategies;
+    private boolean isSelecting;
+    private String title = "Smart testing";
+    private String surroundingCharacter = "=";
```
bartoszmajsak: Why `title` and `surroundingCharacter` are not constants if we don't have any way of changing those?

MatousJobanek:
I removed the empty line to make it smaller and used Logger for it to keep consistency
Now the output looks like this:
~~~
May 30, 2017 4:28:24 PM org.arquillian.smart.testing.surefire.provider.InfoPrinter print
INFO: 
==========  Smart testing  ===========
===  Applied strategies: [failed]  ===
===   Applied usage: [selecting]   ===
======================================
~~~
Which means two additional lines comparing with the original state (the header and the footer).

Let's vote... now it is 1:1 
@lordofthejars @dipak-pawar @hemanik WDYT

bartoszmajsak:
Let's use reaction on the PR description (+1/-1)

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/InfoPrinter.java
MatousJobanek: `print()` is fine for me :-). I can imagine changing name of the class to have:
`new InfoBanner(strategies, isSelectingMode()).print()`

But proposals are welcome ;-)

MatousJobanek:
I changed my mind - when I came back to my computer in the morning with "fresh" mind I realized that the banner is not the best idea and is breaking the Maven logging convention. 
I did some research how logging is done in other providers  - they are basically using the standard output, so for this type of information (printed by provider) I would use the same approach:
stdout for INFO
stderr for WARN
with the fact, that it automatically adds a prefix to make it clear that it is connected to Smart-Testing tool
The resulting info then looks like:
~~~
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
INFO: Smart-Testing - Applied strategies: [failed]
INFO: Smart-Testing - Applied usage: [selecting]
Running org.arquillian.smart.testing.ftest.CalculatorTest
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.013 sec - in org.arquillian.smart.testing.ftest.CalculatorTest
~~~

Issue #36, opened 2017-06-01
Affected strategy does not find tests with different names than Test.java even if we includes pattern in surefire configuration.


dipak-pawar:
*Steps to Reproduce*
1.Take simple project which have one class 'Calculator.java' 
```java
package org.arquillian.smart.testing.ftest;
public class Calculator {
    public static int add( int a, int b) {
        return a + b;
    }
}
```

2.Add simple tests related to it by name 'CalculatorTestCase.java' 
```java
public class CalculatorTestCase {

    @Test
    public void should_add_numbers() {
        Assert.assertEquals(6, Calculator.add(4, 2));
    }
}
```
3. Add surefire Configuration in pom.xml
```xml
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <version>${version.surefire}</version>
        <configuration>
          <properties>
            <strategies>affected</strategies>
            <usage>selecting</usage>
          </properties>
          <includes>
            <include>**/*TestCase.java</include>
          </includes>
        </configuration>
        <dependencies>
          <dependency>
            <groupId>org.arquillian.smart.testing</groupId>
            <artifactId>smart-testing-surefire-provider</artifactId>
            <version>${project.version}</version>
          </dependency>
        </dependencies>
      </plugin>
```
4. Run project with `mvn test`

*Expected behavior*
It should detect & run CalculatorTestCase.java

*Actual behavior*
It's ignoring this test

**Root cause**
Look for [this](https://github.com/arquillian/smart-testing/blob/master/strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/JavaToClassLocation.java#L10) method.

Here we are only looking for class which ends with `Test.java`. We should look for pattern provided in surefire configuration & then transform it.

PR #37, opened 2017-06-01
Now detecting all affected tests for main class changes as per includes

dipak-pawar:
**Fixes**: #36


bartoszmajsak:
In general, it looks good to me. I would like to run it through some e2e tests and then merge. Thanks for this fix @dipak-pawar !

Issue #33, opened 2017-05-31
Unify logging


bartoszmajsak:
It seems that other Surefire providers use `stdout` for `INFO` and `stderr` for WARN logging (https://github.com/arquillian/smart-testing/pull/31#issuecomment-305127074). We have JUL in few places in the code already. We should unify the usage across the project based on [new logger](https://github.com/arquillian/smart-testing/blob/d606c8ccb7481e5ed810aaabe2b68265ee957995/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingProviderLogger.java) introduced in #31.

MatousJobanek:
I'm not sure if we can unify it across the whole project - the stdout and stderr logging doesn't provide us any possibility to log on a lower level (DEBUG, FINEST,..).
We have probably two possibilities:
1. Use the SmartTestingProviderLogger.java only in provider and in other modules use JUL as it is done now. 
2. extend logic of SmartTestingProviderLogger.java so in a case of INFO and WARN use stdout and stderr, for all other levels use JUL. As a result, there will be one logger class that would hide that he is using two different ways of logging. This will also require moving the SmartTestingProviderLogger.java
class to core.

I would vote for the second option. WDYT?

bartoszmajsak:
Variant 2 makes much more sense. Thanks for thinking it through. 

Two additional things I can think of here:

  * We should rename the class - maybe `Logger` is enough here?
  * What about class info when using JUL? If we move it to common class we would end up with `.getLogger(this.getClass())`. This way we are loosing an information about to source of the log.

MatousJobanek:
Yes, both of your comments is a must.
I chose the name as my first intention was to use it only in the provider. And the class info when using JUL is really essential.

bartoszmajsak:
> I chose the name as my first intention was to use it only in the provider. And the class info when using JUL is really essential.

I know :) Just pointing it out that if we move it we might want to change it.

PR #40, opened 2017-06-13
Unify Logging

hemanik:
Unify JUL and Surefire Provider Logger.

- adds a wrapper over JUL to extend logic of SmartTestingProviderLogger.java so in case of INFO and WARN use stdout and stderr, for all other levels use JUL.
- moves SmartTestingProviderLogger.java class to core as Logger.java.

**Fixes**: #33


core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -0,0 +1,93 @@
...
+     * @param msg
+     *     Message to print
+     * @param args
+     *     arguments to use for formatting the given message
+     */
+    public void warn(String msg, Object... args) {
+        System.err.println(getFormattedMsg("WARN", msg, args));
+    }
+
+    public void severe(String msg) {
```
MatousJobanek: add javadoc for this method and for all following...  please add also the possibility of specifying args to be used for formatting of the given message. In the same way, as it is done in case of `warn` and `info`

core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private java.util.logging.Logger jul;
+
+    private static String prefix = "%s: Smart-Testing - ";
+
+    private Logger(java.util.logging.Logger logger) {
+        this.jul = logger;
+    }
+
+    public static Logger getLogger(Class clazz) {
```
MatousJobanek: add javadoc to the factory methods. you could also use this shorter way of creating our Logger directly in the code below.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -163,7 +163,6 @@ private void findParents(JavaClass jclass, Set<String> changedParents) {
             if (changedParents.add(parent.getName())) {
 
                 logger.log(Level.FINEST, String.format("%s test has been added because it depends on %s", parent.getName(), jclass.getName()));
```
MatousJobanek: change it to shorter form: 
`logger.finest("%s test has been added because it depends on %s", parent.getName(), jclass.getName())`

core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -0,0 +1,93 @@
...
+     * @param   level
+     *     One of the message level identifiers, e.g., SEVERE
+     * @param   msg
+     *     The string message (or a key in the message catalog)
+     * @param   args
+     *     arguments to the message
+     */
+    public void log(Level level, String msg, Object... args) {
+        if (jul.isLoggable(level)) {
+            jul.log(level, msg, args);
```
MatousJobanek: I'm not sure if the formatting logic of `jul.log` is same as `String.format`. Please check it - it would be good to have it consistent

core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -0,0 +1,163 @@
...
+     * @return a suitable Logger
+     */
+    public static Logger getLogger(Class clazz) {
+        return getLogger(clazz.getName());
+    }
+
+    /**
+     * Find or create a logger for a named subsystem using java.util.logging.Logger.
+     * If a logger has already been created with the given name it is returned.
+     * Otherwise a new logger is created.
```
bartoszmajsak: Does this description really holds to what the code actually does? I think it is only true for the jul logger we will be delegating to?

Issue #41, opened 2017-06-26
Introduce SPI for loading test results


bartoszmajsak:
We have Surefire specific parts for finding test results in the `failed` strategy module. This should be moved to SPI and implementation should be in `surefire-provider` module.

PR #46, opened 2017-06-30
Introduce Test Result Loader SPI

lordofthejars:

Introduce Test Result Loader SPI

- Created SPI for test result parsers
- JUnit Test Result Parser implementation created

**Fixes**: #41


junit-test-result-parser/src/main/java/org/arquillian/smart/testing/parser/junit/JUnitTestResultParser.java
```diff
@@ -0,0 +1,95 @@
...
+
+                if (event.isEndElement()) {
+                    final EndElement endElementElement = event.asEndElement();
+                    if ("testcase".equalsIgnoreCase(endElementElement.getName().getLocalPart())) {
+                        testResults.add(currentTestResult);
+                    }
+                }
+            }
+        } catch (XMLStreamException e) {
+
```
bartoszmajsak: Shouldn't we do something about this exception? It means we have faulty result and we have no clue at the runtime we actually do have a problem.

junit-test-result-parser/src/main/java/org/arquillian/smart/testing/parser/junit/JUnitTestResultParser.java
```diff
@@ -0,0 +1,95 @@
...
+                            if ("time".equalsIgnoreCase(attribute.getName().toString())) {
+                                duration = attribute.getValue();
+                            }
+                        }
+
+                        final float durationInSeconds = Float.parseFloat(duration);
+                        currentTestResult = new TestResult(classname, name, durationInSeconds);
+                    }
+
+                    if ("failure".equalsIgnoreCase(startElement.getName().getLocalPart())) {
```
bartoszmajsak: I would get this `if`s out to dedicated method for better readability.

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/InProjectTestReporterLoader.java
```diff
@@ -38,11 +47,12 @@
                         }
                     })
                     .map(is -> {
-                        try {
-                            return SurefireReader.loadTestResults(is);
-                        } catch (XMLStreamException e) {
-                            return null;
+                        final Optional<TestResultParser> testResultParser = javaSPILoader.onlyOne(TestResultParser.class, trp -> "junit".equals(trp.type()));
```
bartoszmajsak: Do we really need a predicate here?

lordofthejars: Well it is an easy and clean way to not populate service logic across the code. I mean if not means to get an iterable iterate over it find if it is the type we want, throw exception if required and so on so far. One way should be just create a method that receives the type, but a predicate is not so much complex nor add complexity and allows in the future to do other kind of filters.

bartoszmajsak: I understand the idea of predicate itself, but I am not sure if we really need it here. We are pretty much locking possible parsing to `junit` only. Do we really want it?

lordofthejars: Well notice that the method comes from JavaSPI which is a generic class, then maybe not in this concrete case but in the future, so what you mean now is that I call the `onlyOne` and that's all isn't it? Yes for now it might has sense as well

bartoszmajsak: If we need that in the future we can easily implement it. For now I don't really see a use-case for that. Do you?

lordofthejars: Well searching for type is one, also for example searching for instanceof might be a use case, for example since services might have an empty constructor, you might need to inject some values with setter so you can do `.onlyOne(tpr -> tpr.getClass().instanceOf MyNewTestParser.class)` and then cast to desired object.

lordofthejars: If it was a more complex code I would agree with you but it is really simple and now it is used when filtering by `type`

bartoszmajsak: So what should we do with this?

lordofthejars: For me leave it as is, I am sure any reader understand it, and I see that in near future we will need to filter with other things such as `instanceof` thing.

bartoszmajsak: > .onlyOne(tpr -> tpr.getClass().instanceOf MyNewTestParser.class)

This would be rather odd from an API perspective. For such cases wouldn't you instantiate `MyNewTestParser` directly instead?

bartoszmajsak: But then we lock it to `junit` only which I am not sure we really want. This contradicts the idea of SPI in general - we don't benefit anything. If we are so strict, then simple `new JunitParser` would be sufficient.

lordofthejars: Well not at all, there is the `type` which can be used in the future to filter to pass it as configuration parameter which kind of report do you want to read.

strategies/failed/pom.xml
```diff
@@ -33,6 +33,11 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.arquillian.smart.testing</groupId>
+      <artifactId>junit-test-result-parser</artifactId>
+      <version>${project.version}</version>
```
bartoszmajsak: It shouldn't be here - the whole IDEA of SPI is to make it pluggable, so the "failed" startegy can work for any kind of test result format. This way we make one implementation always available on the classpath and always used.

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/InProjectTestReporterLoader.java
bartoszmajsak: Well in this case yes - we assume here that we always deal with reports in JUnit format because we filter for it using predicate. This is not pluggable and is equal to `new JUnitParser`.

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/InProjectTestReporterLoader.java
```diff
@@ -38,11 +47,12 @@
...
+                        if (!testResultParser.isPresent()) {
+                            throw new IllegalArgumentException("No JUnit Test Result Parser found in classpath");
```
bartoszmajsak: If we use SPI it doesn't necessary mean it is "JUnit Test Result Parser", or does it?

strategies/failed/pom.xml
lordofthejars: Ok then it is another dependency to add when you configure Surefire. Or better in surefire-provider? I agree with you but since now we are only offering one, maybe it is easy in this way of having predefined.

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/InProjectTestReporterLoader.java
bartoszmajsak: Or let me ask you differently - let's assume Serenity BDD has different test reports and we would like to use them to know which tests are failing. How would I parse them in the `failed` strategy with this implementation?

strategies/failed/pom.xml
bartoszmajsak: I would say if we have SPI but do not open it up so we can plug different implementations this really shouldn't be here. For the time being `surefire-provider` might be ok, but this is not the ideal solution either.

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/InProjectTestReporterLoader.java
lordofthejars: No, a new service implementation

bartoszmajsak: But this code enforces to have service implementation with `junit` type...

lordofthejars: yes this is what we discussed yesterday that we will need to provide which type you want to use as a configuration property, but since now we only have one, it is ok in this way.

strategies/failed/pom.xml
lordofthejars: Exactly. This is something to fix when we have more than one.

bartoszmajsak: Let's move it to `surefire-provider`

MatousJobanek:
I'm wondering, why do we need this parser? Don't we get enough information using RunListeners & getting the object `RunResult`? 
http://junit.org/junit4/javadoc/4.12/org/junit/runner/notification/RunListener.html
https://github.com/arquillian/smart-testing/blob/master/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java#L64

MatousJobanek:
I would prefer programmatic way than parsing XML file... Or are there any additional information?

lordofthejars:
Because this is a failing strategy which means it checks previous results to see which failed. These results are stored as XML in JUnit format. They are not the current results.

bartoszmajsak:
On top of that, we cannot simply assume we will be always sticking to JUnit and its APIs.

bartoszmajsak:
@MatousJobanek notice that strategies are unaware of Surefire.

bartoszmajsak:
To make transparent what we have discussed outside of this channel

  * This is an improvement over already existing minimal solution we pulled out for failed strategy
  * #47 will provide implementation of storing this data locally (with an SPI to store it in some other locations too)

core/src/main/java/org/arquillian/smart/testing/spi/JavaSPILoader.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+public class JavaSPILoader {
+
+    public JavaSPILoader() {
+    }
+
+    public <S> Iterable<S> all(Class<S> service) {
```
MatousJobanek: My opinion - instead of single char `S` I would use the whole word that represents the meaning of this parameter: `SERVICE` or `SERVICETYPE`
Same for the following occurences

junit-test-result-parser/src/main/java/org/arquillian/smart/testing/parser/junit/JUnitTestResultParser.java
```diff
@@ -0,0 +1,99 @@
...
+import javax.xml.stream.events.EndElement;
+import javax.xml.stream.events.StartElement;
+import javax.xml.stream.events.XMLEvent;
+import org.arquillian.smart.testing.spi.TestResultParser;
+import org.arquillian.smart.testing.spi.TestResult;
+
+public class JUnitTestResultParser implements TestResultParser {
+
+    @Override
+    public Set<TestResult> parse(InputStream surefireInputStream) {
```
MatousJobanek: The whole method is quite long for my taste. I would split it into small parts.

core/src/main/java/org/arquillian/smart/testing/spi/JavaSPILoader.java
MatousJobanek: Another thing is that for SPI methods, it would good to have some javadoc.

core/src/main/java/org/arquillian/smart/testing/spi/TestResultParser.java
```diff
@@ -0,0 +1,11 @@
+package org.arquillian.smart.testing.spi;
+
+import java.io.InputStream;
+import java.util.Set;
+
+
+public interface TestResultParser {
+
+    Set<TestResult> parse(InputStream surefireInputStream);
```
MatousJobanek: same here - missing javadoc

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/InProjectTestReporterLoader.java
```diff
@@ -6,17 +6,26 @@
...
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
-import javax.xml.stream.XMLStreamException;
-import org.arquillian.smart.testing.strategies.failed.surefire.SurefireReader;
+import org.arquillian.smart.testing.spi.JavaSPILoader;
+import org.arquillian.smart.testing.spi.TestResult;
+import org.arquillian.smart.testing.spi.TestResultParser;
 
-public class SurefireInProjectTestReporterLoader implements TestReportLoader {
+public class InProjectTestReporterLoader implements TestReportLoader {
```
MatousJobanek: Shouldn't be the name `InProjectTestReportLoader`?

core/src/main/java/org/arquillian/smart/testing/spi/TestResultParser.java
bartoszmajsak: And maybe not necessary `surefireInputStream` as a parameter name.

junit-test-result-parser/src/main/java/org/arquillian/smart/testing/parser/junit/JUnitTestResultParser.java
lordofthejars: I have split a bit, now it is just 40 lines, not sure if making it small would means less understanding in these kind of algorithms where you need to deal with states.

core/src/main/java/org/arquillian/smart/testing/spi/JavaSPILoader.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Get all Java services that implements given interface and meets the given predicate.
+     * @param service interface
+     * @param predicate to set filtering options
+     * @param <SERVICE>
+     * @return Iterable of all services implementing service interface, meeting predicate condition and present in classpath.
+     */
+    public <SERVICE> Collection<SERVICE> all(Class<SERVICE> service, Predicate<SERVICE> predicate) {
+        return StreamSupport.stream(all(service).spliterator(), false).filter(predicate).collect(Collectors.toList());
```
bartoszmajsak: Why in some cases we are returning the `Collection` and in others `Iterable`? Should we make it unified?

core/src/main/java/org/arquillian/smart/testing/spi/JavaSPILoader.java
```diff
@@ -0,0 +1,93 @@
...
+        if (all.size() == 1) {
+            return Optional.of(all.iterator().next());
+        }
+
+        if (all.size() > 1) {
+            throw new IllegalStateException(
+                "Multiple service implementations found for " + service + ". " + toClassString(all));
+        }
+
+        return Optional.empty();
```
bartoszmajsak: This code is very similar to `onlyOne(Class<SERVICE> service)` method. Can't we re-use the logic from one in the other?

core/src/main/java/org/arquillian/smart/testing/spi/TestResultParser.java
```diff
@@ -0,0 +1,25 @@
+package org.arquillian.smart.testing.spi;
+
+import java.io.InputStream;
+import java.util.Set;
+
+/**
+ * Java SPI for parsing test results stored in a stream (probably a file).
```
bartoszmajsak: `(probably a file)` - not sure how end-user would understand that part. What do we want to say here?

junit-test-result-parser/src/main/java/org/arquillian/smart/testing/parser/junit/JUnitTestResultParser.java
```diff
@@ -0,0 +1,99 @@
...
+import javax.xml.stream.events.EndElement;
+import javax.xml.stream.events.StartElement;
+import javax.xml.stream.events.XMLEvent;
+import org.arquillian.smart.testing.spi.TestResultParser;
+import org.arquillian.smart.testing.spi.TestResult;
+
+public class JUnitTestResultParser implements TestResultParser {
+
+    @Override
+    public Set<TestResult> parse(InputStream junitInputStream) {
```
bartoszmajsak: Should we close `InputStream` once done with parsing?

junit-test-result-parser/src/main/java/org/arquillian/smart/testing/parser/junit/JUnitTestResultParser.java
```diff
@@ -0,0 +1,99 @@
...
+        final Set<TestResult> testResults = new HashSet<>();
+
+        try {
+            final XMLInputFactory inputFactory = XMLInputFactory.newInstance();
+            XMLEventReader eventReader = inputFactory.createXMLEventReader(junitInputStream);
```
bartoszmajsak: Should we call `eventReader.close()`?

strategies/failed/pom.xml
bartoszmajsak: Why do we have this dependency still here?

core/src/main/java/org/arquillian/smart/testing/spi/JavaSPILoader.java
```diff
@@ -0,0 +1,93 @@
...
+
+public class JavaSPILoader {
+
+    public JavaSPILoader() {
+    }
+
+    /**
+     * Get all Java services that implements given interface.
+     * @param service interface
+     * @param <SERVICE>
```
bartoszmajsak: Are those `@param <SERVICE>` really needed in the Javadoc?

core/src/main/java/org/arquillian/smart/testing/spi/JavaSPILoader.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    /**
+     * Get only one service of given type. This method is used when you want to filter from all possible implementations of given service.
+     * If there are more than one, then {@link IllegalStateException} is thrown.
+     * @param service interface
+     * @param predicate to set filtering options
+     * @param <SERVICE>
+     * @return The service of given type and meeting predicate condition 8if any) or an exception in case of more than one.
```
bartoszmajsak: `8if` :)

junit-test-result-parser/src/main/java/org/arquillian/smart/testing/parser/junit/JUnitTestResultParser.java
lordofthejars: I don't think so, I think that in most APIs that I have used they let callers close the stream.

core/src/main/java/org/arquillian/smart/testing/spi/JavaSPILoader.java
lordofthejars: No, because in this case the onlyOne is about the `predicate` not only about the service type. So for example you might have several services of same type and then you want to filter by the predicate.

Issue #44, opened 2017-06-27
Build time overhead when running with surefire smart test provider


bartoszmajsak:
##### Issue Overview

When playing with `arquillian-core` I've noticed interesting "anomaly". It seems that by simply adding our provider significantly increases the build time (I would rather say it adds 20s rather than doubles it, but it is still a huge cost for relatively small projects).

I added "affected" and "selection" mode and here are the numbers:

#####  Regular build against master branch

```
[INFO] Total time: 27.764 s
[INFO] Finished at: 2017-06-27T19:52:54+02:00
[INFO] Final Memory: 69M/1678M
[INFO] ------------------------------------------------------------------------
arquillian-core [mvn clean install]
```

#####  Build against affected strategy (so no tests executed)

```
[INFO] Total time: 44.933 s
[INFO] Finished at: 2017-06-27T19:54:24+02:00
[INFO] Final Memory: 96M/3387M
[INFO] ------------------------------------------------------------------------
arquillian-core [mvn clean install]
```

Configuration can be found here: https://github.com/bartoszmajsak/arquillian-core/tree/smart_testing_poc

##### Additional Information

I ruled out possibility that we are in fact silently running the tests without notifying maven. For a peek into the maven execution time you can use https://github.com/jcgay/maven-profiler

My maven configuration:
```
Apache Maven 3.5.0 (ff8f5e7444045639af65f6095c62210b5713f426; 2017-04-03T21:39:06+02:00)
Maven home: /usr/bin/mvn
Java version: 1.8.0_121, vendor: Oracle Corporation
Java home: /usr/java/jdk1.8.0_121/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.11.5-100.fc24.x86_64", arch: "amd64", family: "unix"
```

bartoszmajsak:
I think it also applies to other strategies (`new` and `changed`).

bartoszmajsak:
Observartions:

  * calculating changes is invoked every time surefire is executed (that means for each and every module having tests - this leads to excessive IO and repeated git operations for example
  * we should pre-calculate this before we even start running the build
  * one of the possible approaches would be to move this logic to maven extension and implement services which will be returning particular tests for each strategy during surifere build, based on pre-calculated model

bartoszmajsak:
We are investigating usage of Maven Extension model to solve this problem

bartoszmajsak:
Closed via #57 and `ChangeStorage` implementation.

PR #48, opened 2017-07-04
[WIP]Refactors change detector.

hemanik:
 Ability to fetch all git changes at once.

- Refactoring of change detectors.

**Fixes**: #11

### DIFF
--- a/strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/AllChangedFilesDetector.java
+++ b/strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/AllChangedFilesDetector.java
@@ -19,8 +19,8 @@ public class AllChangedFilesDetector extends GitChangesDetector {
     public Collection<String> getTests() {
         final Collection<String> tests = super.getTests();
 
-        final Set<String> newFiles = this.gitChangeResolver.newChanges();
-        final Set<String> modifiedFiles = this.gitChangeResolver.modifiedChanges();
+        final Set<String> newFiles = gitChangeResolver.newChanges();
+        final Set<String> modifiedFiles = gitChangeResolver.modifiedChanges();
 
         List<String> newLocalTests = getLocalTests(newFiles);
         List<String> modifiedLocalTests = getLocalTests(modifiedFiles);
@@ -35,8 +35,13 @@ public class AllChangedFilesDetector extends GitChangesDetector {
         final Set<File> files = super.getFiles();
         final Set<String> newLocalFiles = gitChangeResolver.newChanges();
         final Set<String> modifiedLocalFiles = gitChangeResolver.modifiedChanges();
-        appendLocalFiles(files, newLocalFiles);
-        appendLocalFiles(files, modifiedLocalFiles);
+
+        Set<File> filteredNewLocalFiles = filterLocalFiles(newLocalFiles);
+        Set<File> filteredModifiedLocalFiles = filterLocalFiles(modifiedLocalFiles);
+
+        files.addAll(filteredNewLocalFiles);
+        files.addAll(filteredModifiedLocalFiles);
+
         return files;
     }
 
--- a/strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/AllChangedFilesDetector.java
+++ b/strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/AllChangedFilesDetector.java
@@ -1,39 +1,50 @@
 package org.arquillian.smart.testing.vcs.git;
 
 import java.io.File;
-import java.util.HashSet;
+import java.util.Collection;
+import java.util.List;
 import java.util.Set;
+import org.eclipse.jgit.diff.DiffEntry;
 
 /**
- * The intention of this class is not to provide a final strategy to final user but to be used by other strategies that requires all changes of Git commits.
+ * The intention of this class is not to provide a final strategy to final user but to be used by other strategies that
+ * requires all changes of Git commits.
  */
-public class AllChangedFilesDetector {
+public class AllChangedFilesDetector extends GitChangesDetector {
 
-    private NewFilesDetector newFilesDetector;
-    private ChangedFilesDetector changedFilesDetector;
-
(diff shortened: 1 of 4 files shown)


strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/GitChangesDetector.java
```diff
@@ -57,6 +57,24 @@ private File findRepoRoot(File currentDir) {
         return extractEntries(diffs, this.repoRoot);
     }
 
+    public List<String> getLocalTests(Set<String> files) {
+        final List<String> localTests = files.stream()
+            .filter(this::matchPatterns)
+            .map(file -> {
+                try {
+                    final File sourceFile = new File(repoRoot, file);
+                    return new ClassNameExtractor().extractFullyQualifiedName(sourceFile);
```
lordofthejars: I know that in the original code this was like this, but shouldn't be do `extractFullyQualifiedName`  a static method (Since it requires no context) so we shouldn't have to instantiate every time?

strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/AllChangedFilesDetector.java
```diff
@@ -1,39 +1,50 @@
...
+    public AllChangedFilesDetector(File currentDir, String previous, String head, String... globPatterns) {
+        super(currentDir, previous, head, globPatterns);
     }
 
-    public Set<String> getTests() {
-        final Set<String> tests = new HashSet<>();
+    public Collection<String> getTests() {
+        final Collection<String> tests = super.getTests();
+
+        final Set<String> newFiles = this.gitChangeResolver.newChanges();
```
lordofthejars: This is more an internal doubt, should we use `this` when the attribute is on parent class?

hemanik: I agree that it hardly makes sense to use `this` here, when we are actually making no reference to the current class. Thanks for pointing out @lordofthejars :)

strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/GitChangesDetector.java
```diff
@@ -75,7 +93,7 @@ private File findRepoRoot(File currentDir) {
             .map(diffEntry -> {
                 try {
                     final File sourceFile = new File(repoRoot, diffEntry.getNewPath());
-                    return new ClassNameExtractor().extractFullyQualifiedName(sourceFile);
+                    return ClassNameExtractor.extractFullyQualifiedName(sourceFile);
```
bartoszmajsak: Any particular reason why we made it `static`?

strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/GitChangesDetector.java
```diff
@@ -84,6 +102,15 @@ private File findRepoRoot(File currentDir) {
                 test, previous, head)).collect(Collectors.toList());
     }
 
+    public void appendLocalFiles(Set<File> files, Set<String> localFiles) {
+        files.addAll(
```
bartoszmajsak: Maybe instead of manipulating `files` set we could simply return the filtered collection and decide on the caller side what to do with the result?

bartoszmajsak:
Thanks for your work @hemanik. I will review the overall change and come back with overall feedback by the end of the day.

strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/GitChangesDetector.java
hemanik: Like @lordofthejars pointed above, `extractFullyQualifiedName` does not change across instances and as the `ClassNameExtractor` class follows `Single Responsibilty Principle`, it seems legit to make this `static` in my opinion.

bartoszmajsak: Making it static has nothing to do with SRP, and as such has nothing to do with Object Oriented programming, that said I don't really see a reason to make it static at all :) Maybe better way would be to make this component part of the owning object through composition?

bartoszmajsak: If we won't make it part of this object though making it static as the best we can have to mimic "pure function" in Java is ok, but I am still not sure we benefit a lot.

lordofthejars:
I think that before starting on this https://trello.com/c/eDIqkdOR/262-smart-testing-we-shouldnt-rely-only-on-git-for-detecting-changes we should squash this. cc/ @bartoszmajsak

strategies/changed/src/main/java/org/arquillian/smart/testing/vcs/git/GitChangesDetector.java
lordofthejars: Well my idea was just to not instantiate the object in every call, so making an attribute of the class might be an option too. But also static for me it would has sense, but for me is more related to performance of creation and destruction of the object.

bartoszmajsak: I understand, but I believe this is really minor performance issue comparing to other things we have to address first, such as #44. That said we can leave it like it is.

lordofthejars: Sure, I completely agree with you, just that when I was reviewing this, I noticed this thing.

hemanik: Thanks for the explanation @bartoszmajsak and @lordofthejars. Undoing the change.

bartoszmajsak:
That's good enough first step to simplifying. We need more work on this structure, as it's quite scattered with responsibilities and has quite a bit of duplication, especially if you look at `GitChangesDetector` and all its streaming and filtering logic. I will create separated issue for it.

PR #50, opened 2017-07-07
New approach for writing functional tests 

bartoszmajsak:
This PR brings new way of writing functional tests for our smart testing tool. Here's how the sample test looks like:

```java
@Test
public void should_execute_all_new_tests_and_related_to_production_code_changes() {
   // given
            .executionOrder(NEW, AFFECTED)
            .inMode(SELECTING)
        .enable();

    final List<TestResult> expectedTestResults = project
        .applyAsLocalChanges("Single method body modification - sysout",
            "Inlined variable in a method", "Adds new unit test");

    // when
    final List<TestResult> actualTestResults = project.build();

    // then
    assertThat(actualTestResults).containsAll(expectedTestResults)
             .hasSameSizeAs(expectedTestResults);
    }
}
```

We are using `arquillian-core` project (squashed repo) for running those tests. You can find it here: https://github.com/arquillian/smart-testing-dogfood-repo

## Design idea

Prior to this solution, our `ftests` were simple projects on which we were applying changes and demo the capabilities mostly in the manual fashion. The main goal for **Test Bed** module is to automate this process and simplify functional testing of our test optimization tool by providing a programmatic way for these activities.

#### Project configuration 

Easy way of configuring any Maven-based project in the test itself through simple and self-explanatory fluent API. 

```java
                .executionOrder(NEW, AFFECTED)
                .inMode(SELECTING)
            .enable();
```

This snippet will modify root `pom.xml` of given project with all required elements to enable `smart-testing`. It will also apply test selection criteria as well as execution mode (`selecting` meaning only tests which belong to categories defined in `executionOrder`).

#### Applying changes

I decided to have an ability to apply changes to the selected code base by modifying sample repository directly. This is done by creating atomic commits directly rather than applying code changes in the tests themselves (e.g. by using [Forge Roaster](https://github.com/forge/roaster)). This comes with several advantages:
 
  * Ability to test those changes directly on the project
  * Easier refactoring 
  * Ability to analyze execution when tests are failing by simply opening modified project rather than debugging test itself

##### Creating change sets and test result expectations

Creating atomic change, which will be later used for tests, consists of following steps:

  * Making commit with the change
  * Tagging it with additional message

This message is then used as a `changeDescription` in the project API, for example:

```java
 project.applyAsLocalChanges("Single method body modification - sysout")
```
This will look in the repository for a `tag` with message `Single method body modification - sysout` and apply as local change (no commits created).

Creating such a tag on the commit is as easy as:

```shell
$ git tag affected_01 -m "Single method body modification - sysout"
```
#### Defining test result expectations

In order to verify if our tool works, we have to specify which tests are expected to be executed as well as their statuses. This is done as part of the commit, as at this point we know exactly what we want our tool to execute and what should be the status of our tests based on modifications we made. For example In `arquillian-core` we would like to test `affected` strategy, where we modify `ConfigurationRegistrar#loadConfiguration` by adding `System.out.println` to the method. This should result in finding two related tests `ConfigurationRegistrarTestCase` and  `SyspropReplacementInArqXmlTestCase`. Both of them should be executed successfully. 

In order to define these expectations we create commit with the following message:

```
affected: adds system.out to ConfigurationRegistrar#loadConfiguration

+ org.jboss.arquillian.config.impl.extension.ConfigurationRegistrarTestCase
+ org.jboss.arquillian.config.impl.extension.SyspropReplacementInArqXmlTestCase
```

The first line describes the change, so it's up to you to provide meaningful information. Next two lines define actual expectations. It's the list of the executed tests (fully qualified names) prefixed with an expected status. In this case `+`, which indicates all test methods in the particular class should be executed successfully. See `org.arquillian.smart.testing.ftest.testbed.testresults.Status` for all prefixes.

#### Executing test optimization

Once we configured the project and applied changes we can build it. This is what we want to really test. Under the hood it's using [`Embedded Maven`](https://github.com/shrinkwrap/resolver#embedded-maven). In addition, after the build is finished, it returns actual test results so we can verify expected and actual test execution. It's done this way:

```java
List<TestResult> actualTestResults = project.build();
```

#### Usage of Git

Before the test class starts "project under test" is cloned and for each test method separated copy is created to execute all the changes and builds in isolation. Folders are named using the following convention:

```java
GIT_REPO_FOLDER + "_" + getClass().getSimpleName() + "_" + name.getMethodName()
```

for example:

```shell
smart-testing-dogfood-repo_HistoricalChangesAffectedTestsSelectionExecutionFunctionalTest_should_only_execute_tests_related_to_multiple_commits_in_business_logic_when_affected_is_enabled/
```

### Missing features

This is not a final solution yet, we have several things to address to improve the overall experience. I believe it's a good start though and we will improve on the way. I will create follow-up tasks and update this description.


bartoszmajsak:
The build is failing b/c of forked surefire in Travis.

lordofthejars:
Looks good to me, and the test although at first time you read I have been confused, basically because of `apply-X` method, when you get the point then you see that test bed is really cool way for writing tests.

bartoszmajsak:
>  I have been confused, basically because of apply-X method

Is javadoc helpful enough? https://github.com/arquillian/smart-testing/pull/50/files#diff-df3bd1de056d18ee017bd6449d7785caR40

lordofthejars:
Yes I think it is well described, it is just that my mind let's say that by default expects a commit id and not a string with white spaces. Although I think that the commit message is better than commit id.

bartoszmajsak:
My main reasoning was, that while developing you could rewrite history of your new changes. But also having hash in the test code won't tell you much. Also notice it's not really a commit message, but **tag** message. This way you can re-tag as much as you want and also can have more meaningful description than tag label.

lordofthejars:
@bartoszmajsak yes I completely agree.

Btw are you planning to squash all commits when build succeeds?

bartoszmajsak:
> Btw are you planning to squash all commits when build succeeds?

What do you think? :) It's the only option we have in the GH PRs btw.

lordofthejars:
I mean locally rewriting the history

bartoszmajsak:
> I mean locally rewriting the history

For the project under test? Not it's just adding commits or local (index). Can you shed some more light on where rewriting history would be useful? You got me curious

lordofthejars:
To not have all these list of commits on description when it is squashed to master

bartoszmajsak:
To unblock the work I am merging it. @MatousJobanek please still review and I will improve based on your feedback.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/MavenConfigurator.java
```diff
@@ -0,0 +1,117 @@
...
+        final StringJoiner stringJoiner = new StringJoiner(",");
+        for (final Criteria criterion : criteria) {
+            stringJoiner.add(criterion.name().toLowerCase());
+        }
+        strategies.setValue(stringJoiner.toString());
+        return strategies;
+    }
+
+    private Xpp3Dom defineUsageMode() {
+        final Xpp3Dom usage = new Xpp3Dom("usage");
```
MatousJobanek: I would prefer using constants for the properties `usage` and `strategies` as it should be used in the provider as well.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/configuration/Criteria.java
```diff
@@ -0,0 +1,24 @@
+package org.arquillian.smart.testing.ftest.testbed.configuration;
+
+import org.apache.maven.model.Dependency;
+
+import static org.arquillian.smart.testing.ftest.testbed.project.Project.SMART_TESTING_VERSION;
+
+public enum Criteria {
+    AFFECTED("affected"),
+    NEW("changed");
```
bartoszmajsak: Yup

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/TestBedTemplate.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @ClassRule
+    public static final TemporaryFolder TMP_FOLDER = new TemporaryFolder();
+
+    @Rule
+    public final TestName name = new TestName();
+
+    private static String GIT_REPO_FOLDER;
+
+    protected Project project;
```
bartoszmajsak: Yeah, I was thinking about the rule too, it would serve the same purpose but different way of composing. But I'm not sure if we really have to do it as part of this PR. Might be something to consider as part of #56 ?

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/MavenConfigurator.java
```diff
@@ -0,0 +1,117 @@
...
+        for (final Criteria criterion : criteria) {
+            model.addDependency(criterion.getMavenDependency());
+        }
+    }
+
+    void configureTestRunner() {
+        final Optional<Plugin> testRunnerPluginConfiguration = model.getBuild().getPlugins()
+            .stream()
+            .filter(plugin -> APPLICABLE_PLUGINS.contains(plugin.getArtifactId()))
+            .findFirst();
```
bartoszmajsak: Correct - https://github.com/arquillian/smart-testing/issues/53

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/TestBedTemplate.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private String targetRepoPerTestFolder() {
+        return GIT_REPO_FOLDER + "_" + getClass().getSimpleName() + "_" + name.getMethodName();
+    }
+
+    static void cloneRepository(String repoTarget, String repo) {
+        Spacelift.task(CommandTool.class)
+            .command(new CommandBuilder("git")
+                .parameters("clone", repo, "-b", "master",
```
bartoszmajsak: Seems to work faster and less boilerplate. But we can rewrite it.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/MavenConfigurator.java
bartoszmajsak: This will become obsolete with `Maven Extension` approach, this code doesn't exist anymore.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/TestBedTemplate.java
```diff
@@ -0,0 +1,91 @@
...
+    public final TestName name = new TestName();
+
+    private static String GIT_REPO_FOLDER;
+
+    protected Project project;
+
+    @BeforeClass
+    public static void cloneTestProject() throws IOException {
+        TMP_FOLDER.create();
+        System.out.println(REPO_NAME);
```
bartoszmajsak: No, more a location to be honest :) Leftover - will change it.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/TestBedTemplate.java
MatousJobanek: > Might be something to consider as part of #56 ?

+1

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/TestBedTemplate.java
MatousJobanek: >Seems to work faster and less boilerplate.

Ok, then I'm fine with it - I didn't know it

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/MavenConfigurator.java
MatousJobanek: `Maven Extension` - Yeah, this is something that I'd like to hear more about.
I don't know what are you working on exactly and what is your plan/idea with it...

functional-tests/test-bed/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-model</artifactId>
+      <version>3.3.9</version>
+    </dependency>
```
bartoszmajsak: I wouldn't like to rely on transitive, to be honest, especially that I used mvn model directly to manipulate the POM on the fly. Can it be `version.maven` instead?

MatousJobanek: The thing is that if you change the version of maven-model to a newer one, then it doesn't have to be fully compatible with that one used in SWR. `3.3.9` is used in SWR so there is no problem. Just pointing this out that it could cause unexpected behavior in a future. If you are aware of it, then I'm fine with it

bartoszmajsak: And that's the behaviour we should know ASAP, right? So having explicit model as direct dependency would help us spotting it faster?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/TestBedTemplate.java
bartoszmajsak: Reworked to `jgit` - just a second slower. But less dependencies to take care of :)

functional-tests/test-bed/pom.xml
MatousJobanek: I know that Maven 3.5.0 is not backward compatible and won't work with SWR. Not sure about this specific dependency.

PR #51, opened 2017-07-07
Include / exclude refactoring

dipak-pawar:
Include / exclude refactoring

- Removed  nclude/exclude from Planner implementaions
- Detecting test classes using `**/src/test/java/**/*.java`

**Fixes**: #8

### DIFF
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/MavenConfigurator.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/MavenConfigurator.java
@@ -54,12 +54,12 @@ class MavenConfigurator {
             final Object configuration = testRunnerPlugin.getConfiguration();
             if (configuration != null) {
                 final Xpp3Dom configurationDom = (Xpp3Dom) configuration;
-                final Xpp3Dom properties = getOrCreatePropertiesChild(configurationDom);
+                final Xpp3Dom properties = getOrCreateChild(configurationDom, "properties");
                 properties.addChild(defineUsageMode());
                 properties.addChild(defineTestSelectionCriteria());
-                final Xpp3Dom includes = getOrCreateIncludesChild(configurationDom);
+                final Xpp3Dom includes = getOrCreateChild(configurationDom, "includes");
                 addIncludePattern(includes);
-                final Xpp3Dom excludes = getOrCreateExcludesChild(configurationDom);
+                final Xpp3Dom excludes = getOrCreateChild(configurationDom, "excludes");
                 addExcludePattern(excludes);
             }
         }
@@ -68,60 +68,32 @@ class MavenConfigurator {
     private void addIncludePattern(Xpp3Dom includes) {
         final String[] projectConfiguratorIncludes = projectConfigurator.getIncludes();
         if (projectConfiguratorIncludes != null) {
-            Arrays.stream(projectConfiguratorIncludes).forEach(pattern -> {
-                final boolean matched =
-                    Arrays.stream(includes.getChildren()).anyMatch(child -> child.getValue().equals(pattern));
-                if (!matched) {
-                    includes.addChild(defineInclude(pattern));
-                }
-            });
+            addChildNodeWithPattern(includes, "include", projectConfiguratorIncludes);
         }
     }
 
     private void addExcludePattern(Xpp3Dom excludes) {
         final String[] projectConfiguratorExcludes = projectConfigurator.getExcludes();
         if (projectConfiguratorExcludes != null) {
-
-            Arrays.stream(projectConfiguratorExcludes).forEach(pattern -> {
-                final boolean matched =
-                    Arrays.stream(excludes.getChildren()).anyMatch(child -> child.getValue().equals(pattern));
-                if (!matched) {
-                    excludes.addChild(defineExclude(pattern));
-                }
-            });
+            addChildNodeWithPattern(excludes, "exclude", projectConfiguratorExcludes);
         }
     }
 
-    private Xpp3Dom defineInclude(String pattern) {
-        final Xpp3Dom include = new Xpp3Dom("include");
-        include.setValue(pattern);
-
-        return include;
+    private void addChildNodeWithPattern(Xpp3Dom node, String patternType, String[] patterns) {
+        Arrays.stream(patterns).forEach(pattern -> {
+            final boolean matched =
+                Arrays.stream(node.getChildren()).anyMatch(child -> child.getValue().equals(pattern));
+            if (!matched) {
+                node.addChild(definePattern(patternType, pattern));
+            }
+        });
     }
 
-    private Xpp3Dom defineExclude(String pattern) {
-        final Xpp3Dom exclude = new Xpp3Dom("exclude");
-        exclude.setValue(pattern);
+    private Xpp3Dom definePattern(String patternType, String pattern) {
+        final Xpp3Dom node = new Xpp3Dom(patternType);
+        node.setValue(pattern);
 
-        return exclude;
-    }
-
-    private Xpp3Dom getOrCreateIncludesChild(Xpp3Dom configurationDom) {
-        Xpp3Dom includes = configurationDom.getChild("includes");
-        if (includes == null) {
-            includes = new Xpp3Dom("includes");
-            configurationDom.addChild(includes);
-        }
-        return includes;
-    }
-
-    private Xpp3Dom getOrCreateExcludesChild(Xpp3Dom configurationDom) {
-        Xpp3Dom excludes = configurationDom.getChild("excludes");
-        if (excludes == null) {
-            excludes = new Xpp3Dom("excludes");
-            configurationDom.addChild(excludes);
-        }
-        return excludes;
+        return node;
     }
 
     void update() {
@@ -162,13 +134,13 @@ class MavenConfigurator {
         return usage;
     }
 
-    private Xpp3Dom getOrCreatePropertiesChild(Xpp3Dom configurationDom) {
-        Xpp3Dom properties = configurationDom.getChild("properties");
-        if (properties == null) {
-            properties = new Xpp3Dom("properties");
-            configurationDom.addChild(properties);
+    private Xpp3Dom getOrCreateChild(Xpp3Dom configurationDom, String name) {
+        Xpp3Dom node = configurationDom.getChild(name);
+        if (node == null) {
+            node = new Xpp3Dom(name);
+            configurationDom.addChild(node);
         }
-        return properties;
+        return node;
     }
 
     private Dependency smartTestingProviderDependency() {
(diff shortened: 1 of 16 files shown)


bartoszmajsak:
So instead of having support for `include` and `exclude` which is available in Surefire we simply hardcode location of tests which Maven convention, but can be changed? 

Did you verify how it behaves? What if I would exclude particular test in Surefire, but then change it's code. Would our strategy pick it up and execute?

dipak-pawar:
> So instead of having support for include and exclude which is available in Surefire we simply hardcode location of tests which Maven convention, but can be changed?

This is required to locate only test classes. If it's changed we can give other locations as well.

> Did you verify how it behaves? What if I would exclude particular test in Surefire, but then change its code. Would our strategy pick it up and execute? Is it something getSuites() from original provider takes care of?

If you exclude particular test in surefire, then our tool is not executing it. It's taking care by getSuites & our [filter](https://github.com/arquillian/smart-testing/blob/master/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java#L57)

bartoszmajsak:
> If you exclude particular test in surefire, then our tool is not executing it. It's taking care by getSuites & our filter

Thanks for confirming it. With `test-bed` we should write functional test for it I believe.

> This is required to locate only test classes. If it's changed we can give other locations as well.

It's possible in Maven to configure it, so we can have

```
<build>
  <testSourceDirectory>src/foobar/java</testSourceDirectory>
</build>
```

with our approach we would miss those tests, wouldn't we?

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedChangesDetectorFactory.java
```diff
@@ -36,7 +36,7 @@ public TestExecutionPlanner create(File projectDir, String[] globPatterns) {
         final AllChangedFilesDetector allChangedFilesDetector = new AllChangedFilesDetector(projectDir, previousCommit, commit, "**/src/main/java/**/*.java");
         final Set<File> mainClasses = allChangedFilesDetector.getFiles();
 
-        return new AffectedChangesDetector(new FileSystemTestClassDetector(projectDir, globPatterns), mainClasses);
+        return new AffectedChangesDetector(new FileSystemTestClassDetector(projectDir, "**/src/test/java/**/*.java"), mainClasses);
```
bartoszmajsak: @lordofthejars shouldn't this be pointing to main sources?

dipak-pawar:
@bartosz I will add ftest for this once #50 gets merge

bartoszmajsak:
I've just merged it. Thanks!

bartoszmajsak:
If you will notice any features missing in Test Bed which are not major changes please add them as part of this PR.

dipak-pawar:
@bartoszmajsak I have ran it locally with same command as travis i.e. ` mvn clean verify -P travis -DJAVA_OPTS="-XX:-UseLoopPredicate"`. It's running fine. Can anyone rerun this?

screenshot from 2017-07-11 18-24-14

MatousJobanek:
The reason why it is failing is probably the limited sources available on Travis. I was facing a similar problem when I was running tests for reporter. I solved it by raising the memory allocation pool: https://github.com/arquillian/arquillian-reporter/blob/master/core/reporter-impl/pom.xml#L96-L101
you can try same workaround

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/MavenConfigurator.java
```diff
@@ -56,10 +57,73 @@ void configureTestRunner() {
...
+    }
+
+    private Xpp3Dom getOrCreateExcludesChild(Xpp3Dom configurationDom) {
+        Xpp3Dom excludes = configurationDom.getChild("excludes");
+        if (excludes == null) {
+            excludes = new Xpp3Dom("excludes");
+            configurationDom.addChild(excludes);
+        }
+        return excludes;
+    }
```
MatousJobanek: I see that the methods for includes and excludes have a lot of same code (talking about all of them). isn't it possible to extract the common code into separated method and passing only the string defining if it is include or exclude?

MatousJobanek:
> If you exclude particular test in surefire, then our tool is not executing it. It's taking care by getSuites & our filter

This was supposed to be just a workaround - not stable solution

bartoszmajsak:
> This was supposed to be just a workaround - not stable solution

Can you elaborate what you mean by that?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/affected/LocalChangesAffectedTestsSelectionExecutionFunctionalTest.java
```diff
@@ -30,16 +35,17 @@ public void should_only_execute_tests_related_to_single_local_change_in_producti
...
     @Test
-    public void should_only_execute_tests_related_to_multiple_local_changes_in_production_code_when_affected_is_enabled() throws Exception {
+    public void should_only_execute_tests_related_to_multiple_local_changes_in_production_code_when_affected_is_enabled()
+        throws Exception {
         // given
         project.configureSmartTesting()
-                    .executionOrder(AFFECTED)
-                    .inMode(SELECTING)
+            .executionOrder(AFFECTED)
+            .inMode(SELECTING)
```
bartoszmajsak: I would leave indentations as they were - isn't it more readable in terms of visualizing the flow?

MatousJobanek: I guess that it is a result of auto-formatting process. To be honest, I would do the same thing as Dipak did. I would rather keep formatting specified in `.editorconfig` file

MatousJobanek:
> This was supposed to be just a workaround - not stable solution
> Can you elaborate what you mean by that?

As I wrote in the comment of the issue - I hoped that the strategy appliers could be executed only against the set of test classes retrieved from `getSuites()`.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/affected/LocalChangesAffectedTestsSelectionExecutionFunctionalTest.java
bartoszmajsak: Funny that I personally don't really need autoformatter - IDE really let's you stick to it when you use proper autocompletion. Anyway - we have few options to mitigate this 

  * we stick to this inlining which makes it less readable
  * we introduce regions (in IDEA) so formatter would skip it

>  I would rather keep formatting specified in .editorconfig file

This kind of information is not kept in `.editorconfig` file really, these are rules for IDEs

bartoszmajsak:
> As I wrote in the comment of the issue - I hoped that the strategy appliers could be executed only against the set of test classes retrieved from getSuites().

I'm not sure if that solves the problem to be honest: https://sourcegraph.com/github.com/apache/maven-surefire@c343cc412b986fd9da839696c709542d03679d68/-/blob/surefire-api/src/main/java/org/apache/maven/surefire/providerapi/SurefireProvider.java#L36:27

The comment describing fork modes makes me wonder if we really start with the right assumption

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/affected/LocalChangesAffectedTestsSelectionExecutionFunctionalTest.java
MatousJobanek: > we stick to this inlining which makes it less readable

I would vote for this - for me it is readable - just my opinion

hemanik: I would personally prefer indentation to make it more readable.

bartoszmajsak:
@dipak-pawar for the travis issue please check @MatousJobanek suggestion. We already have these settings in place https://github.com/dipak-pawar/smart-testing/blob/09fa0f13ad8d5c185ac51f1f3a784a862e2f015d/functional-tests/test-bed/pom.xml#L79 but with lower memory.

bartoszmajsak:
`./mvnw clean verify -X -P travis -DJAVA_OPTS="-XX:-UseLoopPredicate"` won't enable full debugging as test-bed runs embedded mvn process. See here https://github.com/arquillian/smart-testing/wiki/Test-Bed#troubleshooting-tests

bartoszmajsak:
But this makes me wonder.... https://github.com/dipak-pawar/smart-testing/blob/604f66cfd546e34802d38976c846a1b9df13ec0d/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java#L130-L133 should appear in the logs. Does mean travis does not show system.out?

bartoszmajsak:
`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project arquillian-config-impl-base: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test failed: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?`

It means that we have resource issues on travis

MatousJobanek:
I was having the very same error in the reporter - raising memory fixed it.
@dipak-pawar you can try some bigger numbers

dipak-pawar:
@MatousJobanek . I will try with big numbers

functional-tests/test-bed/pom.xml
```diff
@@ -70,7 +70,7 @@
           <forkCount>${surefire.fork.count}</forkCount>
           <threadCount>${surefire.thread.count}</threadCount>
           <reuseForks>true</reuseForks>
-          <argLine>-Xmx1024m -XX:MaxPermSize=256m</argLine>
+          <argLine>-Xmx4096m -XX:MaxPermSize=256m</argLine>
```
MatousJobanek: btw, I think that we don't need `-XX:MaxPermSize=256m`
https://travis-ci.org/arquillian/smart-testing/builds/253109080#L7079

bartoszmajsak: Yeah, as it's Java 8

dipak-pawar:
>  you can try some bigger numbers
@MatousJobanek I tried with big number i.e. 4096 getting the same issue.

bartoszmajsak:
Just for the record - Travis worker has 3GB of RAM available for the whole build - so setting it higher won't give any effect

bartoszmajsak:
This has been reworked with Maven Extension approach and the only remaining bit is for `affected` strategy which has dedicated clean-up issue. Thanks for your work, especially around travis (which seems to be working fine now after all the tweaks we put in place!).

PR #58, opened 2017-07-12
Debug modes for Test Bed embedded build.

bartoszmajsak:
### Overview

This feature brings the ability to execute embedded build in Test Bed in the debug mode. This gives is better analysis of the build execution, as it lets us connect to the build execution to analyze behaviour of our extension in the runtime,

### Features

Enhanced project configuration DSL for enabling debug mode directly in the code:

```java
final List<TestResult> actualTestResults = project
    .buildOptions()
        .withRemoteDebugging()
        .withRemoteSurefireDebugging()
        .withDebugOutput()
        .configure()
    .build();
```

Setting up any of the debugging modes would set agent in suspend mode for both maven execution and surefire execution, thus giving you time to attach debuggers to both.

But this can be also achieved through system properties which you can set as part of your "Run Configuration" in the IDE. Following system properties are available:

  * `test.bed.mvn.remote.debug` - enables remote debugging of the embedded maven build with `suspend` set to `y` (so the build will wait until we attach remote debugger to the port) and port `8000` to listen.
  * `test.bed.mvn.remote.debug.suspend` - `true` or `false` (both for maven and surefire)
  * `test.bed.mvn.remote.debug.port` - valid port used for remote debugging of maven execution
  * `test.bed.mvn.surefire.remote.debug` - enables [remote debugging for surefire](http://maven.apache.org/surefire/maven-surefire-plugin/examples/debugging.html) 
  * `test.bed.mvn.surefire.remote.debug.port` - valid port used for remoting debugging of surefire execution
  * `test.bed.mvn.debug.output` - `true` or `false` - sets `-X` flag for embedded maven build for more details build log.

**NOTE**: if both system property and the programmatic option is used system property takes precedence.

### DIFF
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
@@ -121,7 +121,7 @@ public class ProjectBuilder {
         final BuiltProject build = embeddedMaven
                     .setGoals(goals)
                     .setDebug(isMavenDebugOutputEnabled())
-                    .setQuiet(!isMavenDebugOutputEnabled() && quietMode)
+                    .setQuiet(disableQuietWhenAnyDebugModeEnabled() && quietMode)
                     .skipTests(false)
                     .setProperties(systemProperties)
                     .ignoreFailure()
@@ -135,10 +135,13 @@ public class ProjectBuilder {
         return accumulatedTestResults();
     }
 
+    private boolean disableQuietWhenAnyDebugModeEnabled() {
+        return !isMavenDebugOutputEnabled() && !isSurefireRemoteDebuggingEnabled() && !isRemoteDebugEnabled();
+    }
+
     private void enableDebugOptions(PomEquippedEmbeddedMaven embeddedMaven) {
         if (isRemoteDebugEnabled()) {
             final String debugOptions = String.format(MVN_DEBUG_AGENT, shouldSuspend(), getRemotePort());
-            System.out.println(">>> Executing build with debug options: " + debugOptions);
             embeddedMaven.setMavenOpts(debugOptions);
         }
 
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
@@ -23,7 +23,8 @@ import static org.arquillian.smart.testing.ftest.testbed.testresults.SurefireRep
 public class ProjectBuilder {
 
     private static final String TEST_REPORT_PREFIX = "TEST-";
-    private static final String MVN_DEBUG_MODE = "-agentlib:jdwp=transport=dt_socket,server=y,suspend=%s,address=%s";
+    private static final String MVN_DEBUG_AGENT = "-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=%s,address=%s";
+    private static final String SUREFIRE_DEBUG_SETTINGS = " -Xnoagent -Djava.compiler=NONE";
     private static final int DEFAULT_DEBUG_PORT = 8000;
     private static final int DEFAULT_SUREFIRE_DEBUG_PORT = 5005;
 
@@ -38,7 +39,6 @@ public class ProjectBuilder {
     private boolean suspend = true;
     private boolean mvnDebugOutput;
     private boolean enableSurefireRemoteDebugging = false;
-    private boolean surefireRemoteDebuggingEnabled;
 
     ProjectBuilder(Path root, Project project) {
         this.root = root;
@@ -79,6 +79,7 @@ public class ProjectBuilder {
     public ProjectBuilder withRemoteSurefireDebugging(int surefireRemotePort, boolean suspend) {
         this.surefireRemotePort = surefireRemotePort;
         this.suspend = suspend;
+        this.enableSurefireRemoteDebugging = true;
         return this;
     }
 
@@ -136,15 +137,14 @@ public class ProjectBuilder {
 
     private void enableDebugOptions(PomEquippedEmbeddedMaven embeddedMaven) {
         if (isRemoteDebugEnabled()) {
-            final String debugOptions = String.format(MVN_DEBUG_MODE, shouldSuspend(), getRemotePort());
+            final String debugOptions = String.format(MVN_DEBUG_AGENT, shouldSuspend(), getRemotePort());
             System.out.println(">>> Executing build with debug options: " + debugOptions);
             embeddedMaven.setMavenOpts(debugOptions);
         }
 
         if (isSurefireRemoteDebuggingEnabled()) {
             this.systemProperties.put("maven.surefire.debug",
-                String.format("Xdebug -Xrunjdwp:transport=dt_socket,server=y,"
-                    + "suspend=%s,address=%s -Xnoagent -Djava.compiler=NONE", shouldSuspend(), getSurefireDebugPort()));
+                String.format(MVN_DEBUG_AGENT, shouldSuspend(), getSurefireDebugPort()) + SUREFIRE_DEBUG_SETTINGS);
         }
     }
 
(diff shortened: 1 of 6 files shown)


functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/newtests/HistoricalChangesNewTestsSelectionExecutionFunctionalTest.java
```diff
@@ -47,7 +51,9 @@ public void should_only_execute_new_test_when_multiple_commits_with_changed_test
 
         // when
         final List<TestResult> actualTestResults = project
-            .withEnvVariables("git.last.commits", "3")
+            .buildOptions()
+                .withEnvVariables("git.last.commits", "3")
```
lordofthejars: I know the convention, but if we are using the fully qualified form in all methods, why not rename this to `withEnvironmentVariables`. I think that in case of system properties we would call it `withSystemProperties` and not `withSysProperties`

bartoszmajsak: Correct, forgot to rename it

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/newtests/HistoricalChangesNewTestsSelectionExecutionFunctionalTest.java
```diff
@@ -47,7 +51,9 @@ public void should_only_execute_new_test_when_multiple_commits_with_changed_test
 
         // when
         final List<TestResult> actualTestResults = project
-            .withEnvVariables("git.last.commits", "3")
+            .buildOptions()
+                .withSystemProperties("git.last.commits", "3")
+                .configure()
```
MatousJobanek: I don't think that this is correct formatting. I understand why you did this, but if I apply auto-formatting on this, then the lines 55 and 56 would be moved one tab left...

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
```diff
@@ -5,35 +5,109 @@
...
+            final String debugOptions = String.format(MVN_DEBUG_MODE, shouldSuspend(), getRemotePort());
+            System.out.println(">>> Executing build with debug options: " + debugOptions);
+            embeddedMaven.setMavenOpts(debugOptions);
+        }
+
+        final BuiltProject build = embeddedMaven
                     .setGoals(goals)
-                    .setQuiet()
+                    .setDebug(isMavenDebugOutputEnabled())
+                    .setQuiet(!isMavenDebugOutputEnabled() && quietMode)
```
MatousJobanek: By default, the quite mode is true which means that nothing from the maven build is printed. But when I run with remote debugging enabled then I cannot see the line 
`Listening for transport dt_socket at address: 5005`
which can be confusing. Or WDYT?

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
```diff
@@ -5,35 +5,109 @@
...
 import static org.arquillian.smart.testing.ftest.testbed.testresults.Status.FAILURE;
 import static org.arquillian.smart.testing.ftest.testbed.testresults.Status.PASSED;
 import static org.arquillian.smart.testing.ftest.testbed.testresults.SurefireReportReader.loadTestResults;
 
-class ProjectBuilder {
+public class ProjectBuilder {
 
     private static final String TEST_REPORT_PREFIX = "TEST-";
+    private static final String MVN_DEBUG_MODE = "-agentlib:jdwp=transport=dt_socket,server=y,suspend=%s,address=%s";
+    private static final int DEFAULT_DEBUG_PORT = 5005;
```
MatousJobanek: isn't the default port 8000?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/newtests/HistoricalChangesNewTestsSelectionExecutionFunctionalTest.java
bartoszmajsak: Then we shouldn't have autoformatting enabled? Or do we ignore formatting for tests? Or regions?

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
bartoszmajsak: Depends on what you use - IDEA has 5005 as default so that's what I used

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
bartoszmajsak: That's why you have `System.out.println(">>> Executing build with debug options: " + debugOptions);` on line `101`  - you will see it on the whole test class output

I'm not sure if I really want to see the whole log when I'm debugging, but  we can also change it to
```java
.setQuiet(!(isMavenDebugOutputEnabled() || isRemoteDebugEnabled())   && quietMode)
```

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
MatousJobanek: I was talking about Maven debug mode - when you run `mvnDebug` which is provided by Maven. `5005` is a default port for debug mode of surefire-plugin

bartoszmajsak: Correct, I will change it to `"-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8000"`

and this would imply to have log enabled too.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/newtests/HistoricalChangesNewTestsSelectionExecutionFunctionalTest.java
MatousJobanek: I probably don't understand your questions... By default, I'm trying to use the auto formatting only for changed part of the code. I just don't like when some part of the code has different formatting that it is set by default in our styles or `.editorconfig` file - then if I added a new code, then it would look different then your code...

bartoszmajsak: Can we have this discussion in one place? 
https://github.com/arquillian/smart-testing/pull/51#discussion_r126934151

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
bartoszmajsak: So what do we prefer? :)

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
bartoszmajsak: Fixed.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
MatousJobanek: I thought that you have already answered it: 

>Correct, I will change it to "-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8000"
and this would imply to have log enabled too.

But If you prefer having the log quite, I'm fine. After all, you're using it now, so you should know what is the best user experience ;-)

bartoszmajsak: there should be 'but' instead of 'and' :P Lemme think about it. I added improvements based on your suggestions from original PR #50 .

bartoszmajsak: Consensus is: we keep it quiet unless enabled by `withMavenDebugOutput` or `quiet(false)`.

lordofthejars: For me are two different things, one is that you want to debug and the other one is that you want to print things, tester should enable both if he wants to see both. But it is just my opinion and any other way is fine for me.

Issue #42, opened 2017-06-27
Ability to disable surefire provider using a flag


bartoszmajsak:
##### Issue Overview

There is no way of disabling `smart-testing-surefire-provider` other than commenting out / removing it from the `pom.xml`. We should implement simple flag switch so we can disable it in the same way as one can disable tests using `-DskipTests` or `-DskipITs`. 

Our flag (e.g. `disableSmartTesting` and shorter form `disableST`) should follow the same convention

  * if only name specified assume it's enabled
  * if value (`true` or `false`) is provided it should be respected

PR #60, opened 2017-07-13
Adds option to disable-smart-testing.

hemanik:
Decides the selection of tests for executions based on the flag. 

- If `disableSmartTesting` is set or `disableSmartTesting = true` , all tests are executed regardless of any `mode` or `strategy`.
- If `disableSmartTesting = false`, tests are executed based on the selected `mode` and `strategy`. This is also the the default behavior.

- Reads and validates system property for `disableSmartTesting` flag value. 
- Makes selection of tests to run based on the value.
- Adds some functional tests to verify the behavior.

**Fixes**: #42


bartoszmajsak:
Why is it still `[WIP]`?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/affected/HistoricalChangesAffectedTestsSelectionExecutionFunctionalTest.java
```diff
@@ -54,6 +54,64 @@ public void should_only_execute_tests_related_to_multiple_commits_in_business_lo
...
+        assertThat(actualTestResults).hasSize(77);
+    }
+
+    @Test
+    public void should_execute_all_tests_when_smart_testing_is_disabled_irrespective_of_strategy() throws Exception {
+        // given
+        project.configureSmartTesting()
+            .executionOrder(AFFECTED)
+            .inMode(SELECTING);
+            //.enable();
```
bartoszmajsak: Having it done this way will never really bring our extension into the maven build. This test is equal to running `$ mvn clean install -DdisableSmartTesting=true` on any random Maven project. It will run normally as our extension is not even configured. Essentially `.enable()` takes care of adding our extension to the build configuration. Why is it commented out?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/affected/HistoricalChangesAffectedTestsSelectionExecutionFunctionalTest.java
```diff
@@ -54,6 +54,64 @@ public void should_only_execute_tests_related_to_multiple_commits_in_business_lo
...
+
+        project
+            .applyAsCommits("Single method body modification - sysout",
+                "Inlined variable in a method");
+
+        // when
+        final List<TestResult> actualTestResults = project
+            .buildOptions()
+            .withSystemProperties("git.last.commits", "2", "disableSmartTesting", "true")
```
bartoszmajsak: Why do we need `"git.last.commits", "2"`?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/simple/DisabledSmartTestingTestSelection.java
```diff
@@ -0,0 +1,24 @@
...
+public class DisabledSmartTestingTestSelection extends TestBedTemplate {
+
+    @Test
+    public void should_execute_all_tests_when_disable_smart_testing_is_disabled() throws Exception {
+        // when
+        final List<TestResult> actualTestResults = project
+            .buildOptions()
+                .withSystemProperties("disableSmartTesting", "true")
+                .configure()
+            .build();
```
bartoszmajsak: Same as `should_execute_all_tests_when_smart_testing_is_disabled_irrespective_of_strategy` - there is no smart-testing extension configured so we just run regular build.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
```diff
@@ -18,16 +17,18 @@
...
     private ProviderParameters bootParams;
+    private final TestsToRun testsToRun;
+
 
     public SmartTestingSurefireProvider(ProviderParameters bootParams) {
         this.bootParams = bootParams;
         this.paramParser = new ProviderParametersParser(this.bootParams);
         this.providerClass = new ProviderList(this.paramParser).resolve();
         this.surefireProvider = createSurefireProviderInstance();
+        this.testsToRun = (TestsToRun) getSuites();
```
bartoszmajsak: I don't think this is the optimal flow of doing things. We can simply check if the flag is set and return `testsToRun` at the moment we get it from the parent (so second line of `getTestsToRun()`) before we kick in all the strategies. 

That would make it simpler to reason about.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/Validate.java
```diff
@@ -9,4 +9,12 @@ public static boolean isEmpty(String string){
     public static boolean isNotEmpty(String string){
         return !isEmpty(string);
     }
+
+    public static boolean isDisableSmartTesting() {
```
bartoszmajsak: This can be one line and if so it probably shouldn't belong to this class but can be called directly where we need that.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/affected/HistoricalChangesAffectedTestsSelectionExecutionFunctionalTest.java
hemanik: It was because the tests were failing because of https://github.com/arquillian/smart-testing/issues/62.  Added `setProjects` as a workaround for the time being to include limited scope.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/affected/HistoricalChangesAffectedTestsSelectionExecutionFunctionalTest.java
hemanik: We need this otherwise, no tests get included in the build.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
```diff
@@ -27,8 +28,14 @@ public SmartTestingSurefireProvider(ProviderParameters bootParams) {
     }
 
     private TestsToRun getTestsToRun() {
+
         final TestsToRun testsToRun = (TestsToRun) getSuites();
 
+        if (isDisableSmartTesting())
+        {
```
bartoszmajsak: Why such a strange formatting?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/simple/DisabledSmartTestingTest.java
```diff
@@ -0,0 +1,85 @@
...
+            .enable();
+
+        project
+            .applyAsCommits("Single method body modification - sysout",
+                "Inlined variable in a method");
+
+        // when
+        final List<TestResult> actualTestResults = project
+            .buildOptions()
+                .withSubProjects("config/impl-base")
```
bartoszmajsak: Why do we build only this project if we are in `SELECTING` mode?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/affected/HistoricalChangesAffectedTestsSelectionExecutionFunctionalTest.java
bartoszmajsak: But if you disable smart testing how should that matter?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/simple/DisabledSmartTestingTest.java
hemanik: This was done in order to skip `testNG tests`.

bartoszmajsak: I understand, but I thought it's only when we have "ORDERING" mode?

hemanik: The issue is both for `ORDERING` as well as `SELECTING` mode.

bartoszmajsak: Interesting that I am not able to reproduce this issue locally, but that might have something to do with race conditions. Let's fix #62 then.

I understood original bug report saying

> Tests should execute as per the strategy used for ordering.

as it implies "ORDERING" mode. Thanks for explanation.

bartoszmajsak:
We still have [failures](https://travis-ci.org/arquillian/smart-testing/builds/254756236#L2232) on the PR build related to new tests. @hemanik do you know what might be the reason for that?

hemanik:
@bartoszmajsak, probably because of missing dependencies for the project `config/impl-base`.
```
Failed to execute goal on project arquillian-config-impl-base: Could not resolve dependencies for project org.jboss.arquillian.config:arquillian-config-impl-base:jar:1.2.0-SNAPSHOT: The following artifacts could not be resolved: org.jboss.arquillian.config:arquillian-config-api:jar:1.2.0-SNAPSHOT, org.jboss.arquillian.core:arquillian-core-api:jar:1.2.0-SNAPSHOT, org.jboss.arquillian.core:arquillian-core-spi:jar:1.2.0-SNAPSHOT, org.jboss.arquillian.core:arquillian-core-impl-base:jar:tests:1.2.0-SNAPSHOT, org.jboss.arquillian.core:arquillian-core-impl-base:jar:1.2.0-SNAPSHOT: Could not find artifact org.jboss.arquillian.config:arquillian-config-api:jar:1.2.0-SNAPSHOT in sonatype-snapshots (https://oss.sonatype.org/content/repositories/snapshots/)
```

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/simple/DisabledSmartTestingFunctionalTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+        project
+            .applyAsCommits("Single method body modification - sysout",
+                "Inlined variable in a method");
+
+        // when
+        final List<TestResult> actualTestResults = project
+            .buildOptions()
+                .withExcludeProjects(excludedProjects)
+                .withSystemProperties("git.last.commits", "2", "disableSmartTesting", "true")
```
bartoszmajsak: If we disable our plugin why do we need `"git.last.commits", "2"` ?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/simple/DisabledSmartTestingFunctionalTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.arquillian.smart.testing.ftest.testbed.testresults.TestResult;
+import org.junit.Test;
+
+import static org.arquillian.smart.testing.ftest.testbed.configuration.Criteria.AFFECTED;
+import static org.arquillian.smart.testing.ftest.testbed.configuration.Mode.SELECTING;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class DisabledSmartTestingFunctionalTest extends TestBedTemplate {
+
+    private String[] excludedProjects = new String[] {"testng", "testng/core", "testng/container", "testng/standalone"};
```
bartoszmajsak: Can we give this field a better name? Also - it can be static.

bartoszmajsak:
Upstream 1b5aa3f

Issue #49, opened 2017-07-06
Report test selection as a build artifact


bartoszmajsak:
Next to printing out which strategies and modes have been used it would useful to report it as a build artifact (report file). This report could contain following information:

  * Selected strategies/criteria and mode
  * List of tests 

XML format is all that we need I believe. It will not only be useful for storing historical executions (and maybe fine tune the selection based on it), but also will let us write stronger tests as we will have a way to peek into the selection in the functional tests.

### Reporting selected tests

Maybe next to having fully qualified test classes it would be useful to indicate to which categories particular test belonged. This will imply changing

```java
public interface TestExecutionPlanner {

    Collection<String> getTests();
}
```

To return type representing selected test with needed information. As we are storing those tests in a Set we would need to make sure that we compare only class names and not other metadata this new type will contain.

Thoughts?

PR #63, opened 2017-07-18
Implemented report test selection as build artifact

dipak-pawar:
Now test selection is reporting as build artifact

- Generate xml report for test selection
- configure reportDir using `smart.testing.report.dir` & reportName using `smart.testing.report.name`
- implmeneted required builders to generate reports

**Fixes**: #49 

**Report structure**
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<smart-testing-execution>
    <module>core</module>
    <executionConfiguration>
        <usageMode>ordering</usageMode>
        <strategies>
            <strategy>new</strategy>
            <strategy>changed</strategy>
        </strategies>
        <properties>
            <property name="smart.testing" value="new,changed"/>
            <property name="smart.testing.report.dir" value="target/smart-testing-report/"/>
            <property name="smart.testing.report.name" value="report"/>
        </properties>
    </executionConfiguration>
    <selection>
        <tests>
            <test name="org.arquillian.smart.testing.report.SmartTestExecutionReportTest" strategies="new,changed"/>
            <test name="org.arquillian.smart.testing.report.SmartTestExecutionReportUsingPropertyTest" strategies="new"/>
        </tests>
    </selection>
</smart-testing-execution>
```


bartoszmajsak:
It seems that the unit test is failing on [Travis](https://travis-ci.org/arquillian/smart-testing/builds/262129846#L2252):

```
Failed tests: 
  TestExecutionReportMarshallerTest.should_generate_report_in_default_dir:82 
Expecting file:
  </home/travis/build/arquillian/smart-testing/core/target/smart-testing-report.xml>
to exist.
```

core/src/main/java/org/arquillian/smart/testing/report/Strategy.java
```diff
@@ -0,0 +1,63 @@
+package org.arquillian.smart.testing.report;
```
bartoszmajsak: How about moving all the data classes to `model` subpackage?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
```diff
@@ -37,15 +42,45 @@ TestsToRun apply(Configuration configuration) {
...
+        final List<Strategy> strategyList = new ArrayList<>();
         final List<String> strategies = Arrays.asList(configuration.getStrategies());
         final Set<TestSelection> selectedTests = new LinkedHashSet<>();
         for (final String strategy : strategies) {
             final TestExecutionPlanner plannerForStrategy = testExecutionPlannerLoader.getPlannerForStrategy(strategy);
-            selectedTests.addAll(plannerForStrategy.getTests());
+            final Collection<TestSelection> selections = plannerForStrategy.getTests();
+
+            selectedTests.addAll(selections);
+            if (!reportDisabled) {
```
bartoszmajsak: Maybe extract method once folded to one `if`?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
```diff
@@ -37,15 +42,45 @@ TestsToRun apply(Configuration configuration) {
...
+                strategyList.add(Strategy.newBuilder()
+                    .withStrategy(strategy)
+                    .withTests(selections.stream().map(TestSelection::getClassName).collect(Collectors.toList()))
+                    .build()
+                );
+            }
+
         }
         logger.info("Applied strategies: %s", strategies);
```
bartoszmajsak: Move this up so you don't have to have two `if`s.

dipak-pawar: I think it's not possible. because we are iterating over strategies to get all classes for that strategy in first if where we are building `Strategy` List.

Once we iterate over entire fore loop, we do have list of strategy ["new", "changed"] & list of Strategy built using builder in above for loop. so we can just simply build object with these values after for loop.

bartoszmajsak:
@dipak-pawar can you take care of merge conflict?

core/src/test/java/org/arquillian/smart/testing/report/SmartTestExecutionReportUsingPropertyTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.junit.contrib.java.lang.system.RestoreSystemProperties;
+import org.junit.experimental.categories.Category;
+
+import static java.util.Arrays.asList;
+import static org.arquillian.smart.testing.Configuration.SMART_TESTING;
+import static org.arquillian.smart.testing.report.SmartTestExecutionReport.SMART_TESTING_REPORT_DIR;
+import static org.arquillian.smart.testing.report.SmartTestExecutionReport.SMART_TESTING_REPORT_NAME;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Category(NotThreadSafe.class)
```
bartoszmajsak: We should have default values for report directory etc. Users should only have a way to configure it in exceptional cases (convention over configuration). With this we also don't need to mark this test as not thread safe, as it will not touch system properties at all.

docs/reports.adoc
```diff
@@ -0,0 +1,32 @@
...
+`smart.testing.report.dir`:: To configure dir location to put generated reports.
+`smart.testing.report.name`:: To configure the file name of the generated report.
+
+=== Sample Report
+
+Here’s how the sample report looks like:
+
+[source, xml, indent=0]
+----
+include::../core/src/test/resources/sample-report.xml[]
```
bartoszmajsak: I like that you used it as living documentation and also that it's used as verification - this way we ensure it's always up-to-date with our implementation. :1st_place_medal:

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
```diff
@@ -44,18 +48,29 @@ TestsToRun apply(Configuration configuration) {
                 + "For details on how to configure it head over to http://bit.ly/st-config");
             return Collections.emptySet();
         }
-        final Set<TestSelection> selectedTests = new LinkedHashSet<>();
+        final List<TestSelection> selectedTests = new ArrayList<>();
```
bartoszmajsak: Is it really necessary to have it as a `List`? With `Set` we were implicitly removing duplicates. Does it still hold true?

core/src/test/java/org/arquillian/smart/testing/report/SmartTestExecutionReportUsingPropertyTest.java
dipak-pawar: we do have default values for report directory which is I mention in documentation. Here we are verifying, if we set dir other than default, is it behaving as expected?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
dipak-pawar: yes, but to get all strategies list, we have to use i as `List`, otherwise we will miss strategies for other `testClasses`, as they are not adding if we are using set.

We are converting entire List into Set at the end of this function

dipak-pawar:
@bartoszmajsak Resolved conflicts

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
bartoszmajsak: I think it should be all done in one place and probably this is the place to do it.

core/src/test/java/org/arquillian/smart/testing/report/SmartTestExecutionReportUsingPropertyTest.java
bartoszmajsak: Then maybe one test checking it in a separated module would be enough.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
dipak-pawar: but then may be we have to merge it before adding duplicate `TestSelection` to set, then we can do it

bartoszmajsak: that's exactly what I meant. how much work do you think it is?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
```diff
@@ -44,18 +51,27 @@ TestsToRun apply(Configuration configuration) {
...
+            final ExecutionReporter
+                executionReporter = new ExecutionReporter(testSelections, configuration);
+            executionReporter.createReport();
+        }
+
+        return testSelections
             .stream()
             .map(TestSelection::getClassName)
-            .filter(this::presentOnClasspath)
             .filter(this::isInTestToRun)
```
bartoszmajsak: Shouldn't this be also in the `filterAndMergeTestSelection`?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
```diff
@@ -45,10 +46,15 @@ private TestsToRun getOptimizedTestsToRun(TestsToRun testsToRun) {
                 return testsToRun.getClassByName(className) != null;
             });
 
-        return new TestStrategyApplier(testsToRun, testExecutionPlannerLoader, bootParams.getTestClassLoader()).apply(
+        return new TestStrategyApplier(testsToRun, testExecutionPlannerLoader, bootParams.getTestClassLoader(), getBasedir()).apply(
```
bartoszmajsak: We need to inverse the control of creating those object. Things we are explicitly passing around is getting too big. I will work on the DI.

bartoszmajsak:
Great work @dipak-pawar! :medal_military:

Issue #56, opened 2017-07-07
Persisting project under test folder for analysis in case of test failure


bartoszmajsak:
Each test class uses `TemporaryFolder` rule to store original repository as well as copies used by each and  every test method. This makes it problematic in case of tests failures, as after each test execution the folder is erased. There should be a way to configure `TestBed` to either:

  * not deleting the folder - which would imply not using `TemporaryFolder` rule, but ours with extended capabilities (most likely could be just extended)
  * hooking after test execution and copying folder to pre-defined location

bartoszmajsak:
Depends on PR #50

bartoszmajsak:
Related discussion https://github.com/arquillian/smart-testing/pull/50#discussion_r126489751

PR #66, opened 2017-07-21
persisting test project with system property for test-bed

dipak-pawar:

- Implemented `TestBed`, `GitClone`, `PersistTempFolder`, `CopyProjectOnTestFailure` Junit `@Rule`
- Now persisting test project under `/tmp/` directory if `test.bed.project.persist` property is set
- In case of test failure you will get projects for failed tests under `/target/projects/` directory

**Fixes**: #56


bartoszmajsak:
What would be also useful is to persist project whenever the test failed. Could you add this functionality?

dipak-pawar:
@bartoszmajsak Added functionality to persist project whenever test failed

dipak-pawar:
@MatousJobanek Yes, we can use `TestBedTemplate` class as non-abstract class.

bartoszmajsak:
I think Matous was asking about not having a need to extend any class at all but have a rule instead. That was also the original idea for this task.

functional-tests/test-bed/pom.xml
```diff
@@ -23,7 +23,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <scope>test</scope>
+      <scope>compile</scope>
```
bartoszmajsak: We don't have to be explicit - `compile` is the default scope when not specified.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/GitClone.java
```diff
@@ -0,0 +1,52 @@
...
+        Git.cloneRepository()
+                .setURI(repo)
+                .setDirectory(new File(repoTarget))
+                .setCloneAllBranches(true)
+            .call()
+                .checkout()
+                    .setName("master")
+            .call();
+
+        System.out.println("cloned test repository to: " + repoTarget);
```
bartoszmajsak: We might want to use our logger instead of sysouts

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/CopyProjectOnTestFailure.java
```diff
@@ -0,0 +1,45 @@
...
+public class CopyProjectOnTestFailure extends TestWatcher {
+
+    TestBed testBed;
+
+    public CopyProjectOnTestFailure(TestBed testBed) {
+        this.testBed = testBed;
+    }
+
+    @Override
+    protected void failed(Throwable e, Description description) {
```
bartoszmajsak: If we copy the project on failure do we still need `PersistTempFolder`?

dipak-pawar:
@bartoszmajsak I couldn't understand that from this [comment](https://github.com/arquillian/smart-testing/pull/50#discussion_r126489751). No worries. I will update this PR.

Thanks

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/testbed/ProjectPersistFail.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.arquillian.smart.testing.ftest.TestBedTemplate;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ProjectPersistFail extends TestBedTemplate {
+
+    @Test
+    public void should_fail() throws Exception {
+        Assert.assertFalse(true);
```
bartoszmajsak: How about making it an inner class for `ProjectPersistTest` - it would be easier to understand the purpose of this test class.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/testbed/ProjectPersistTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.junit.contrib.java.lang.system.RestoreSystemProperties;
+import org.junit.runner.JUnitCore;
+import org.junit.runner.Result;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ProjectPersistTest {
+
+    @Rule
+    public final RestoreSystemProperties restoreSystemProperties = new RestoreSystemProperties();
```
bartoszmajsak: Do we really need it if we don't change properties?

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/TestBed.java
```diff
@@ -0,0 +1,81 @@
...
+    private Path createPerTestRepository() throws IOException {
+        final Path source = Paths.get(GitClone.GIT_REPO_FOLDER);
+        final Path target = Paths.get(targetRepoPerTestFolder);
+        final List<Path> sources = Files.walk(source).collect(toList());
+        final List<Path> targets = sources.stream().map(source::relativize).map(target::resolve)
+            .collect(toList());
+        for (int i = 0; i < sources.size(); i++) {
+            Files.copy(sources.get(i), targets.get(i));
+        }
+        System.out.println("Cloned test repository to: " + target);
```
bartoszmajsak: Logger + "Copied test repository"

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/TestBed.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private Project project;
+
+    public String getTargetRepoPerTestFolder() {
+        return targetRepoPerTestFolder;
+    }
+
+    private String targetRepoPerTestFolder;
+
+    public TestBed() {
```
bartoszmajsak: Do we need default constructor if that's the only one we have?

functional-tests/test-bed/pom.xml
dipak-pawar: yes you  are right, but it's overriding from [parent](https://github.com/arquillian/smart-testing/blob/master/pom.xml#L89) where it's defined as `test` scope.

bartoszmajsak: Beauty of github reviews - never a full picture :)

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/CopyProjectOnTestFailure.java
dipak-pawar: The idea to keep `PersistTempFolder` is  you can persist entire projects by setting system property without considering test results. It means it will persist even tests is passing.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/testbed/ProjectPersistTest.java
dipak-pawar: no

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/TestBed.java
dipak-pawar: I missed it. good catch

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/CopyProjectOnTestFailure.java
bartoszmajsak: That might be useful sometimes, but then maybe we could have it as one rule instead? Where on success it will persist only when you have a property (or maybe passing `true` through the constructor)?

bartoszmajsak:
Great job @dipak-pawar! Could you update Test Bed wiki with new features?

bartoszmajsak:
Upstream 951a377

Issue #43, opened 2017-06-27
Affected strategy doesn't pick up changes made in the other module


bartoszmajsak:
##### Issue Overview

While running `affected` strategy on [arquillian-core](https://github.com/arquillian/arquillian-core) I noticed that not all the changes are leading to the discovery of related tests.

This could be related to the fact that we use `FileSystemTestClassDetector` only.

##### Current Behaviour

Code modified in one module is not treated as affecting tests from the other one. 

Generally speaking, it is a fair assumption that tests and code under test belong to the same module, but as soon as we cross the boundaries of unit testing this might not hold true anymore.

##### Expected Behaviour

When changing source in one module, tests which are related to it, but belong to another module, should be detected and run if the whole project is built.

##### Steps To Reproduce

I created a [branch with initial setup for Arquillian Core](https://github.com/bartoszmajsak/arquillian-core/tree/smart_testing_poc). From there just edit [`StupidClass`](https://github.com/bartoszmajsak/arquillian-core/tree/smart_testing_poc).

Expected behaviour: [`ConfigurationRegistrarTestCase`](https://github.com/bartoszmajsak/arquillian-core/blob/bfa27257f8c2bc7134a0b45e5cfce965c31915b6/config/impl-base/src/test/java/org/jboss/arquillian/config/impl/extension/ConfigurationRegistrarTestCase.java#L50) should be executed, even though it belongs to another module.

bartoszmajsak:
After #44

Issue #54, opened 2017-07-07
Make smart-testing version and git test bed repo configurable in functional tests


bartoszmajsak:
We have locked version of smart-testing which is used for functional tests. This should be configurable. 

  * `testbed.smart.testing.version` environment variable can be used to define the desired version of smart-testing tool used for functional tests
  * if aforementioned property is not defined, it should be defaulted to current project version
  * add method to configure other git repo for test bed instead of the default one.

bartoszmajsak:
Depends on PR #50

bartoszmajsak:
Needs to be aligned with #66 - thus pending.

PR #67, opened 2017-07-21
fix(#43): Adds imports that has not already imported in graph 

lordofthejars:
Affected strategy creates a graph between tests and their imports. In this way for a given change on business code, we are able to detect which tests are directly involved on testing this business code. The problem is that when the business code was inside a `JAR` file instead of current project, then imports are not added as dependency in graph because path is not a physical `file::` path but a `jar:` path.

The solution might be to just make the path optional in case of business objects.


bartoszmajsak:
#54 upstream 05e6974
#43 upstream f1417e3

Issue #62, opened 2017-07-17
TestNG Tests in the dogfooding repo giving error when trying to run with specified mode.


hemanik:
##### Issue Overview

When executing TestNG tests in the dogfooding repo with Smart Testing provider, we get the following error, 
```
The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
Command was /bin/sh -c cd /tmp/junit3013285195732663576/smart-testing-dogfood-repo_HistoricalChangesAffectedTestsSelectionExecutionFunctionalTest_should_execute_all_tests_when_smart_testing_is_disabled_irrespective_of_strategy/testng/core && /usr/java/jdk1.8.0_60/jre/bin/java -jar /tmp/junit3013285195732663576/smart-testing-dogfood-repo_HistoricalChangesAffectedTestsSelectionExecutionFunctionalTest_should_execute_all_tests_when_smart_testing_is_disabled_irrespective_of_strategy/testng/core/target/surefire/surefirebooter3375551549339713544.jar /tmp/junit3013285195732663576/smart-testing-dogfood-repo_HistoricalChangesAffectedTestsSelectionExecutionFunctionalTest_should_execute_all_tests_when_smart_testing_is_disabled_irrespective_of_strategy/testng/core/target/surefire 2017-07-17T19-42-19_270-jvmRun1 surefire6725514113717898404tmp surefire_08997551454508629680tmp
Error occurred in starting fork, check output in log
```  
Here's the related issue: https://gist.github.com/juherr/6eb3e93e2db33979b7e90b63ddadc888

##### Expected Behaviour
Tests should execute as per the strategy and mode selected.

##### Current Behaviour
The output is as follows,
```
cat org.jboss.arquillian.testng.TestNGIntegrationTestCase-output.txt 
Configuring TestNG with: TestNG60Configurator
Cannot use a threadCount parameter less than 1; 1 > 0
Usage: <main class> [options] The XML suite files to run
  Options:
    -configfailurepolicy
       Configuration failure policy (skip or continue)
    -d
       Output directory
    -dataproviderthreadcount
       Number of threads to use when running data providers
    -excludegroups
       Comma-separated list of group names to  exclude
    -groups
       Comma-separated list of group names to be run
    -junit
       JUnit mode
       Default: false
    -listener
       List of .class files or list of class names implementing ITestListener or
       ISuiteListener
    -methods
       Comma separated of test methods
       Default: []
    -methodselectors
[...]

PR #71, opened 2017-07-26
issue #62 When TestNG provider is used, then non-positive threadcount param is removed

MatousJobanek:
When TestNG provider is used, then if the threadcount value is lower than 1 then the param is removed. Having the method `convertProviderParameters` opens possibilities for other future changes/workarounds.

Part of this commit is also a change from ProviderList to SurefireProviderFactory to keep related logic in one place

### DIFF
--- a/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/TestFramework.java
+++ b/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/TestFramework.java
@@ -1,6 +1,0 @@
-package org.arquillian.smart.testing.surefire.provider.info;
-
-public enum TestFramework {
-
-    TESTNG, JUNIT;
-}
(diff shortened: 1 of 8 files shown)


surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/TestFramework.java
```diff
@@ -0,0 +1,6 @@
+package org.arquillian.smart.testing.surefire.provider.info;
+
+public enum TestFramework {
+
+    TESTNG, JUNIT;
```
MatousJobanek: of course that they are not needed - zombie :-)

MatousJobanek: Good catch 👍

bartoszmajsak:
Upstream b85301e. Thanks!

PR #78, opened 2017-07-31
feat(#53): Allows test bed user to set using failsafe plugin

### LINKED ISSUE #53: Ability to configure test project to run failsafe
Currently `Test Bed` automatically configures project under test to run `surefire`. We should also cover usage of `failsafe`. This would imply following:

* Test Bed can be configurable through `Project` API to decide between failsafe or surefire usage (with surefire being the default)
* We would need to alternate project under test which we are using, as it's based on surefire, so either we:
   * manipulate POM on the fly to disable surefire and enable failsafe with matching `include`s
   * OR we do this change in the repo directly, with the new branch
      * this would require additional methods in the API, as we would need to switch the branch, which is currently set to `master`, e.g. `project.checkoutBranch(branchName)` before applying any changes

lordofthejars:
Allows test bed user to set using failsafe plugin instead of surefire

Closes #53

### DIFF
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/Project.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/Project.java
@@ -4,6 +4,10 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.List;
 import org.arquillian.smart.testing.ftest.testbed.testresults.TestResult;
+import org.eclipse.jgit.api.CreateBranchCommand;
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.api.errors.RefNotFoundException;
 import org.eclipse.jgit.lib.Repository;
 import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
 
@@ -12,10 +16,12 @@ public class Project implements AutoCloseable {
     private final Path root;
     private final Repository repository;
     private final ProjectBuilder projectBuilder;
+    private final Git git;
 
     public Project(Path root) throws IOException {
         this.root = root;
         this.repository = getRepository(root);
+        this.git = new Git(this.repository);
         this.projectBuilder = new ProjectBuilder(root, this);
     }
 
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/TestBed.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/TestBed.java
@@ -19,12 +19,17 @@ import static java.util.stream.Collectors.toList;
 
 public class TestBed implements TestRule {
 
+    private final GitClone gitClone;
     private Project project;
 
     private String targetRepoPerTestFolder;
 
     public static final Logger LOGGER = Logger.getLogger(TestBed.class.getName());
 
+    public TestBed(GitClone gitClone) {
+        this.gitClone = gitClone;
+    }
+    
     public Project getProject() {
         return project;
     }
@@ -101,7 +106,7 @@ public class TestBed implements TestRule {
     }
 
     private String targetRepoPerTestFolder(Description description) {
-        return GitClone.GIT_REPO_FOLDER
+        return gitClone.getGitRepoFolder()
             + "_"
             + description.getTestClass().getSimpleName()
             + "_"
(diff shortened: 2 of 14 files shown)


functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/Project.java
```diff
@@ -36,6 +42,32 @@ public ProjectConfigurator configureSmartTesting() {
         return new ChangeApplier(repository).applyAsCommits(changeDescriptions);
     }
 
+    Project checkoutBranch(String branchName) {
+        try {
+            git.checkout().setName(branchName).call();
```
bartoszmajsak: I am not sure if this really should belong here. Maybe instead you can simply `.applyAsCommit("Disable surefire and enable just failsafe plugin")` like we do with other changes?

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/TestBed.java
```diff
@@ -101,15 +106,15 @@ private void after() throws Exception {
     }
 
     private String targetRepoPerTestFolder(Description description) {
-        return GitClone.GIT_REPO_FOLDER
+        return gitClone.getGitRepoFolder()
```
bartoszmajsak: Why do we really need that as a field?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/newtests/LocalChangesNewTestsSelectionExecutionFunctionalTest.java
```diff
@@ -41,6 +42,31 @@ public void should_only_execute_new_tests_related_to_single_local_change() throw
...
+            .executionOrder(NEW)
+            .inMode(SELECTING)
+            .useFailsafe()
+            .enable();
+
+        final List<TestResult> expectedTestResults = project
+            .applyAsLocalChanges("Adds new unit test");
+
+        // when
+        final List<TestResult> actualTestResults = project.build("clean", "verify");
```
bartoszmajsak: Default won't work for this case if we hook failsafe into the right phase?

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/Project.java
lordofthejars: Well in original issue after discussing we agreed to do using a branch for this reason I used this approach, but I agree that maybe using the commit thing might be better at least to have something more similar

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/TestBed.java
lordofthejars: well inspecting the code I saw that this was an static field, so we can have problems when run in parallel threads.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/Project.java
bartoszmajsak: Original discussion didn't live up to expectations then

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/newtests/LocalChangesNewTestsSelectionExecutionFunctionalTest.java
lordofthejars: Probably yes, but probably this is not the real case for users who are going to hook failsafe in verify goal.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/Project.java
lordofthejars: well at that time we had two options:

* manipulate POM 
*  do this change in the repo directly, with the new branch

we agreed to do the second one, for this reason this is done in that way, saying that I am more than open to change it because I suspect there are some problems with this approach.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/TestBed.java
bartoszmajsak: > we can have problems when run in parallel threads.

We do have or we can have?

Why would it be a problem? `GitClone#before` is called as a class rule so it should be invoked only once.

lordofthejars: We have it.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/Project.java
bartoszmajsak: There is no need to defend anything here (why it was done this way, where it was written to be done like this etc.). The conclusion about the code is what matters now. I think having it as a one-liner in the test would make it cleaner. WDYT?

PR #82, opened 2017-08-03
chore(refactor): Adds disable property to Configuration class

lordofthejars:


core/src/main/java/org/arquillian/smart/testing/Configuration.java
```diff
@@ -2,27 +2,34 @@
 
 public class Configuration {
 
+    public static final String DEFAULT_MODE = "ordering";
+    public static final String DEFAULT_STRATEGIES = "";
+
     public static final String SMART_TESTING = "smart.testing";
     public static final String SMART_TESTING_MODE = "smart.testing.mode";
     public static final String SMART_TESTING_PLUGIN = "smart.testing.plugin";
```
bartoszmajsak: What is `SMART_TESTING_PLUGIN`?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/PreBuildTestOptimizer.java
```diff
@@ -30,12 +30,12 @@
     @Override
     public void afterProjectsRead(MavenSession session) throws MavenExecutionException {
 
-        if (isDisableSmartTesting()) {
+        configuration = Configuration.read();
```
lordofthejars: Yes maybe `load` might be better I don't know which could be better in this case

core/src/main/java/org/arquillian/smart/testing/Configuration.java
lordofthejars: If you want to use `failsafe` or `surefire` for smart testing. By default both are modified.

bartoszmajsak: Maybe we could have a better name for the property then - this is a bit misleading.

bartoszmajsak: e.g. `smart.testing.apply.to=failsafe,surefire` ?

lordofthejars: well I will not do it CSV because by default they are both, so user will only use this property to  select one.

bartoszmajsak: sure, that was just an example

bartoszmajsak:
Can you rebase on top of master as I might have fixed travis 56c6f8330b?

lordofthejars:
Yes, I rebase when merged these other changes and then I rebase and squash

PR #86, opened 2017-08-04
fix(surefire 2.20 compatibility) extension adds surefire-api dependency to the list of dependencies

MatousJobanek:
Our `smart-testing-surefire-provider` uses the `surefire-api` dependency (2.19.1 version). If the user uses a newer version of surefire plugin, then it throws this exception:

`java.lang.ClassNotFoundException: org.apache.maven.plugin.surefire.log.api.ConsoleLogger`

The reason is that this class has been moved to a new module `surefire-logger-api` which is (in a normal way) transitively fetched. Since our smart-testing-provider has this dependency in the list of dependencies, then the version `2.19.1` is transitively fetched and (thus) the new logger module is missing.
This is why I have to add the `surefire-api`dependency to the project `Model`- to override the our one.

### DIFF
--- a/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
+++ b/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
@@ -55,9 +55,8 @@ class MavenProjectConfigurator {
     private void addSurefireApiDependency(Model model) {
         if (surefireVersion != null && surefireVersion.toString() != null) {
             boolean alreadyContains = model.getDependencies().stream()
-                .filter(dep ->
-                    dep.getGroupId().equals("org.apache.maven.surefire") && dep.getArtifactId().equals("surefire-api"))
-                .findFirst().isPresent();
+                .anyMatch(dep ->
+                    "org.apache.maven.surefire".equals(dep.getGroupId()) && "surefire-api".equals(dep.getArtifactId()));
             if (!alreadyContains) {
                 model.addDependency(getSurefireApiDependency(surefireVersion.toString()));
             }
--- a/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
+++ b/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
@@ -3,7 +3,6 @@ package org.arquillian.smart.testing.mvn.ext;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.StringWriter;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -20,12 +19,12 @@ import org.codehaus.plexus.util.xml.Xpp3Dom;
 class MavenProjectConfigurator {
 
     private static final Version MINIMUM_VERSION = Version.from("2.19.1");
-    private static final List<String> APPLICABLE_PLUGINS =
-        Arrays.asList("maven-surefire-plugin", "maven-failsafe-plugin");
 
     private static Logger logger = Logger.getLogger(MavenProjectConfigurator.class);
 
     private final Configuration configuration;
+
+    private Version surefireVersion;
 
     MavenProjectConfigurator(Configuration configuration) {
         this.configuration = configuration;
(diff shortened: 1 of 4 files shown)


mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -47,11 +46,33 @@ void addRequiredDependencies(Model model) {
...
         } catch (IOException e) {
             throw new RuntimeException("Unable to load strategy definitions", e);
         }
     }
 
+    private void addSurefireApiDependency(Model model) {
+        if (surefireVersion != null && surefireVersion.toString() != null) {
+            boolean alreadyContains = model.getDependencies().stream()
+                .filter(dep ->
+                    dep.getGroupId().equals("org.apache.maven.surefire") && dep.getArtifactId().equals("surefire-api"))
```
lordofthejars: I would reverse the `equals` so first `String` and then the `dep.get...` to avoid a NPE.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -47,11 +46,33 @@ void addRequiredDependencies(Model model) {
...
+                .findFirst().isPresent();
```
lordofthejars: Use `anyMatch` so you don't need to get the `Optional` and the pass with `isPresent`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
MatousJobanek: I believe that it shouldn't happen that groupId or artifactId would be null

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
MatousJobanek: right - fast writing and less thinking :-)

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -47,11 +46,32 @@ void addRequiredDependencies(Model model) {
...
+            boolean alreadyContains = model.getDependencies().stream()
+                .anyMatch(dep ->
+                    "org.apache.maven.surefire".equals(dep.getGroupId()) && "surefire-api".equals(dep.getArtifactId()));
+            if (!alreadyContains) {
+                model.addDependency(getSurefireApiDependency(surefireVersion.toString()));
+            }
+        }
+    }
+
+    private Dependency getSurefireApiDependency(String version){
```
bartoszmajsak: It might make sense to move dependency related methods to a new class. I will take care of it.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/PreBuildTestOptimizer.java
```diff
@@ -75,8 +75,8 @@ private void configureExtension(MavenSession session, Configuration configuratio
         final MavenProjectConfigurator mavenProjectConfigurator = new MavenProjectConfigurator(configuration);
         session.getAllProjects().forEach(mavenProject -> {
             final Model model = mavenProject.getModel();
-            mavenProjectConfigurator.addRequiredDependencies(model);
             mavenProjectConfigurator.configureTestRunner(model);
+            mavenProjectConfigurator.addRequiredDependencies(model);
```
bartoszmajsak: Is the order significant here?

lordofthejars: I think that order is important I am not 100% sure but for example `configureTestRunner` checks the surefire version is the minimum required one. So probably it is better to first detect these things and then add the dependencies and not the other way  around.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
MatousJobanek: Funny, I had the same idea when I was writing the code, but I was too lazy to create a new class just because of few lines of code :-D.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/PreBuildTestOptimizer.java
MatousJobanek: now it is - in the first step (`configureTestRunner`) it [sets the surefireVersion ](https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java#L90) which is then read in the `addRequiredDependencies` method [here](https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java#L61).
Actually, it opens an interesting question - why do we have two methods that are called in sequence and not only one method that would handle both operations. 
I'll open a PR for it

bartoszmajsak: I've figured it out and refactored a bit, see #88

MatousJobanek: Ah, cool thanks

PR #90, opened 2017-08-04
chore(refactor): Abstracts graph elements form JavaAssist

lordofthejars:
Abstracts graph elements form JavaAssist to a custom type, so equals does not depend on anything except the information we require.
With this solution we have `JavaElement` class which abstracts us in the case of JavaAssist cannot parse a Java Class because it is out of classloader.

### DIFF
--- a/strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
+++ b/strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
@@ -28,8 +28,6 @@
 package org.arquillian.smart.testing.strategies.affected;
 
 import java.io.File;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -50,7 +48,7 @@ public class ClassFileIndex {
     private Logger logger = Logger.getLogger(ClassFileIndex.class);
 
     private final JavaClassBuilder builder;
-    private DirectedGraph<JavaClass, DefaultEdge> graph;
+    private DirectedGraph<JavaElement, DefaultEdge> graph;
     private List<String> globPatterns;
 
     public ClassFileIndex(ClasspathProvider classpath) {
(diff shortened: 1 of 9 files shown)


lordofthejars:
With latest commit I have fixed two things:

* Now method names are has sense in our context
* Sometimes imports were resolved transitively and sometimes not depending if import class was in current classloader or not. Since 0.0.1 we agreed to not have this transitively now code has been simplified a lot so each test class has only imports and not the imports of the imports.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -82,96 +80,99 @@ public ClassFileIndex(ClasspathProvider classpath, List<String> globPatterns) {
...
+        for (String importz : imports) {
+            JavaElement importClass = new JavaElement(importz);
+            if (!importClass.equals(javaElementParentClass)) {
+                if (graph.containsVertex(importClass)) {
+                    graph.addEdge(javaElementParentClass, importClass);
                 } else {
-                    graph.addVertex(childClass);
-                    graph.addEdge(parentClass, childClass);
+                    graph.addVertex(importClass);
+                    graph.addEdge(javaElementParentClass, importClass);
```
bartoszmajsak: We always `graph.addEdge(javaElementParentClass, importClass);` so maybe reverse the `if` statement to only add vertex when does not exist?

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/JavaElement.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public JavaElement(String className) {
+        this.className = className;
+        this.classFile = null;
+    }
+
+    /**
+     * Constructor for creating a Java element.
+     * @param className of the java element.
+     * @param classFile location of the java element. Can be null.
```
bartoszmajsak: What happens if it's `null`?

lordofthejars: Now nothing since this is only used when `JavaElement` is a test which is known because it is the leaf of the graph. This can be null in case of classes that are inside a JAR.

PR #95, opened 2017-08-04
chore(surefire-api test) Verifies if the dependency is correctly added

MatousJobanek:
this is test for this PR: https://github.com/arquillian/smart-testing/pull/86
It just verifies if the dependency is added when there is none and also with a correct version.

### DIFF
--- a/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java
+++ b/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java
@@ -60,8 +60,8 @@ public class DependencyResolver {
     }
 
     static class SurefireApiDependency extends Dependency {
-        private static final String GROUP_ID = "org.apache.maven.surefire";
-        private static final String ARTIFACT_ID = "surefire-api";
+        static final String GROUP_ID = "org.apache.maven.surefire";
+        static final String ARTIFACT_ID = "surefire-api";
 
         SurefireApiDependency(String version) {
             setGroupId(GROUP_ID);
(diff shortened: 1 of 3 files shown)


mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java
```diff
@@ -60,8 +60,8 @@ private Dependency smartTestingProviderDependency() {
     }
 
     static class SurefireApiDependency extends Dependency {
-        private static final String GROUP_ID = "org.apache.maven.surefire";
-        private static final String ARTIFACT_ID = "surefire-api";
+        static final String GROUP_ID = "org.apache.maven.surefire";
+        static final String ARTIFACT_ID = "surefire-api";
```
bartoszmajsak: Do we really need to make it package-protected? Don't see any use of it in this PR.

MatousJobanek: nope - it's zombie from my previous version - I realized when I left the office, but it was too late :-)

bartoszmajsak: I got ya covered ;)

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolverTest.java
```diff
@@ -97,4 +102,65 @@ public void should_overwrite_default_properties_by_those_specified_in_the_file_a
                 "org.arquillian.smart.testing:smart-testing-strategy-affected:0.0.2-SNAPSHOT",
                 "org.arquillian.smart.testing:smart-testing-strategy-failed:0.0.13-SNAPSHOT");
     }
+
+    @Test
+    public void should_add_surefire_dependency()
```
bartoszmajsak: They are testing `DependencyResolver` so probably belong to the other test class. But many thanks for writing them!

Would it make sense to add one covering `-SNAPSHOT` dependency too?

PR #99, opened 2017-08-07
Adds transitivity in affectes strategy

### LINKED ISSUE #98: Transitivity in Affected strategy
Currently Affected strategy just read the imports from the test class and uses this information to decide if given business class should be executed or not.

This means that if you have Test A that depends on  A class and this class at the same time depends on B class `ATest -> A -> B` and if the modified class is `B` then `ATest` is not executed.

Ideally affected strategy should also resolve transitive dependencies of business classes.

lordofthejars:
This is the initial implementation for adding transitivity detection in affected strategy. The important bits are in `f888` commit, the previous ones are because of integration with https://github.com/arquillian/smart-testing/pull/90

Current implementation just gets  imports of imports as well, and the number of recursivity is set by `smart.testing.depth.level` property being `1` the default one.

Also the algorithm chosen to construct the graph was based on flatten packages so the graph always has two levels, bottom level has the Test classes and upper level the business classes that the tests depends directly or indirectly (because of transitivity)

Fixes #98 

Open things to discuss: 

* Probably we need to make this configurable or provide some defaults like TomEE does: https://github.com/apache/tomee/blob/master/container/openejb-core/src/main/resources/default.exclusions To avoid scanning for example org.springframework.*, org.apache.*, ...  and provide a flag to avoid it as well

* Or another option would be do something inclusion which by default could be the first two parts of test package (i.e or.mycompany)

* Update documentation


bartoszmajsak:
Let's use `Status: In Progress` for WIP PRs. When it's finalized, lets simply switch to `Status: Ready`.

lordofthejars:
Current implementation uses regexp-like expressions setting as inclusions or exclusions, having priority exclusions.

The properties to set are: `smart.testing.affected.exclusions` and `smart.testing.affected.inclusions` in CSV format i.e `org.mypackage.*, org.myotherpackage.MyClass`

Also you can set them as a file by using `smart.testing.affected.exclusions.file` and `smart.testing.affected.inclusions.file`

bartoszmajsak:
> Also you can set them as a file by using `smart.testing.affected.exclusions.file` and `smart.testing.affected.inclusions.file`

Can we have one file for that? I believe this would more pleasant to use. For the time being simple property file is fine, but I would like to open the discussion here if we should consider switching to other format in the nearest future (I am mainly thinking about YAML).

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java
```diff
@@ -0,0 +1,61 @@
...
+        return joiner.toString().trim();
+    }
+
+    private static String readFile(String location) {
+        if (location == null) {
+            return null;
+        }
+
+        try (Stream<String> lines = Files.lines(Paths.get(location))) {
+            return lines.collect(Collectors.joining(", ")).trim();
```
MatousJobanek: The string `", "` could be as a constant and reuse it also here https://github.com/arquillian/smart-testing/pull/99/files#diff-83c6b0e6ef034dfbc8970a4e1ed134f8R32

lordofthejars:
The thing is that how you identify what is an inclusion and what exclusion with just one file.

lordofthejars:
> I am mainly thinking about YAML

I think that having two plain files (considering that I think that usually users uses or inclusions or exclusions) it is enough for this use case. But of course maybe I might be wrong and users uses both of them always. What I mean that maybe we implement a YAML format for one use case that it is not widely adopted.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -114,66 +112,59 @@ private void replaceVertex(JavaElement newClass) {
...
     public Set<String> findTestsDependingOn(Set<File> classes) {
-        final Set<JavaElement> javaClasses = classes.stream()
+        return classes.stream()
             .map(javaClass -> JavaToClassLocation.transform(javaClass, globPatterns))
             .map(this.builder::classFileChanged)
             .map(this.builder::getClass)
             .map(JavaElement::new)
             .filter(graph::containsVertex)
+            .map(this::getParents)
+            .flatMap(l -> l.stream())
```
MatousJobanek: `l` please use complete words - to be honest, I got lost after several maps combined with one filter and I'm not sure what I could expect here. Is it a test?

bartoszmajsak:
> The thing is that how you identify what is an inclusion and what exclusion with just one file.

```
inclusions=....,....
exclusions=....,...
```
In my opinion better than two files.

I don't think it would be ever super long so for the time being properties file is enough. But that's exactly the reason why I am proposing different format (`YAML`, `XML` or whatever else makes sense). Don't property files support `\` though? See http://docs.oracle.com/javase/6/docs/api/java/util/Properties.html

> What I mean that maybe we implement a YAML format for one use case that it is not widely adopted.

there is not that much to implement for this, to be honest.

lordofthejars:
>  two files

But I am not sure if users are going to use inclusions and exclusions at same time, I think (or at least in my experience) that you use either exclusions or inclusions so usually you have two properties but only one file.

bartoszmajsak:
> use either exclusions or inclusions so usually you have two properties but only one file.

how would that work with your two properties passed as part of the build command pointing to one file? What would be in this file?

lordofthejars:
> What would be in this file

You would have one file that you set as inclusion or exclusion using the property `inclusions.file` or `exclusions.file`

bartoszmajsak:
I don't think this makes much sense for the user experience perspective. @MatousJobanek what do you think?

lordofthejars:
> I don't think this makes much sense for the user experience perspective

Why if you want to exclude elements you do `-Dsmart.testing.exclusions.file=exclusions.list` and that's all, what is wrong with this?

And if you want to use inclusions approach you do `-Dsmart.testing.inclusions.file=inclusions.list`

What's wrong with this?

MatousJobanek:
> You would have one file that you set as inclusion or exclusion using the property inclusions.file or exclusions.file

I would like to avoid this... I can easily imagine cases when you need to define both.
From my point of view, the most uncluttered way would be having one file with two properties - as @bartoszmajsak  has suggested - the reason for this is when you define both, then you can easily check what is set as inclusion and as exclusion - without opening two files

bartoszmajsak:
1. Too many things to type to configure it
2. It might happen (and it does) that you want to have both inclusions and exclusions defined, which are related to each other. Having this in one file makes it easier to wrap your head around it.

That's what I think is painful with the current solution.

lordofthejars:
Then I implement as a YAML right now. I don't want to do one thing now and then in 3 weeks implement in another way basically because now I have in mind everything.

bartoszmajsak:
> Then I implement as a YAML right now. 

I don't think it makes sense now. Let's stick to the simple property file.

MatousJobanek:
btw, I believe that the example https://github.com/apache/tomee/blob/master/container/openejb-core/src/main/resources/default.exclusions  is slightly different - in our case, it should contain packages - so this should be probably enough: `org.openejb.*`

lordofthejars:
@MatousJobanek it was just an example of this kind of files.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -114,66 +112,59 @@ private void replaceVertex(JavaElement newClass) {
 
     private void updateJavaElementWithImportReferences(JavaElement javaElementParentClass,
         String[] imports) {
+
         for (String importz : imports) {
+
+            if (addImport(javaElementParentClass, importz)) {
+                if (filter.shouldBeIncluded(importz)) {
```
bartoszmajsak: Can this be `&&`?

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -114,66 +112,59 @@ private void replaceVertex(JavaElement newClass) {
...
-        return null;
-    }**/
 
     public Set<String> findTestsDependingOn(Set<File> classes) {
-        final Set<JavaElement> javaClasses = classes.stream()
+        return classes.stream()
             .map(javaClass -> JavaToClassLocation.transform(javaClass, globPatterns))
             .map(this.builder::classFileChanged)
             .map(this.builder::getClass)
             .map(JavaElement::new)
```
bartoszmajsak: Can we combine this `map`s to one? It's hardly readable.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -114,66 +112,59 @@ private void replaceVertex(JavaElement newClass) {
...
-                return jClass;
-            }
-        }
-        return null;
-    }**/
 
     public Set<String> findTestsDependingOn(Set<File> classes) {
-        final Set<JavaElement> javaClasses = classes.stream()
+        return classes.stream()
             .map(javaClass -> JavaToClassLocation.transform(javaClass, globPatterns))
```
bartoszmajsak: Any chance of removing `globPatterns` now?

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/fakeproject/test/ATest.java
```diff
@@ -0,0 +1,17 @@
+package org.arquillian.smart.testing.strategies.affected.fakeproject.test;
+
+import org.arquillian.smart.testing.strategies.affected.fakeproject.main.A;
+import org.junit.Ignore;
+import org.junit.Test;
+
+@Ignore
```
bartoszmajsak: Would be worth to add explanation to `@Ignore` - might help in the future to avoid confusion why we have such "tests"

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedTestsDetector.java
```diff
@@ -59,6 +59,9 @@ public String getName() {
         // TODO this operations should be done in extension to avoid scanning for all modules.
         // TODO In case of Arquillian core is an improvement of 500 ms per module
         // Scan disk finding all tests of current project
+
+        final long beforeDetection = System.currentTimeMillis();
```
bartoszmajsak: Are `TODO`s above still holding true?

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -114,66 +112,59 @@ private void replaceVertex(JavaElement newClass) {
...
+
+        if (!importz.startsWith("java")) {
+
             JavaElement importClass = new JavaElement(importz);
             if (!importClass.equals(javaElementParentClass)) {
                 if (!graph.containsVertex(importClass)) {
                     graph.addVertex(importClass);
                 }
+
+                // This condition is required because we are flattering imports in graph
```
bartoszmajsak: I'm not sure we are **flattering** imports :) Probably flattening.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -114,66 +112,59 @@ private void replaceVertex(JavaElement newClass) {
...
+                // This operation is fast since it is a containsKey call in a Map instance
```
bartoszmajsak: Doesn't it depend on how the `Map` is actually implemented underneath? I'm not sure if really need this comment here to be honest.

bartoszmajsak:
I don't understand this part:

https://github.com/lordofthejars/smart-testing/blob/f3256e7af4d3ee401eb857a7eab370cc0023f3da/strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java#L70-L90

What is the purpose of `Set<JavaClass> changedTestClasses = new HashSet<>();` ? It's never read.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -114,66 +112,59 @@ private void replaceVertex(JavaElement newClass) {
...
+                if (!graph.containsEdge(javaElementParentClass, importClass)) {
+                    graph.addEdge(javaElementParentClass, importClass);
+                    return true;
+                }
                 graph.addEdge(javaElementParentClass, importClass);
```
bartoszmajsak: What is this code really doing?

`boolean addImport(...)` - I would assume `boolean` value here indicates if adding connection was performered or not. However it's not what this code really does. Line `142` and `145` do the same, but in case of condition defined in `141` is not satisfied we add an `edge` (or overwrite it) but return `false`? Then maybe we could simple change this to:

```java
boolean edgeAlreadyDefined = graph.containsEdge(javaElementParentClass, importClass);
graph.addEdge(javaElementParentClass, importClass);
return !edgeAlreadyDefined;
```

But then still there is something blurry in this whole concept.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -114,66 +112,59 @@ private void replaceVertex(JavaElement newClass) {
...
+                        updateJavaElementWithImportReferences(javaElementParentClass, javaClass.getImports());
+                    }
+                }
+            }
+        }
+    }
+
+    private boolean addImport(JavaElement javaElementParentClass, String importz) {
+
+        if (!importz.startsWith("java")) {
```
bartoszmajsak: Couldn't this be part of default `Filter`?

lordofthejars:
> Set<JavaClass> changedTestClasses = new HashSet<>()

I think that during the clean up I miss to remove it

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
lordofthejars: No, I mean it could have but having in filter means that the import itself will be part of the graph, since it is not necessary to have java dependencies in graph it is done in this place.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedTestsDetector.java
lordofthejars: Yes

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
lordofthejars: I fix this because as you pointed the final addEdge was not necessary

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
lordofthejars: I have renamed the methods so can be understood better and combined 3 maps. In my case having several maps is good and I can read them good but I understand also your point and also for debugging is easier.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
lordofthejars: This is because of nature of recursive thing, if you want we can discuss in bjn and I explain why I did in this way, it is just a matter of not adding information that it is not useful

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
lordofthejars: Good question that I don't know if I want to answer 😄 let me dig into

lordofthejars: Probably we should change the variable name since it is not the glob patterns that comes from surefire, this is a glob pattern used to detect if a `File` is a test or not https://github.com/arquillian/smart-testing/blob/master/strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/detector/FileSystemTestClassDetector.java

So if we want to remove it which is possible and correct me if I am wrong I can use the `TestVerifier` class instead of glob resolution.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
bartoszmajsak: I understand why it's there, but as we are having a concept for it - `Filter` - maybe we can also use it for it?

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
bartoszmajsak: Yup

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
lordofthejars: ok, yes it could be used

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/detector/FileSystemTestClassDetector.java
```diff
@@ -14,54 +14,34 @@
...
-        }
-
-        return tests;
-    }
-
-    private Collection<? extends File> findTests(File rootDirectory, PathMatcher pathMatcher) throws IOException {
         try (Stream<Path> stream = Files.walk(rootDirectory.toPath())) {
             return stream
-                .filter(pathMatcher::matches)
+                .filter(path -> !Files.isDirectory(path) && isJavaFile(path) && verifier.isTest(path))
```
MatousJobanek: Why not using `Files.isRegularFile` instead of the negation of the directory check

lordofthejars:
Fixes #99

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -46,42 +47,39 @@
 public class ClassFileIndex {
 
     private Logger logger = Logger.getLogger(ClassFileIndex.class);
+    private static final Filter javaJdkFilter = new Filter("", "java*");
```
bartoszmajsak: This is minor, but `JDK` stands for `Java Development Kit` ;) Maybe `coreJavaFilter` would be more appropriate here?

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/AffectedTestsDetectorTest.java
```diff
@@ -76,4 +75,31 @@ private Path getJavaPath(Class<?> clazz) {
...
+            if (resource.toString().endsWith("MyBusinessObject.java")) {
+                // Since core class is also in test directory, we need that first time core class is detected as such
+                // but second tiem when real location of .class is found returns it is a class so it is resolved correctly to
+                // test-classes directory
+                if (coreClassCount == 0) {
+                    coreClassCount++;
+                    return false;
+                } else {
+                    return true;
+                }
```
bartoszmajsak: This could be replaced as one liner `return (coreClassCount++) != 0`. ;)

lordofthejars:
>  would it make sense to have this transitive use case covered as Test Bed

It is already covered because by default it is enabled this transitivity thing

core/src/main/java/org/arquillian/smart/testing/filter/TestVerifier.java
```diff
@@ -8,4 +8,8 @@
 public interface TestVerifier {
 
     boolean isTest(Path resource);
+
+    default boolean isCore(Path resource) {
```
lordofthejars: Affected needs to know if something is a test or it is a core class.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -46,42 +47,39 @@
 public class ClassFileIndex {
```
lordofthejars: I was thinking exactly the same, maybe `ClassDependenciesGraph`?

bartoszmajsak:
>> would it make sense to have this transitive use case covered as Test Bed

> It is already covered because by default it is enabled this transitivity thing

So how come our tests were passing before we fixed that? I believe we don't have usecase working for  ATest->A->B where we change B and expect ATest to be run.

core/src/main/java/org/arquillian/smart/testing/filter/TestVerifier.java
bartoszmajsak: Isn't it then sufficient to just negate `isTest` on the caller side?

lordofthejars:
Well yes not exactly this use case but I mean that graph logic and so on is covered. Also I think that unit tests that we covered transitivity are really real use cases, in my case I am not really sure if it worth to do a functional test about it. The transitivity thing is something that is really tight to `ClassFileIndex` and it is not exported outside, as I said I am really confident that it is well tested because it is a change really located in one place.

bartoszmajsak:
I would prefer replacing trivial functional test with this use case, rather than having it only in unit tests. The execution time would be the same anyway.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassFileIndex.java
```diff
@@ -114,66 +108,57 @@ private void replaceVertex(JavaElement newClass) {
 
     private void updateJavaElementWithImportReferences(JavaElement javaElementParentClass,
         String[] imports) {
+
         for (String importz : imports) {
+
+            if (addImport(javaElementParentClass, importz) && filter.shouldBeIncluded(importz)) {
```
lordofthejars: Yes it is important the order because you might add something as child, but then you don't want to continue the transitive exploration.

core/src/main/java/org/arquillian/smart/testing/filter/TestVerifier.java
lordofthejars: Well I prefer this way because the logic is encapsulated so anyone can use either one or another. I think it is better for reader read the affirmative case and not the negative.

bartoszmajsak:
Huge work! Thanks @lordofthejars!

PR #102, opened 2017-08-08
Disables extension when test execution is skipped

### LINKED ISSUE #73: Disable extension when test execution is skipped
##### Issue Overview

When user sets `-DskipTests` or `-DskipITs`, depending on the configuration to which smart-testing should be applied, we should disable registration of our provider.

##### Expected Behaviour

Smart Testing is not enabled based on the Maven flags.

##### Current Behaviour

Extension is registering all required dependencies even though `surefire` or `failsafe` is not executed.

hemanik:

- Check for `maven.test.skip`, `skipTests` and `skipITs` property honored by Surefire and Failsafe plugins. 
- Ignore smart-testing-extension registration/configuration.
- Tests to verify the change depending on the logger message. 

Fixes #73

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/Configuration.java
+++ b/core/src/main/java/org/arquillian/smart/testing/Configuration.java
@@ -68,5 +68,4 @@ public class Configuration {
     public boolean isDisabled() {
         return disabled;
     }
-
 }
--- a/core/src/main/java/org/arquillian/smart/testing/Configuration.java
+++ b/core/src/main/java/org/arquillian/smart/testing/Configuration.java
@@ -68,4 +68,5 @@ public class Configuration {
     public boolean isDisabled() {
         return disabled;
     }
+
 }
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
@@ -35,6 +35,7 @@ public class BuildConfigurator {
     private boolean enableSurefireRemoteDebugging = false;
     private boolean ignoreBuildFailure = false;
     private boolean skipTests = false;
+    private String mavenLog = "false";
     private String mavenOpts = "-Xms512m -Xmx1024m";
 
     BuildConfigurator(ProjectBuilder projectBuilder) {
@@ -124,6 +125,11 @@ public class BuildConfigurator {
         return this;
     }
 
+    public BuildConfigurator withMavenLog() {
+        this.mavenLog = "true";
+        return this;
+    }
+
     public BuildConfigurator logBuildOutput() {
         return logBuildOutput(false);
     }
@@ -187,6 +193,14 @@ public class BuildConfigurator {
         }
     }
 
+    void setMavenLog(String mavenLog) {
+        this.mavenLog = mavenLog;
+    }
+
+    String getMavenLog() {
+        return mavenLog;
+    }
+
     void addMavenOpts(String options) {
         this.mavenOpts += " " + options;
     }
@@ -213,6 +227,10 @@ public class BuildConfigurator {
 
     boolean isSkipTestsEnabled() {
         return skipTests;
+    }
+
+    boolean isMavenLoggingEnabled() {
+        return Boolean.valueOf(mavenLog);
     }
 
     String getMavenOpts() {
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
@@ -34,6 +34,7 @@ public class BuildConfigurator {
     private boolean mvnDebugOutput;
     private boolean enableSurefireRemoteDebugging = false;
     private boolean ignoreBuildFailure = false;
+    private boolean skipTests = false;
     private String mavenOpts = "-Xms512m -Xmx1024m";
 
     BuildConfigurator(ProjectBuilder projectBuilder) {
@@ -115,6 +116,11 @@ public class BuildConfigurator {
 
     public BuildConfigurator ignoreBuildFailure() {
         this.ignoreBuildFailure = true;
+        return this;
+    }
+
+    public BuildConfigurator skipTests(boolean skipTests) {
+        this.skipTests = skipTests;
         return this;
     }
 
@@ -203,6 +209,10 @@ public class BuildConfigurator {
 
     boolean isQuietMode() {
         return quietMode;
+    }
+
+    boolean isSkipTestsEnabled() {
+        return skipTests;
     }
 
     String getMavenOpts() {
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
@@ -35,7 +35,6 @@ public class BuildConfigurator {
     private boolean enableSurefireRemoteDebugging = false;
     private boolean ignoreBuildFailure = false;
     private boolean skipTests = false;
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/arquillian/smart/testing/Configuration.java
```diff
@@ -45,6 +45,10 @@ public boolean isModeSet() {
         return this.mode != null;
     }
 
+    public boolean isSkipTests() {
+        return Boolean.valueOf(System.getProperty("maven.test.skip", "false"));
```
lordofthejars: also you can set it by using `skipTests`property. If it is set then using `-DskipTests` then tests are also skipped.

bartoszmajsak: The thing to check here would be if `-DskipTests` does not imply `maven.test.skip` set implictly.

bartoszmajsak:
- [ ] Handle failsafe in the same way

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/DisabledSmartTestingFunctionalTest.java
```diff
@@ -50,4 +50,21 @@ public void should_execute_all_tests_when_smart_testing_is_disabled_irrespective
...
+    @Test
+    public void should_return_empty_set_when_test_execution_is_skipped() throws Exception {
+        // given
+        final Project project = testBed.getProject();
+
+        // when
+        final List<TestResult> actualTestResults = project
+            .build()
+                .options()
+                    //.withSystemProperties("maven.test.skip", "true")
```
bartoszmajsak: What are we testing here?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -33,7 +33,7 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
 
         configuration = Configuration.load();
 
-        if (configuration.isDisabled()) {
+        if (configuration.isDisabled() && configuration.isSkipTests()) {
```
bartoszmajsak: `isSkipTestsSet` reflects better what is the use case here. It doesn't have to follow the artificial javabean spec.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -28,17 +28,21 @@
 
     private Configuration configuration;
 
+    private MavenProjectConfigurator mavenProjectConfigurator;
```
MatousJobanek: It is slightly confusing - storing the `MavenProjectConfigurator` as a global variable just because of reading system properties. If you retrieve the information from system properties, then it can be used as a `public static` method from separated class. The other case would be storing a boolean flag as a information that the ST tool has been initialized (based on information that it has (not) been disabled or the test has (not) been skipped)

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final String EXPECTED_LOG_PART = "Enabling Smart Testing";
+    private ByteArrayOutputStream logCapturingStream;
+    private PrintStream oldPrintStream;
+
+    @Before
+    public void attachLogCaterer() {
+        logCapturingStream = new ByteArrayOutputStream();
+        PrintStream newPrintStream = new PrintStream(logCapturingStream);
+        oldPrintStream = System.out;
+        System.setOut(newPrintStream);
```
bartoszmajsak: `BuiltProject` from Embedded Maven we are using under the hood in `ProjectBuilder` exposes `getMavenLog()`. So I was thinking - maybe instead of replacing the output stream to ours for capturing how about making this log available through DSL? This would also let us run those tests not sequentially.

bartoszmajsak:
Can you check why the build is failing?

hemanik:
When using the `build` with projects, the related project dependencies are not resolved. 
In order to test for instance config/impl-base project, we also need to include dependent projects - config/api, core/api, core/spi, core/impl-base. 
So for the current scenario then we should either include these 5 projects for the test or the complete build.

bartoszmajsak:
> When using the build with projects, the related project dependencies are not resolved.
In order to test for instance config/impl-base project, we also need to include dependent projects - config/api, core/api, core/spi, core/impl-base.
So for the current scenario then we should either include these 5 projects for the test or the complete build.

This is how Maven works unfortunately.

Other option would be to take one of the most upper projects, like `-api` which does not have any dependency on the other modules from the project.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
```diff
@@ -34,6 +34,8 @@
     private boolean mvnDebugOutput;
     private boolean enableSurefireRemoteDebugging = false;
     private boolean ignoreBuildFailure = false;
+    private boolean skipTests = false;
+    private String mavenLog = "false";
```
bartoszmajsak: This is rather fuzzy. I don't really see a reason to have it as a string / container for the log content and a "boolean flag".

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
```diff
@@ -118,6 +120,16 @@ public BuildConfigurator ignoreBuildFailure() {
         return this;
     }
 
+    public BuildConfigurator skipTests(boolean skipTests) {
+        this.skipTests = skipTests;
+        return this;
+    }
+
+    public BuildConfigurator withMavenLog() {
```
bartoszmajsak: It's not really needed. Maven log is always in the case of how we use embedded maven, so simple method for retrieving it from the build is enough. This is making it overcomplicated considering how it's done underneath.

hemanik: Updated

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -58,18 +84,26 @@ void configureTestRunner(Model model) {
...
+    private boolean isSkipTests() {
+        return Boolean.valueOf(System.getProperty("skipTests", "false"));
+    }
+
+    private boolean isSkipITs() {
+        return Boolean.valueOf(System.getProperty("skipITs", "false"));
+    }
+
+    private boolean isSkip() {
+        return Boolean.valueOf(System.getProperty("maven.test.skip", "false"));
```
MatousJobanek: Please, move these methods outside of this class. The logic retrieves static information - so, they can be static methods. 
See this comment: https://github.com/arquillian/smart-testing/pull/102#discussion_r132418126

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -28,17 +28,23 @@
...
 
+    private boolean skipExtension;
+
     @Override
     public void afterProjectsRead(MavenSession session) throws MavenExecutionException {
 
         configuration = Configuration.load();
 
-        if (configuration.isDisabled()) {
+        final MavenProjectConfigurator mavenProjectConfigurator = new MavenProjectConfigurator(configuration);
```
MatousJobanek: As I have written above - it is not necessary to create an instance of the class just because of retrieving system properties...

bartoszmajsak:
Looks good to me, but before we merge we should update documentation as part of this PR.

PR #103, opened 2017-08-09
Adds tests for TestSelection

dipak-pawar:
Adds test for TestSelection

- Adds tests for `TestSelection`
- Changed `HashSet` to `LinkedHashSet` to preserve the order, So we can use it reports with same order as user defined in strategy
- 

Fixes #


bartoszmajsak:
Thanks a lot @dipak-pawar! Once the build is successful go ahead with the merge.

Suggested title: `chore: adds tests for TestSelection (#103)`

PR #107, opened 2017-08-10
Docs

lordofthejars:
Adds new configuration properties for affected strategy and adds in refcard section, a reference for each configuration property of each strategy.

### DIFF
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -89,7 +89,7 @@ inclusions/exclusions system proeprty:: you can set inclusions and/or exclusions
 For example: `-Dsmart.testing.affected.inclusions=org.mysuperbiz.*`.
 
 config properties file:: you can create a properties file containing inclusions and/or exclusions and set its location using `smart.testing.affected.config` system property.
-For example `.Dsmart.testing.affected.config=affected-configuration.properties`
+For example `-Dsmart.testing.affected.config=affected-configuration.properties`
 
 .affected-configuration.proeprties
 ----
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -70,9 +70,7 @@ If the test exercises a business class that has been modified, we treat it as *i
 
 .About transitivity
 ****
-Currently we are resolving in transitive way imports of classes.
-This means that any import of a business class is considered as a dependency of the test too.
-Let's look at next example:
+Our approach does not find only direct test classes related to the business code which was changed, but takes into an account transitivity. This means that any import of a business class is considered as a dependency of the test too.
 
 Suppose we have `ATest.java` which imports `A.java`.
 At the same time `A.java` imports `B.java` (ATest -> A -> B).
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -68,8 +68,43 @@ TIP: `git.commit`, `git.previous.commit` and `git.last.commits` Java system prop
 When this strategy gets all changes then inspect all tests of current project checking which ones _imports_ these classes.
 If the test exercises a business class that has been modified, we treat it as *important*  so it will be executed earlier in the test plan.
 
+.About transitivity
+****
+Currently any we are resolving in transitive way imports.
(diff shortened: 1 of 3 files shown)


docs/configuration.adoc
```diff
@@ -68,9 +68,44 @@ TIP: `git.commit`, `git.previous.commit` and `git.last.commits` Java system prop
 When this strategy gets all changes then inspect all tests of current project checking which ones _imports_ these classes.
 If the test exercises a business class that has been modified, we treat it as *important*  so it will be executed earlier in the test plan.
 
+.About transitivity
+****
+Currently any we are resolving in transitive way imports.
```
bartoszmajsak: any what? I think we have to work on this sentence.

docs/configuration.adoc
```diff
@@ -68,9 +68,44 @@ TIP: `git.commit`, `git.previous.commit` and `git.last.commits` Java system prop
...
+This means that any import of a business class is considered as a dependency of the test too.
+Let's look an example:
```
bartoszmajsak: at the example?

docs/configuration.adoc
```diff
@@ -68,9 +68,44 @@ TIP: `git.commit`, `git.previous.commit` and `git.last.commits` Java system prop
...
+
+Suppose we have `ATest.java` which imports `A.java`.
+At the same time `A.java` imports `B.java` (ATest -> A -> B).
+If `B.java` is modified, then `ATest.java` is considered an *important* test too.
+
+By default this import transitivity is applied to all imports except the ones from `java` and `javax`.
+
+Sometimes you might want to stop this transitivity before reaching these imports, for example in case of developing an application with Spring, you'd probably want to *exclude* `org.springframework` imports.
+Or maybe just *include* imports from your business code, for example all imports from `org.superbiz`.
```
bartoszmajsak: I would avoid any particular library. Let's make sth up. The point is that we use additional libraries which won't change - so we want to optimize it.

docs/configuration.adoc
```diff
@@ -68,9 +68,44 @@ TIP: `git.commit`, `git.previous.commit` and `git.last.commits` Java system prop
...
+----
+
+INFO: Exclusions has precedence over inclusions.
+
+****
+
 IMPORTANT: This strategy is currently only applicable for _white box_ testing approach. At this point our approach is to
  analyze direct code dependencies, but we are working on broader use cases.
 
+WARNING: This strategy does not work with Java 9.
```
bartoszmajsak: "At this moment this strategy does not work with Java 9." ?

docs/what-is-this.adoc
```diff
@@ -28,7 +28,7 @@ and `maven-failsafe-plugin`.
 
 === Minimum requirements
 
-Java:: Requires Java 8.
+Java:: Requires Java 8. In case of using <<Affected>> strategy, Java 9 is not supported yet.
```
dipak-pawar: This looks little confusing. Are we supporting java9 for other strategies. I don't remember anyone tested smart-testing tool with java9

docs/refcard.adoc
```diff
@@ -35,3 +35,41 @@ a|`smart.testing.apply.to`
 a|`surefire`, `failsafe`
 a|`surefire, failsafe`
 |===
+
+=== Strategies
```
dipak-pawar: Can we give change this? I believe following table has more information than strategies

lordofthejars: No, it only contains properties that only configures strategies things

docs/what-is-this.adoc
lordofthejars: @bartoszmajsak did and found not working `affected` because of JavaAssist

bartoszmajsak: Sure, but I haven't checked the other. Something to investigate deeper.

lordofthejars: I think that for the others it should work because we are not manipulating the `.class` but something to take a  look.

PR #108, opened 2017-08-10
feat(#54): makes git repository configurable

### LINKED ISSUE #54: Make smart-testing version and git test bed repo configurable in functional tests

bartoszmajsak:

Makes git repo url configurable through rule constructor (with default using "dogfood").

- ability to configure repo url per test class
- supports both remote and bundles
- reworked unit tests for persisting projects to use bundle taken from small repo - makes it way faster
- minor improvements in `GitClone` rule

Fixes #54


functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/GitClone.java
```diff
@@ -25,24 +42,21 @@ protected void before() throws Throwable {
 
     protected void after() {
         if (tempFolder != null) {
-            recursiveDelete(tempFolder);
+            try {
+                Files.walk(tempFolder.toPath(), FileVisitOption.FOLLOW_LINKS)
+                    .sorted(Comparator.reverseOrder())
+                    .map(Path::toFile)
+                    .forEach(File::deleteOnExit);
```
dipak-pawar: Nice Improved

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/GitClone.java
```diff
@@ -2,21 +2,38 @@
...
 
+    private final String gitRepo;
     private String gitRepoFolder;
+    private File tempFolder;
 
-    public static final Logger LOGGER = Logger.getLogger(GitClone.class.getName());
+    public GitClone() {
+        try {
+            this.gitRepo = new URL("https://github.com/arquillian/smart-testing-dogfood-repo.git").toExternalForm();
+        } catch (MalformedURLException e) {
```
dipak-pawar: We can use `https://github.com/arquillian/smart-testing-dogfood-repo`, so no need of `replace(".git", "")` [here](https://github.com/arquillian/smart-testing/pull/108/files#diff-fe0cc2f698561ea3c4cd3de6d883faa5R75)

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/GitClone.java
```diff
@@ -2,21 +2,38 @@
...
+    private static final Logger LOGGER = Logger.getLogger(GitClone.class.getName());
 
+    private final String gitRepo;
     private String gitRepoFolder;
+    private File tempFolder;
 
-    public static final Logger LOGGER = Logger.getLogger(GitClone.class.getName());
+    public GitClone() {
+        try {
+            this.gitRepo = new URL("https://github.com/arquillian/smart-testing-dogfood-repo.git").toExternalForm();
```
lordofthejars: It is really stupid but why not calling `this("https://.....")`

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/GitClone.java
bartoszmajsak: But with `replace` it supports both.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/rules/GitClone.java
bartoszmajsak: Because `new URL("https://github.com/arquillian/smart-testing-dogfood-repo.git")` throws an exception which would need to end up in constructor signature and make it awful?

lordofthejars: Well the constructor with parameter could receive it as String instead of URL. So it will be easier for the one who wants to use this constructor (Since now this Malformed Exception is passed to test side)

bartoszmajsak: My first comment was before the coffee :) Fix is in.

PR #111, opened 2017-08-16
Renames Git/SCM Properties

### LINKED ISSUE #52: Git properties
We should rename git properties to make them easier to understand. Having `git.commit` and `git.previous.commit` does not clearly explain their purpose nor it's easy to guess. 

I would suggest following changes:

  * `git.commit` -> `scm.range.head`
  * `git.previous.commit` -> `scm.range.tail`
  * `git.last.commits` -> `scm.last.changes`

Thoughts?

hemanik:
Renames Git/SCM Properties to make them easier to understand.

GitRunnerProperties class renamed and Git Properties have been renamed as
- git.commit -> scm.range.head
- git.previous.commit -> scm.range.tail
- git.last.commits -> scm.last.changes

Fixes #52


core/src/main/java/org/arquillian/smart/testing/scm/git/GitRunnerProperties.java
```diff
@@ -2,9 +2,9 @@
 
 public class GitRunnerProperties {
 
-    public static final String PREVIOUS_COMMIT = "git.previous.commit";
-    public static final String COMMIT = "git.commit";
-    static final String LAST_COMMITS = "git.last.commits";
+    public static final String PREVIOUS_COMMIT = "git.range.tail";
```
lordofthejars: The issue mentions using `scm` prefix instead of `git`. Maybe you've talked with @bartoszmajsak about this, but I think it is better to name them with `scm` without any reference to `git` so if we implement for subversion or mercurial we can use the same properties for this purpose.

bartoszmajsak: Alex you are in holidays... 😂 but you are right, these should be as described in the issue.

hemanik: Since it was git properties, I thought  prefix `git` was better. I'll change it to `scm` to make it  extensible.

lordofthejars:
Probably rename class as well and ?move to core?

El 17 ago. 2017 6:49 a. m., "Hemani" <notifications@github.com> escribió:

*@hemanik* commented on this pull request.
------------------------------

In core/src/main/java/org/arquillian/smart/testing/scm/
git/GitRunnerProperties.java
<https://github.com/arquillian/smart-testing/pull/111#discussion_r133622832>
:

> @@ -2,9 +2,9 @@

 public class GitRunnerProperties {

-    public static final String PREVIOUS_COMMIT = "git.previous.commit";
-    public static final String COMMIT = "git.commit";
-    static final String LAST_COMMITS = "git.last.commits";
+    public static final String PREVIOUS_COMMIT = "git.range.tail";

Since it was git properties, I thought prefix git was better. I'll change
it to scm to make it extensible.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/arquillian/smart-testing/pull/111#discussion_r133622832>,
or mute the thread

PR #118, opened 2017-08-22
Adds support for maven test property to execute specific tests.

### LINKED ISSUE #39: Clarify/fix behavior when a concrete set of test classes is specified
When I specify `-Dtest=SomeTest`, I use `testAll=false`, and the test doesn’t fall into the selection, then what we should do? Currently, it doesn't run anything I guess, but the expected behavior would be that the test is run. Right?

hemanik:
Enables the user to specify concrete set of test classes to execute using maven `test` property.

##### Proposed Solution
*To Run a Single Test*

`$ mvn -Dtest=SampleTest test`

NOTE: This configuration disables smart testing and executes only the specified sample test.

*To Run Multiple Test Classes*

You can choose to execute multiple test classes using patterns as supported by maven.

`$ mvn -Dtest=RemoteReusable*Test test`

This property can also be coupled with smart testing's `selecting` and `ordering` mode to choose specific test classes.

When used with `ordering` mode, the specified tests are executed in the order defined by the smart testing strategy used.
If none is applicable to the strategy, they are executed one after the other in no specific order.

When used with `selecting` mode, the specified tests that fit the defined strategy are executed.
However, if none fulfil the strategy filter, no tests would be executed resulting in build failure.
To avoid build failure, either execute using `ordering` mode or by disabling smart testing.

- Add support for maven `test` property.
- Functional Test to verify the behavior when using specific tests.
- Documentation related to it.

Fixes #39


functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SpecificTestClassExecutionFunctionalTest.java
```diff
@@ -0,0 +1,146 @@
...
+            .enable();
+
+        final Collection<TestResult> expectedTestResults = project
+            .applyAsCommits("Adds new unit test");
+
+        // when
+        final Collection<TestResult> actualTestResults = project
+            .build(module)
+                .options()
+                    .withWorkingDirectory(module)
```
hemanik: @bartoszmajsak , @MatousJobanek 
In order to execute the mvn command against a particular sub-module, I want to execute a command similar to `mvn clean test -f config/impl-base/ -Dtest=Properties*`
Is there a built-in functionality available in embedded maven that has a functionality similar to setting `-f` parameter ?

bartoszmajsak: Can't you do something similar like [here](https://github.com/arquillian/smart-testing/commit/b26a69a74b508670995b3528155f42652a2e6efd#diff-d7b0ca86a6c38fc16017af7fe32281a1R64)>

MatousJobanek: @bartoszmajsak that won't work - Maven will try to execute the test for the parent module anyway. And yeah, Hemani has tried it - take a look at line 75.

as for the code - I guess that the `module` in `.withWorkingDirectory(module)` contains only `config/impl-base` which is relative to the current directory of the test bed project, not relative to the dogfooding repo...

bartoszmajsak: How about setting a flag `failIfNoTests=false`?

MatousJobanek: > How about setting a flag failIfNoTests=false?

That should definitely work!
The weird thing is why the working directory is not properly set.

bartoszmajsak: I would suggest following for @hemanik as a follow-up:

  * let's use `failIfNoTests` to unblock the PR
  * work on improvements for Maven resolver
    * introduce `-f` functionality (PR to [`shrinkwrap-resolver`](https://github.com/shrinkwrap/resolver))
    * figure out the problem with `workingDirectory` (PR to [`shrinkwrap-resolver`](https://github.com/shrinkwrap/resolver) if feasible)
  * when both merged change this test to one of the options 

Would you be up for that?

hemanik: Thanks @bartoszmajsak and @MatousJobanek. 
@bartoszmajsak, yes will work on the shrinkwrap-resolver bits. :)

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SpecificTestClassExecutionFunctionalTest.java
```diff
@@ -0,0 +1,140 @@
...
+            .applyAsCommits("Adds new unit test");
+
+        // when
+        final Collection<TestResult> actualTestResults = project
+            .build(module)
+                .options()
+                    .withSystemProperties("scm.range.head", "HEAD", "scm.range.tail", "HEAD~", "test", "Properties*", "failIfNoTests", "false")
+                    .ignoreBuildFailure()
+                .configure()
+            .run("clean", "test");
```
hemanik: Default goals also work. But since the Maven documentation was using test goal with `test` property. I thought that would be preferred one.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SpecificTestClassExecutionFunctionalTest.java
```diff
@@ -0,0 +1,140 @@
...
+        final Collection<TestResult> actualTestResults = project
+            .build(module)
+                .options()
+                    .withSystemProperties("scm.range.head", "HEAD", "scm.range.tail", "HEAD~", "test", "PropertiesParserTestCase", "failIfNoTests", "false")
+                .configure()
+            .run("clean", "test");
+
+        // then
+        String capturedMavenLog = project.getMavenLog();
+        assertThat(capturedMavenLog).doesNotContain(EXPECTED_LOG_PART);
```
hemanik: Haven't tried it. Will check if it is feasible.

hemanik: The tool isn't feasible for us because we do not directly get the instance of logger. In our test-bed, we get execute the main classes from the embedded maven and thus instance of the logger somewhere gets lost and we get null values.
So, we'll have to use maven logs only.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SpecificTestClassExecutionFunctionalTest.java
bartoszmajsak: That's fine too. Also I don't think you need to run `clean` in any of those tests, as each build is executed once so there are no artifacts from any previous build in the project folder.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
```diff
@@ -70,6 +70,8 @@ private BuiltProject executeGoals(String... goals) {
                     .setProperties(asProperties(buildConfigurator.getSystemProperties()))
                     .skipTests(buildConfigurator.isSkipTestsEnabled())
                     .setMavenOpts(buildConfigurator.getMavenOpts())
+                    .setWorkingDirectory(buildConfigurator.getWorkingDirectory())
+                    .setAlsoMake(true)
```
dipak-pawar: How about making this configurable? Instead of setting it true for all the time

bartoszmajsak: Do we benefit from not having this option enabled? Is there any reason for which you would like to change it to `false`?

dipak-pawar: otherwise it'll install all dependent projects even if it's present on your machine which will increase build time.

e.g. In case of one sub-module is dependent on other sub-modules. I already have this submodules installed in my `~/.m2` which is my local, but still it will install it again as we set `also-make` always true which is not necessary here, but required for `CI` environment

bartoszmajsak: I think `also-make` should be enabled by default if we specify only particular projects to build. Otherwise `false` with the ability to enable (? do we really need that then).

I will make new issue for that, let's not spoil this PR if it does not have tremendous impact on the build time. Can you check it @dipak-pawar ?

dipak-pawar: > I think also-make should be enabled by default if we specify only particular projects to build. Otherwise false with the ability to enable (? do we really need that then).

Yes it should be ideally like this. 

> I will make new issue for that, let's not spoil this PR if it does not have tremendous impact on the build time. 
I think it doesn't have much impact we can create new issue for it to work on it later.

dipak-pawar: @bartoszmajsak My assumption was wrong about `-am` option. It's not needed to be configurable as it's applicable only when you project list is specified.

docs/usage.adoc
```diff
@@ -46,6 +46,33 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
...
 To do it you just need to set `smart.testing.apply.to` with either `surefire` or `failsafe` to just enable in one of them.
 
+=== Specifying Concrete Set of Test Classes
+
+During development, you may want to run a single test class or a set of concrete test classes repeatedly. 
+You can do this by setting the maven `test` property to specific test class(es).
+
+==== To Run a Single Test
+
+`$ mvn -Dtest=SampleTest test`
```
MatousJobanek: the format `mvn test -Dtest=SampleTest` is more readable for me

docs/usage.adoc
```diff
@@ -46,6 +46,33 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
...
+
+This property can also be used in combination with smart testing's <<configuration#_modes, `selecting`>> and 
+<<configuration#_modes,`ordering`>> mode to choose specific test classes.
+
+When used with `ordering` mode, the specified tests are executed in the order defined by the smart testing 
+strategy used. If none is applicable to the strategy, they are executed one after the other in no specific order.
+
+When used with `selecting` mode, the specified tests that fit the defined strategy are executed.
+However, if none fulfil the strategy criteria, no tests would be executed resulting in build failure.
+
```
MatousJobanek: There is missing the case when a user specifies multiple tests in the specific order:
-Dtest=TestOne,TestTwo

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -18,4 +23,13 @@ static private boolean isSkip() {
         return Boolean.valueOf(System.getProperty("maven.test.skip", "false"));
     }
 
+    static boolean isSpecificTestClassSet() {
+        String specificTestClasses = System.getProperty("test");
+        return specificTestClasses != null && !containsPattern(specificTestClasses);
+    }
+
+    private static boolean containsPattern(String specificTestClasses) {
+        Matcher matcher = TEST_CLASS_PATTERN.matcher(specificTestClasses);
```
MatousJobanek: this doesn't match the "two tests" case that I mentioned above:
`-Dtest=TestOne,TestTwo`
IIRC our discussion, in this case, it should run both of them & disable ST - unless there is a regex inside of any test name:
`-Dtest=TestOne,TestT*`
then the ST should be activated.
Am I correct?

hemanik: I thought we disable the extension if only a single test needs to be executed. 
When "two tests" given we could still have smart-testing enabled. 

Example: when I have smart testing enabled with suppose `selecting` mode and I specify two tests. If only one of those fulfills the criteria, only one be executed.  

What do you suggest ? (I'll fix it if you suggest otherwise.)

hemanik: And probably I can update the documentation that if the user wants to run specific multiple tests irrespective of what is configured with ST that can be done by explicitly disabling smart-testing.

bartoszmajsak: From what I recall the outcome from the discussion was following:

  * if concrete tests specified - no ST
  * if pattern - ST

docs/usage.adoc
```diff
@@ -46,6 +46,38 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
...
+`$ mvn test -Dtest=SampleTestOne, SampleTestTwo`
+
+NOTE: When multiple specific tests are set without any pattern, smart testing is disabled just like the case 
+for single test.
+
+You can further choose to execute multiple test classes in combination with smart testing's <<configuration#_modes, 
+`selecting`>> and <<configuration#_modes,`ordering`>> mode by using patterns as 
+http://maven.apache.org/surefire/maven-surefire-plugin/examples/single-test.html[supported by Maven].
+
+`$ mvn test -Dtest=RemoteReusable*Test`
```
MatousJobanek: In the examples above, you are using test names containing a string `Sample` - I would use the same approach also here.

docs/usage.adoc
```diff
@@ -46,6 +46,38 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
...
+
+`$ mvn test -Dtest=SampleTest`
+
+NOTE: This configuration disables smart testing and executes only the specified sample test.
+
+==== To Run Multiple Test Classes
+
+You can also choose to run only a concrete set of multiple test classes.
+
+`$ mvn test -Dtest=SampleTestOne, SampleTestTwo`
```
MatousJobanek: IIRC, the convention is putting the word `Test` at the end of the class name

docs/usage.adoc
```diff
@@ -46,6 +46,38 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
...
+for single test.
+
+You can further choose to execute multiple test classes in combination with smart testing's <<configuration#_modes, 
+`selecting`>> and <<configuration#_modes,`ordering`>> mode by using patterns as 
+http://maven.apache.org/surefire/maven-surefire-plugin/examples/single-test.html[supported by Maven].
+
+`$ mvn test -Dtest=RemoteReusable*Test`
+
+When used with `ordering` mode, the specified tests are executed in the order defined by the smart testing
+strategy used. If none is applicable to the strategy, they are executed one after the other in no specific order.
```
MatousJobanek: > in no specific order

Isn't it in the order given by Surefire?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -18,4 +24,14 @@ static private boolean isSkip() {
         return Boolean.valueOf(System.getProperty("maven.test.skip", "false"));
     }
 
+    static boolean isSpecificTestClassSet() {
```
MatousJobanek: how about creating a test for it...?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -18,4 +24,14 @@ static private boolean isSkip() {
...
+        String specificTestClasses = System.getProperty("test");
+        return specificTestClasses != null && !containsPattern(specificTestClasses);
+    }
+
+    private static boolean containsPattern(String specificTestClasses) {
+        return Arrays.stream(specificTestClasses.split(","))
+            .map(TEST_CLASS_PATTERN::matcher)
+            .anyMatch(Matcher::find);
```
MatousJobanek: You shouldn't check if there is at least one specific test specified. You should check if there is some part containing a pattern. 
eg:
`-Dtest=SimpleOneTest,Multiple*Test`
or am I missing something?
The test could show more

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -80,6 +95,6 @@ private void logStrategiesNotDefined() {
     }
 
     private boolean isSkipExtension() {
-        return configuration.isDisabled() || isSkipTestExecutionSet();
+        return configuration.isDisabled() || isSkipTestExecutionSet() || isSpecificTestClassSet();
```
MatousJobanek: How about storing the information that the extension hasn't been loaded? 
The system properties could be tricky as anything could be changed during the test execution. In addition, system properties are not thread-safe, but I'm not sure if we would face this problem in this case.
WDYT @bartoszmajsak

docs/usage.adoc
```diff
@@ -46,6 +46,38 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
...
+
+When used with `selecting` mode, the specified tests that fit the defined strategy are executed.
+However, if none fulfil the strategy criteria, no tests would be executed resulting in build failure.
```
MatousJobanek: > resulting in build failure

really? the build will fail. If so, then I don't think that this something that user would expect. WDYT?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
hemanik: I don't get this one. Here, I am splitting the list of test classes set with `test` property and traversing over each test to check if any of it contains pattern.  Probably the variable`specificTestClasses` is confusing. 
Will fix that and add a few more tests for all the cases.  I had added few generic test cases because in my opinion the functionality was more or less the same.

docs/usage.adoc
hemanik: Yes, It will fail with error `Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project arquillian-core-api: No tests were executed!  (Set -DfailIfNoTests=false to ignore this error.)` because no tests would be executed in this case.
But good point, will look for a way to handle this condition to fail gracefully.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
bartoszmajsak: > Will fix that and add a few more tests for all the cases

Did we ever fix that?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
bartoszmajsak: Is it covered now?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
bartoszmajsak: I guess you mean logging from the loaded extension that its execution has been skipped. I think simply logging it would make sense. If you agree let's do it like that. Log that we skip the extension and why.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
hemanik: Yes, @bartoszmajsak. This is covered. https://github.com/arquillian/smart-testing/pull/118/files#diff-7e2181222a897b1b345a6e3cc8712122R31

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
hemanik: @bartoszmajsak, we do this here, https://github.com/arquillian/smart-testing/pull/118/files#diff-d4e45adf3ad3bc619ccc2eef6c35f04bR50.
Do you expect something else ?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -18,4 +24,14 @@ static private boolean isSkip() {
...
+    static boolean isSpecificTestClassSet() {
+        String testClasses = System.getProperty("test");
+        return testClasses != null && !containsPattern(testClasses);
+    }
+
+    private static boolean containsPattern(String testClasses) {
+        return Arrays.stream(testClasses.split(","))
+            .map(TEST_CLASS_PATTERN::matcher)
+            .anyMatch(Matcher::find);
+    }
```
hemanik: Yes @bartoszmajsak, I have added the tests and renamed it to `testClasses` as `specificTestClasses` was slightly confusing.

docs/usage.adoc
```diff
@@ -46,6 +46,38 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
 
 To do it you just need to set `smart.testing.apply.to` with either `surefire` or `failsafe` to just enable in one of them.
 
+=== Specifying Concrete Set of Test Classes
+
+During development, you may want to run a single test class or a set of concrete test classes repeatedly.
```
MatousJobanek: > repeatedly

what do you mean by that?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
MatousJobanek: @hemanik Yeah, you are right - this is correct - I misread it.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
bartoszmajsak: Ah storing in this sense :) Ok. Sure, it's read only once and never changed during the execution. It's a small optimization though.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -29,12 +29,17 @@
...
 
+    private boolean extensionNotLoaded;
+
     @Override
     public void afterProjectsRead(MavenSession session) throws MavenExecutionException {
 
         configuration = Configuration.load();
 
-        if (isSkipExtension()) {
+        extensionNotLoaded = isSkipExtension();
```
bartoszmajsak: I would hide it in `isSkipExtension` (also maybe called it `shouldSkipExtensionInstallation`)

```java
private Boolean skipExtensionInstallation;

private boolean shouldSkipExtensionInstallation() {
  if (skipExtensionInstallation == null) {
    // initialize
  }
  return skipExtensionInstallation;
}


```

PR #119, opened 2017-08-22
now ordering looking for previous configuration.

### LINKED ISSUE #77: Ordering process should take into account previous configuration/strategies
`Test` - normal test
`CTest` - Changed test
`FTest` - Failed test
`CFTest` - changed and failed test

Consider this situation:
we use `-Dsmart.testing=changed, failed`
and we get this list of test classes to reorder:

`Test1, CTest1, CFTest1, FTest1, CTest2, Test2, CTest3, CFTest2`

After applying only the `changed` strategy, then the resulting set would be:

`CTest1, CFTest1, CTest2, CTest3, CFTest2, Test1, FTest1, Test2`

When we apply both strategies `changed, failed`, then the resulting set contains:

`CTest1, CFTest1, CTest2, CTest3, CFTest2, FTest1, Test1, Test2`

Which means that only `FTest1` is moved

What I would expect is that after the first ordering, the classes would be divided into two sets - the one containing classes that fall into the condition of the strategy and the rest - in the case of `changed`:

`[ CTest1, CFTest1, CTest2, CTest3, CFTest2 ]  [ Test1, FTest1, Test2 ]`

Then the ordering process for the second strategy would be applied on bots sets separately:

`[ [ CFTest1, CFTest2] [ CTest1, CTest2, CTest3 ] ]   [ [ FTest1 ]  [ Test1, Test2 ] ]`

So the very first classes that will be run are those that fall into the most strategies.

In an analogical way, I would take into account `runOrder` parameter in Surefire configuration. In case that user uses `alphabetical` order and for smart testing the strategy `new` then all classes in both sets should be in the alphabetical order:

`[ ANewTest, BNewTest, CNewTest ] [ ATest, BTest ]`

dipak-pawar:
Ordering process takes into account for previous configuration/strategies

- Introduces comparator to order entire testselection based on priority values of strategies.
- sorting entire collection depending on comaprator & if more than 1 strategy is defined.

Fixes #77


surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/TestSelectionOrderingTest.java
```diff
@@ -0,0 +1,48 @@
...
+public class TestSelectionOrderingTest {
+
+    @Mock
+    private TestsToRun testsToRun;
+
+    @Mock
+    private TestExecutionPlannerLoader testExecutionPlannerLoader;
+
+    @Test
+    public void should_order_test_selection_with_previous_strategy_configuration() {
```
bartoszmajsak: After 2 minutes looking at this test I don't understand what you want to test here. I don't see correlation between test method name which should explain expected behaviour and the test code.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/StrategyComparator.java
```diff
@@ -0,0 +1,31 @@
...
+    private final List<String> strategies;
+
+    StrategyComparator(List<String> strategies) {
+        this.strategies = strategies;
+        this.priorities = priority();
+    }
+
+    private Map<String, Integer> priority() {
+        return strategies.stream()
+            .collect(Collectors.toMap(Function.identity(), s -> (strategies.size() - strategies.indexOf(s)) * 100));
```
bartoszmajsak: Consider following case: we have `[a,b,c,d,e]` strategies. Having this coding tests falling into `[b,e]` (`(5-1) * 100 + (5-4) * 100 = 400 + 100 = 500`) are equal to those which are in `[a]` (`500`) category which is not necessarily what we want.

I would rather think of having some sort of "bitset" representation here, where `i` being a position on the list (but starting from 1) would be treated as `10^(size-i)`. Then having `[a,b,c,d,e]` a test falling into `[b,e]` would be equal to 01001 and the one failling into `[a]` would be 10000, so comparing it would result in `[a]` being executed before `[b,e]` as desired.

Or am I missing something here?

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/TestSelectionOrderingTest.java
bartoszmajsak: Maybe we should test what I have described in the comment related to `StrategyComparator` instead?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/StrategyComparator.java
dipak-pawar: no you are right. I am missing this corner case. Good catch. I was thinking about this while implementing but couldn't find this case. Thanks for pointing it to me.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/StrategiesComparator.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Override
+    public int compare(Collection<String> list1, Collection<String> list2) {
+        final double sum1 = list1.stream().mapToDouble(priorities::get).sum();
+        final double sum2 = list2.stream().mapToDouble(priorities::get).sum();
+
+        return Double.compare(sum2, sum1);
+    }
+
+    private Map<String, Double> priority() {
```
MatousJobanek: I would rename the method to: `getPriorities(List<String> strategies)` which tells what the method does...

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/StrategiesComparator.java
```diff
@@ -0,0 +1,35 @@
...
+        return strategies.stream()
+            .collect(Collectors.toMap(Function.identity(), this::priority));
+    }
+
+    private Double priority(String strategy) {
```
MatousJobanek: This one could be `getPriority(String strategy)`

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
```diff
@@ -85,4 +92,25 @@ private boolean presentOnClasspath(String testClass) {
             throw new IllegalStateException(e);
         }
     }
+
+    private Collection<TestSelection> filterMergeAndOrderTestSelection(Collection<TestSelection> selectedTests,
+        List<String> strategies) {
+
+        final Collection<TestSelection> testSelections = selectedTests
+            .stream()
+            .filter(testSelection -> presentOnClasspath(testSelection.getClassName()))
```
MatousJobanek: this is also checked here https://github.com/arquillian/smart-testing/pull/119/files#diff-356cfd962197d56c7e69c381ba349281R65

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/TestStrategyApplier.java
```diff
@@ -44,15 +49,17 @@ TestsToRun apply(Configuration configuration) {
...
+        final Collection<TestSelection> selectedTests = new ArrayList<>();
         for (final String strategy : strategies) {
             final TestExecutionPlanner plannerForStrategy = testExecutionPlannerLoader.getPlannerForStrategy(strategy);
             selectedTests.addAll(plannerForStrategy.getTests());
         }
         logger.info("Applied strategies: %s", strategies);
         logger.info("Applied usage: [%s]", configuration.getMode().getName());
 
-        return selectedTests
+        final Collection<TestSelection> testSelections = filterMergeAndOrderTestSelection(selectedTests, strategies);
```
MatousJobanek: This doesn't have to be `Collection` - can be `List`

dipak-pawar: I totally understand. but what's wrong with collection if we are using `LinkedHashSet` for collecting it. I believe it's also give you ordered collection

MatousJobanek: Yeah, sure - it can stay as a collection - it was just the first thing that came to my mind when I saw it.
When I was reading the code, I saw the collection and started asking myself - OK, what kind of collection - ordered or unordered one? so, then I had to decipher the logic in the method and figure it out. 
This particular comment is just my opinion, nothing that I would request to change...

PR #120, opened 2017-08-22
Adds documentation related to skipTest functionality.

### LINKED ISSUE #73: Disable extension when test execution is skipped

hemanik:
Adds documentation related to skipTest functionality.

- Updates usage document with skipping test functionality.

Fixes #73


docs/usage.adoc
```diff
@@ -46,6 +46,28 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
...
+
+`$ mvn install -DskipITs`
+
+To skip execution as well as compiling the tests, you can use the `maven.test.skip` property.
+`maven.test.skip` is honored by both Surefire and Failsafe Plugins.
+
+`$ mvn install -Dmaven.test.skip=true`
+
+NOTE: Setting `skipTests` or `maven.test.skip` also disables smart testing extension.
+
```
MatousJobanek: I think that this describes the behavior of surefire, which can be easily traceable: http://maven.apache.org/surefire/maven-surefire-plugin/examples/skipping-tests.html and similar... 
What I would expect from this text is describing what happens (from the Smart Testing point of view) when somebody uses the properties...

hemanik: > What I would expect from this text is describing what happens (from the Smart Testing point of view) when somebody uses the properties...

In my opinion, this is what the text describes, the `skipTest` behavior from ST point of view disables ST extension and is exactly same as the standard Maven behavior.

MatousJobanek: Well, this information is briefly contained in line 69. The rest of the text describes only the "Surefire skip tests" behavior - this I can read in the Surefire documentation. What I want to know is what happens (with Smart Testing tool) if I use any of the properties: `skipTests`, `skipITs` or `maven.test.skip` and why.

docs/usage.adoc
```diff
@@ -46,6 +46,19 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
...
+To skip running the tests for a particular project, Smart Testing extension provides support with 
+http://maven.apache.org/surefire/maven-failsafe-plugin/examples/skipping-tests.html[Maven Surefire/Failsafe 
+plugins' skip test functionality] offered by system properties `skipTests`, `skipITs` and `maven.test.skip`.
+
+Setting properties `skipTests` or `maven.test.skip` disables Smart Testing extension and exhibits standard 
+Surefire/Failsafe plugin behavior.
+
+`skipITs` when used with Smart Testing extension in default mode with both Surefire and Failsafe plugins, 
+skips only the integration tests being run by the Failsafe plugin and continues with unit test(s) execution by
```
MatousJobanek: > and continues with unit test(s) execution

the unit tests are executed as first ones, then the ITs

docs/usage.adoc
```diff
@@ -46,6 +46,19 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
...
+To skip running the tests for a particular project, Smart Testing extension provides support with 
+http://maven.apache.org/surefire/maven-failsafe-plugin/examples/skipping-tests.html[Maven Surefire/Failsafe 
+plugins' skip test functionality] offered by system properties `skipTests`, `skipITs` and `maven.test.skip`.
+
+Setting properties `skipTests` or `maven.test.skip` disables Smart Testing extension and exhibits standard 
+Surefire/Failsafe plugin behavior.
+
+`skipITs` when used with Smart Testing extension in default mode with both Surefire and Failsafe plugins, 
+skips only the integration tests being run by the Failsafe plugin and continues with unit test(s) execution by 
+Surefire plugin configured with Smart Testing configuration.
```
MatousJobanek: > Surefire plugin configured with Smart Testing configuration

this seems to be quite confusing for me. IMHO, there should be mentioned that (on the contrary to `skipTests` or `maven.test.skip` cases) in this case the extension is activated.

PR #121, opened 2017-08-23
feat(api) Introduced SmartTesting fluent API

MatousJobanek:
Moved the API and related implementation to the Core module to make it decoupled from the provider module.
Now it is easier to use Smart Testing in other tools such as Che.

Usage of the API

* in surefire provider:
```java
Set<TestSelection> selection = SmartTesting
            .with(className -> testsToRun.getClassByName(className) != null)
            .in(getBasedir())
            .applyOnClasses(testsToRun);

new TestsToRun(SmartTesting.getClasses(selection));
```

* in Che:
```java
Set<TestSelection> selection = SmartTesting
            .with(className -> suite.contains(className), configuration)
            .in(projectDir)
            .applyOnNames(suite);

SmartTesting.getNames(selection);
```

some additional explanations:
* `applyOnNames` - method is necessary for Che as I get a list of class names that will be executed there
* `TestVerifier.isTest(String classname)` - the reason is again Che usage. In addition, by default the classname was retrieved via `ClassNameExtractor().extractFullyQualifiedName(resource)` anyway so I put it into the default method of the interface
* `in(File projectDirectory)` - in Che I call ST tool from different directory than where the root of the project is.
* in `TestStrategyApplierImpl.selectTests` I removed the check if the class is on classpath because the verification using TestVerifier (if the test is in the list of `TestsToRun`) should be enough


core/src/main/java/org/arquillian/smart/testing/api/TestVerifier.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * Checks whether the given path to a resource represents a non-test class or not
+     *
+     * @param resource A path to a resource to be checked
+     * @return Whether the given path to a resource represents a non-test class or not
+     */
+    default boolean isCore(Path resource) {
+        return !isTest(resource);
```
bartoszmajsak: I'm not convinced if we need that really, because what does it really mean `Core`? But it's not to do in this PR, but to discuss.

core/src/main/java/org/arquillian/smart/testing/impl/ConfiguredSmartTestingImpl.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.io.File;
+import java.util.Set;
+import org.arquillian.smart.testing.Configuration;
+import org.arquillian.smart.testing.api.ConfiguredSmartTesting;
+import org.arquillian.smart.testing.api.TestStrategyApplier;
+import org.arquillian.smart.testing.api.TestVerifier;
+import org.arquillian.smart.testing.spi.JavaSPILoader;
+
+public class ConfiguredSmartTestingImpl implements ConfiguredSmartTesting {
```
bartoszmajsak: I'm just wondering do we really need interfaces and classes implemeting them? Would we ever have 2 implementations for one interface considering DSL usage?

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
```diff
@@ -45,8 +45,10 @@ public GitChangeResolver() {
             System.getProperty(COMMIT, HEAD));
     }
 
-    public GitChangeResolver(File dir) {
-        this(dir, "HEAD", "HEAD");
+    public GitChangeResolver(File projectDir) {
+        this(projectDir,
+            System.getProperty(PREVIOUS_COMMIT, getPrevCommitDefaultValue()),
+            System.getProperty(COMMIT, HEAD));
```
bartoszmajsak: The idea was  that if you only provide a file then we only look at local changes. Why do we need this change?

core/src/test/java/org/arquillian/smart/testing/impl/TestStrategyApplierTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+    }
+
+    private TestExecutionPlannerLoader prepareLoader(final Set<Class<?>> testsToRun, Set<TestSelection> strategyTests) {
+        TestExecutionPlanner testExecutionPlanner = Mockito.mock(TestExecutionPlanner.class);
+        Mockito.when(testExecutionPlanner.getTests()).thenReturn(strategyTests);
+
+        TestExecutionPlannerLoader testExecutionPlannerLoader = Mockito.mock(TestExecutionPlannerLoader.class);
+        Mockito.when(testExecutionPlannerLoader.getPlannerForStrategy("static")).thenReturn(testExecutionPlanner);
+        Mockito.when(testExecutionPlannerLoader.getVerifier())
```
bartoszmajsak: How about static imports?

core/src/main/java/org/arquillian/smart/testing/api/SmartTesting.java
```diff
@@ -0,0 +1,15 @@
+package org.arquillian.smart.testing.api;
```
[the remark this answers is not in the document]
MatousJobanek: does it really matter? :-)
Actually, you don't build anything - you configure and then invoke...

core/src/main/java/org/arquillian/smart/testing/api/TestVerifier.java
MatousJobanek: Yeah, I was asking myself the same question. I just took and kept the code there from the previous state...

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
MatousJobanek: I don't understand your question. 
This is the same constructor as this one: https://github.com/arquillian/smart-testing/pull/121/files#diff-f1255f2030b502bc7deffb7a6797da92R42 with the only difference that you set the path to the project directory. As I have mentioned in the description of the PR - in Che I call ST tool from a different directory than where the root of the project is, which means that this cannot work in my case:
`Paths.get("").toAbsolutePath().toFile()`
But yeah, I can change the referenced constructor to:
```java
    public GitChangeResolver() {
        this(Paths.get("").toAbsolutePath().toFile());
    }
```
to make it more clear

core/src/test/java/org/arquillian/smart/testing/impl/TestStrategyApplierTest.java
MatousJobanek: yeah, good idea :-)

core/src/main/java/org/arquillian/smart/testing/impl/ConfiguredSmartTestingImpl.java
MatousJobanek: Well, probably we don't need it - we could get rid of the `ConfiguredSmartTesting` interface as well as the `TestExecutionPlannerLoader` interface. In case of `TestStrategyApplier` we have two different implementations. 
I used this approach (using interfaces) as IMHO it is a good practice in the case of API...

core/src/main/java/org/arquillian/smart/testing/api/SmartTesting.java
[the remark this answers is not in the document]
bartoszmajsak: `api.configuration` then? I'm not sure if `api` should be one big bag for everything

core/src/main/java/org/arquillian/smart/testing/api/TestVerifier.java
bartoszmajsak: Sounds to me like refactoring task

core/src/main/java/org/arquillian/smart/testing/api/SmartTesting.java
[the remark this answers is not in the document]
bartoszmajsak: `api.configuration` then? I'm not sure if `api` should be one big bag for everything

core/src/main/java/org/arquillian/smart/testing/impl/ConfiguredSmartTestingImpl.java
bartoszmajsak: What does it improve in this given case?

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
bartoszmajsak: Before it `this(dir, "HEAD", "HEAD");` According to the diff, so it is different.

MatousJobanek: Yeah, the `this(dir, "HEAD", "HEAD");` was used only in tests which have been appropriately changed

core/src/main/java/org/arquillian/smart/testing/api/SmartTesting.java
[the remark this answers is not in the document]
MatousJobanek: So, what would you expect from the content of `api` package if we moved all the interfaces to the `api.configuration`. Or do you want to move only some of them... I'm slightly missing what you want to change. 

> one big bag for everything

Com'on, there are four classes/interfaces and I don't expect that it will grow significantly. If there was 10 of 15 of them, then I would say any word, but four?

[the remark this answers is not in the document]
bartoszmajsak: It's not really about amount but boundaries, however if you disagree then we have 1:1 and I would say we leave it.

PR #128, opened 2017-09-06
fix(projectDir) The root dir is not retrived only from the reporterConfiguration

MatousJobanek:
When `basedir` property is null (multiple forks) then project directory is retrieved from `testSourceDirectory` property. 
I expect that the location of the test-source directory is changed less frequently than the location of the reports.

Added also some refactoring changes & test verifying the presence of generated reports in all fork-config test

### DIFF
--- a/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SurefireForksConfigurationTest.java
+++ b/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SurefireForksConfigurationTest.java
@@ -1,11 +1,16 @@
 package org.arquillian.smart.testing.ftest.configuration;
 
+import java.io.File;
 import java.util.Collection;
 import java.util.List;
 import org.arquillian.smart.testing.ftest.testbed.project.Project;
+import org.arquillian.smart.testing.ftest.testbed.project.ProjectBuilder;
 import org.arquillian.smart.testing.ftest.testbed.rules.GitClone;
 import org.arquillian.smart.testing.ftest.testbed.rules.TestBed;
 import org.arquillian.smart.testing.ftest.testbed.testresults.TestResult;
+import org.assertj.core.api.FileAssert;
+import org.assertj.core.api.JUnitSoftAssertions;
+import org.jboss.shrinkwrap.resolver.api.maven.embedded.BuiltProject;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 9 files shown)


surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
```diff
@@ -64,14 +64,17 @@ private TestsToRun getTestsToRun(Object forkTestSet) throws TestSetFailedExcepti
...
 
         return new TestsToRun(SmartTesting.getClasses(selection));
     }
 
-    private String getBasedir() {
-        final String path = this.bootParams.getReporterConfiguration().getReportsDirectory().getPath();
-        return path.substring(0, path.indexOf(File.separator + "target"));
+    private File getProjectDir() {
+        if (System.getProperty("basedir") == null) {
+            return bootParams.getTestRequest().getTestSourceDirectory().getParentFile().getParentFile().getParentFile();
```
bartoszmajsak: Humm... is it always only three levels up? Maybe recursively until we find first `pom.xml`? I know it's unlikely to have it more nested, but why not making it ready for that?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SurefireForksConfigurationTest.java
```diff
@@ -66,15 +75,38 @@ private void verifyTestSuiteExecution(String... systemPropertiesPairs){
...
+        final List<TestResult> actualTestResults = projectBuilder
+            .options()
+                .withSystemProperties(systemPropertiesPairs)
+                .withSystemProperties(ENABLE_REPORT_PROPERTY, "true")
+            .configure()
+            .run();
 
         // then
-        assertThat(actualTestResults).containsAll(expectedTestResults).hasSameSizeAs(expectedTestResults);
+        softly.assertThat(actualTestResults).containsAll(expectedTestResults).hasSameSizeAs(1);
```
bartoszmajsak: `softly.assertThat(actualTestResults).containsAll(expectedTestResults).hasSameSizeAs(1);` this will always break... :)

bartoszmajsak:
Fixed and upstream 57e1e87

Thanks!

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SurefireForksConfigurationTest.java
MatousJobanek: dammit - forgot to change before pushed - I needed to break the test to try it locally. 
Sorry for it :-)

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
MatousJobanek: The reason why I did it in this way is the case when you don't use the default name for the pom file. But yeah, this is really rare and mainly, usually, you have the default `pom.xml` file there and another (custom) one next to it...

bartoszmajsak: Fixed  anyway

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SurefireForksConfigurationTest.java
bartoszmajsak: No worries :)

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SurefireForksConfigurationTest.java
```diff
@@ -66,15 +75,38 @@ private void verifyTestSuiteExecution(String... systemPropertiesPairs){
             .applyAsLocalChanges("Inlined variable in a method", "Adds new unit test", "fixes tests");
 
         // when
-        final List<TestResult> actualTestResults =
-            project
-                .build("config/impl-base")
```
bartoszmajsak: I specifically narrowed it to this module, as it takes 2 extra minutes to run this test if we run the whole build... :)

PR #130, opened 2017-09-07
Sets skipAfterFailureCount to 1 for ordering mode (#96)

### LINKED ISSUE #96: Ordering mode not practical now
After doing some tests with `spark` I noticed that `ordering` mode might not be useful right now. Let me explain two scenarios:

*Project with single module*

In this case `surefire` plugin is used. If one test fails, the build does not fails and continue the execution until the end. So ordering is executing the "affected" test first but it doesn't care if fails at first since you still need to wait the whole build. To fix this `surefire` comes with this property http://maven.apache.org/surefire/maven-surefire-plugin/examples/skip-after-failure.html but does not work in case of `reuseForks` equals to `false`. So maybe in configuration phase if we detect that reuseForks is set to true (default value) then we could configure automatically this flag.

*Multimodule project*

In this case it is the same as *single module* but I think it is even worst. In this case suppose that the modified test is on latest module of the project. Then since surefire is executing modules one after the other, until the module is not executed the build will not fail, so we are effectively ordering tests inside module but not the whole build.

With microservices which everything should be small probably everything is single module but in legacy code or libraries might be a problem.

So maybe we need to think about ordering modules as well.

dipak-pawar:
Setting `skipAfterFailureCount` to 1 in case ordering mode

- Sets `skipAfterFailureCount` to 1
- You can override `skipAfterFailureCount` using `surefire.skipAfterFailureCount` property for `ordering` mode

Fixes #96

### DIFF
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -12,7 +12,7 @@ and then the rest.
 
 In order to define the mode use `smart.testing.mode` Java system property with either one.
 
-IMPORTANT: To get fast feedback loop *Smart testing* is setting `skipAfterFailureCount` to `1`. However you can overwrite it by setting system
+IMPORTANT: To get fast feedback loop *Smart testing* is setting `http://maven.apache.org/surefire/maven-surefire-plugin/examples/skip-after-failure.html[skipAfterFailureCount]` to `1`. However you can overwrite it by setting system
 property `surefire.skipAfterFailureCount` to required value. This behaviour is applicable only if user is using `surefire` plugin to execute
 test having smart testing enabled with `ordering` mode. This is not applicable for `selecting` mode because user can execute all `new/modified/affected`
 tests, identify all failed tests using results and rework on it at once.
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -12,6 +12,11 @@ and then the rest.
 
 In order to define the mode use `smart.testing.mode` Java system property with either one.
 
+IMPORTANT: To get fast feedback loop *Smart testing* is setting `skipAfterFailureCount` to `1`. However you can overwrite it by setting system
+property `surefire.skipAfterFailureCount` to required value. This behaviour is applicable only if user is using `surefire` plugin to execute
+test having smart testing enabled with `ordering` mode. This is not applicable for `selecting` mode because user can execute all `new/modified/affected`
+tests, identify all failed tests using results and rework on it at once.
+
 === Strategies
 
 Until now, you've read that smart testing is changing test execution plan running or only including important tests.
--- a/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipAfterFailureCountTest.java
+++ b/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipAfterFailureCountTest.java
@@ -1,11 +1,10 @@
 package org.arquillian.smart.testing.ftest.configuration;
 
-import java.util.LinkedHashMap;
-import java.util.Map;
 import org.arquillian.smart.testing.ftest.testbed.project.Project;
+import org.arquillian.smart.testing.ftest.testbed.project.TestResults;
 import org.arquillian.smart.testing.ftest.testbed.rules.GitClone;
 import org.arquillian.smart.testing.ftest.testbed.rules.TestBed;
-import org.arquillian.smart.testing.ftest.testbed.testresults.Status;
+import org.arquillian.smart.testing.ftest.testbed.testresults.TestResult;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
@@ -14,6 +13,7 @@ import static org.arquillian.smart.testing.ftest.testbed.configuration.Mode.ORDE
 import static org.arquillian.smart.testing.ftest.testbed.configuration.Strategy.CHANGED;
 import static org.arquillian.smart.testing.ftest.testbed.testresults.Status.SKIPPED;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.tuple;
 
 public class SkipAfterFailureCountTest {
 
@@ -24,7 +24,7 @@ public class SkipAfterFailureCountTest {
     public final TestBed testBed = new TestBed(GIT_CLONE);
 
     @Test
-    public void should_skip_all_remaining_tests_with_first_failure_when_ordering_enabled() throws Exception {
+    public void should_not_execute_remaining_tests_after_first_failure_as_skipAfterFailureCount_set_to_one_by_smart_testing_when_ordering_enabled() throws Exception {
         // given
         final Project project = testBed.getProject();
 
@@ -36,22 +36,26 @@ public class SkipAfterFailureCountTest {
         project.applyAsCommits("Introduces error by changing return value");
 
         // when
-        final Map<Status, Long> resultCount = new LinkedHashMap<>();
-
-        project.build("container/impl-base")
+        final TestResults testResults = project.build("container/impl-base")
             .options()
                 .ignoreBuildFailure()
             .configure()
-            .run(resultCount);
-
+            .run();
 
         // then
-        final Long skippedCount = resultCount.get(SKIPPED);
-        assertThat(skippedCount).isGreaterThan(4);
+        assertThat(testResults.testsWithStatuses(SKIPPED))
+            .extracting(TestResult::getClassName, TestResult::getTestMethod)
+            .contains(
+                tuple("org.jboss.arquillian.container.impl.client.container.ContainerLifecycleControllerTestCase", "org.jboss.arquillian.container.impl.client.container.ContainerLifecycleControllerTestCase"),
+                tuple("org.jboss.arquillian.container.impl.client.container.DeploymentExceptionHandlerTestCase", "org.jboss.arquillian.container.impl.client.container.DeploymentExceptionHandlerTestCase"),
+                tuple("org.jboss.arquillian.container.impl.client.container.ContainerRegistryCreatorTestCase", "org.jboss.arquillian.container.impl.client.container.ContainerRegistryCreatorTestCase"),
+                tuple("org.jboss.arquillian.container.impl.client.container.ContainerDeployControllerTestCase", "org.jboss.arquillian.container.impl.client.container.ContainerDeployControllerTestCase"),
+                tuple("org.jboss.arquillian.container.impl.client.deployment.ArchiveDeploymentExporterTestCase", "org.jboss.arquillian.container.impl.client.deployment.ArchiveDeploymentExporterTestCase")
+            );
     }
 
     @Test
-    public void should_override_skip_after_failure_count_system_property_when_ordering_enabled() throws Exception {
+    public void should_execute_remaining_tests_after_failure_as_skipAfterFailureCount_set_to_zero_with_system_property_when_ordering_enabled() throws Exception {
         // given
         final Project project = testBed.getProject();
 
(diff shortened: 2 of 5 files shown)


mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -50,9 +51,30 @@ void configureTestRunner(Model model) {
...
-            effectiveTestRunnerPluginConfigurations
-                .forEach(dependencyResolver::addAsPluginDependency);
+    private void addSkipAfterFailureCount(PluginExecution pluginExecution) {
+        final Xpp3Dom pluginConfiguration = getOrCreatePluginConfiguration(pluginExecution);
+        final String skipAfterFailureCountProperty = "skipAfterFailureCount";
+        final Xpp3Dom count = pluginConfiguration.getChild(skipAfterFailureCountProperty);
+        if (count == null) {
+            final Xpp3Dom xpp3Dom = new Xpp3Dom(skipAfterFailureCountProperty);
+            final String skipAfterFailureCount = System.getProperty("surefire.skipAfterFailureCount");
+            final String skipAfterFailureCountValue = skipAfterFailureCount != null ? skipAfterFailureCount : "1";
```
bartoszmajsak: This can be oneliner `System.getProperty("surefire.skipAfterFailureCount", "1")`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -53,7 +54,7 @@ private void logExtensionDisableReason() {
         String reason = "Not Defined";
 
         if (configuration.isDisabled()) {
-            reason = "System Property `SMART_TESTING_DISABLE` is set.";
+            reason = "System Property " + SMART_TESTING_DISABLE + " is set.";
```
bartoszmajsak: Good catch!

lordofthejars:
just two general comments, the first one is if we should log that this property is set by ST and second can we add some unit test?

bartoszmajsak:
Would it make sense to have some tests for it?

bartoszmajsak:
I think it's important to document this behaviour once we have the final solution in place (so after the code review).

bartoszmajsak:
BTW `SkipAfterFailureCountTest.should_skip_all_remaining_tests_with_first_failure_when_ordering_enabled` is failing with

```
java.lang.AssertionError: 
Expecting:
 <4L>
to be greater than:
 <4L> 
```

when I run it in IDEA (maven seems to work though).

dipak-pawar:
@bartoszmajsak I ran it 10 times(5 times single test + 5 times entire class)  from IDE, not failed even single time. Am I missing something here?

bartoszmajsak:
Not sure, I'm just saying what happened to me (few times). Anyway, I think we should rework these tests.

bartoszmajsak:
I think we are confusing what `SKIPPED` means - at least from what one can understand reading the tests. The tests which are still part of the execution, but marked as `@Ignore` are treated as skipped. Tests which are never run have no status as there is no surefire report created for them. 

In addition, we assume certain execution of modules and assert based on it (`isEqualTo(4)`) - which is based on tests skipped by other, dependent modules. This is fine as long as we have fixed repo structure, but what if Maven/Surefire will change the way how tests are executed in the newer version? These tests will become flaky (if they already are not). As a side note: a similar thing happened when Java 8 changed reflection bits and all of the sudden the order of test methods executed by JUnit were not always the same (actually prior to JDK8 test methods were executed by JUnit in an order how they appeared in the source code).

Another observation I have is that test method names are not really reflecting the desired behaviour but rather a side effect (`should_skip_all_remaining_tests_with_first_failure_when_ordering_enabled`) or action which leads to certain behaviour (`should_override_skip_after_failure_count_system_property_when_ordering_enabled`).

Last, but not least, passing `Map` to internally mutate its state by project builder and then use it to assert certain things is not what I would think about. I proposed in the separated [PR](https://github.com/arquillian/smart-testing/pull/139) how we can change the way of returning `TestResult`s so that we keep the same flow but also have a way to see exact statuses for each and every test method.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -70,10 +70,9 @@ private void addSkipAfterFailureCount(PluginExecution pluginExecution) {
         final Xpp3Dom count = pluginConfiguration.getChild(skipAfterFailureCountProperty);
         if (count == null) {
             final Xpp3Dom xpp3Dom = new Xpp3Dom(skipAfterFailureCountProperty);
-            final String skipAfterFailureCount = System.getProperty("surefire.skipAfterFailureCount");
-            final String skipAfterFailureCountValue = skipAfterFailureCount != null ? skipAfterFailureCount : "1";
-
-            xpp3Dom.setValue(skipAfterFailureCountValue);
+            final String failureCount = System.getProperty("surefire.skipAfterFailureCount", "1");
```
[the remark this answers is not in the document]
dipak-pawar: In case of `selecting` mode. I think user always want to execute all `new/changed/affected` tests, as anyways he has to verify behavior for  `new/changed/affected` tests. so at the end of run he can identify only failed tests from entire test set. 

e.g. If we enable `skipAfterFailureCount`  option only for `selecting` mode then execution will exit after single failure. But in this case user has to fix tests for that single failure & run it again. It might be possible in this case he will get next failure in second run, then again he has to fix & run it again.

Instead of this we should execute all tests. so he can get status for entire `new/changed/affected` test sets. So he can fix only failed tests & run build again.

User can opt-out using `surefire.skipAfterFailureCount` to `0` in case of any mode. In case of `failsafe` I will change it not to apply this flag.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -70,10 +70,9 @@ private void addSkipAfterFailureCount(PluginExecution pluginExecution) {
...
+            xpp3Dom.setValue(failureCount);
+            logger.info("Setting `skipAfterFailureCount` to %s by Smart Testing", failureCount);
```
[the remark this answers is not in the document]
dipak-pawar: :+1:

dipak-pawar:
> I think we are confusing what `SKIPPED` means - at least from what one can understand reading the tests. The tests which are still part of the execution, but marked as `@Ignore` are treated as skipped. Tests which are never run have no status as there is no surefire report created for them. 

If test method is skipped because of `@ignore` then in  surefire report you'll get that method with status as skipped.
```xml
  <testcase name="should_merge_test_selection_with_different_strategy_for_same_class" classname="org.arquillian.smart.testing.TestSelectionTest" time="0">
    <skipped/>
  </testcase>
```

In case of `skipAfterFailureCount` it will show entire class as skipped with one method with name as className

```xml
<testcase name="org.arquillian.smart.testing.hub.storage.LocalChangeStorageTest" classname="org.arquillian.smart.testing.hub.storage.LocalChangeStorageTest" time="0">
    <skipped message="org.junit.runner.notification.StoppedByUserException"/>
  </testcase>

File: docs/configuration.adoc [hunk 5]
```diff
@@ -12,6 +12,11 @@ and then the rest.
 
 In order to define the mode use `smart.testing.mode` Java system property with either one.
 
+IMPORTANT: To get fast feedback loop *Smart testing* is setting `skipAfterFailureCount` to `1`. However you can override it by setting system
+property `surefire.skipAfterFailureCount` to required value. This behaviour is applicable only if user is using `surefire` plugin to execute
+test having smart testing enabled with `ordering` mode. This is not applicable for `selecting` mode because user can execute all `new/modified/affected`
+tests, identify all failed tests using results and rework on it at once.
```
bartoszmajsak: Small thing - maybe link to the official documentation about this flag in surefire website?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipAfterFailureCountTest.java
```diff
@@ -36,22 +36,26 @@ public void should_skip_all_remaining_tests_with_first_failure_when_ordering_ena
...
-
+            .run();
 
         // then
-        final Long skippedCount = resultCount.get(SKIPPED);
-        assertThat(skippedCount).isGreaterThan(4);
+        assertThat(testResults.testsWithStatuses(SKIPPED))
+            .extracting(TestResult::getClassName, TestResult::getTestMethod)
+            .contains(
+                tuple("org.jboss.arquillian.container.impl.client.container.ContainerLifecycleControllerTestCase", "org.jboss.arquillian.container.impl.client.container.ContainerLifecycleControllerTestCase"),
```
bartoszmajsak: Why class names and test methods are the same? This looks odd when reading the test.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -99,36 +98,10 @@ private void addSkipAfterFailureCount(PluginExecution pluginExecution) {
         }
 
         return testRunnerPluginConfigurations.stream()
-            .filter(
-                testRunnerPlugin -> !(testRunnerPlugin.getArtifactId().equals("maven-failsafe-plugin") && isSkipITs()))
+            .filter(testRunnerPlugin -> !(isFailsafePlugin(testRunnerPlugin) && isSkipITs()))
```
bartoszmajsak: Maybe extract filter condition to meaningful method - would be easier to understand the `why` of filtering

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipAfterFailureCountTest.java
dipak-pawar: This denotes it's skipped on class level.

bartoszmajsak: I know it because we discussed that. But does another reader will? I am not sure. We should make it more self-explanatory.

dipak-pawar: may be add some comments to make it more self-explanatory wdyt?

bartoszmajsak: Are comments the best we can do here?

dipak-pawar: Yeah I think so. I am thinking for following approach but it has drawbacks
If I removed verification of `TestMethod` name & keep verifying only `TestClass` with uniqueness, in future someone might change that Test with only one method, think of it's executing, then it's ignored due to `@Ignore`,it's not ignored due to `skipAfterFailureCount`, so for this case test should not pass, but it's passing only because of validation of only `className`

bartoszmajsak: I am not sure if there is a "future" in this fixed repo, as every test controls the source code of the target project precise enough.

bartoszmajsak: So what we could do here instead is to introduce custom assertion saying `assertThat(testResults.testsWithStatuses(SKIPPED)).hasSkippedClasses("...")` and within it we could have this detailed check with tuples and a comment if needed about why it is expected to be like that. So we stick to strict verification but from the use of assertions and by reading it, it will become clear what are the expectations

MatousJobanek:
I'm closing this PR as setting the property `skipAfterFailureCount` should be done by a user (not automatically). For more information see this comment https://github.com/arquillian/smart-testing/issues/96#issuecomment-329310444 and also this PR https://github.com/arquillian/smart-testing/pull/206 that adds the explanation to the documentation.

PR #131, opened 2017-09-07
Introduces Debug mode

### LINKED ISSUE #91: Debug information
We have to provide to the end users ability to have diagnostic information. This will not only help them trying to understand what is going on, but also would be very important input for us in case issues will arise.

The goal of this story would be to design a solution with following aspects:

  * how this should be enabled
  * where and how it should be presented
  * what should be part of diagnostic information

One bit which in particular would be useful an effective POM we are using after Maven Extension configures all the required bits. This is implemented now in a very ad-hoc manner as [`showPom` method](https://github.com/arquillian/smart-testing/blob/0b961dfba1dfc71ecc55be0b74feed6c5a517ea3/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java#L32).

Thoughts and ideas about the topic more than welcome.

hemanik:
A debug mode is presented to the user for diagnostic purposes. To use the smart testing extension with the debug mode set system property `smart.testing.debug`. 

The debug mode presents the following information:
- the modifiedPom.xml used for smart testing
- the system properties
- the other ST debug logs
- the selected tests.

At the moment the debug log is only present in the console.

[Sample Debug Mode Log](https://gist.github.com/hemanik/9ef4144b6cd158fab92afd0f5ed359c3)

- introduced a system property for enabling `debug mode` with smart testing.
- storage of modifiedPom.xml in the target directory of the sub-modules.
- tests for the debug mode.

Fixes #91


bartoszmajsak:
> `System Properties Applied: {new=Dependency {groupId=org.arquillian.smart.testing, artifactId=smart-testing-strategy-changed, version=0.0.1-SNAPSHOT, type=jar}, failed=Dependency {groupId=org.arquillian.smart.testing, artifactId=smart-testing-strategy-failed, version=0.0.1-SNAPSHOT, type=jar}, affected=Dependency {groupId=org.arquillian.smart.testing, artifactId=smart-testing-strategy-affected, version=0.0.1-SNAPSHOT, type=jar}, changed=Dependency {groupId=org.arquillian.smart.testing, artifactId=smart-testing-strategy-changed, version=0.0.1-SNAPSHOT, type=jar}} `

Are those really system properties? Maybe something more contextual here?

hemanik:
Todo for this feature

- [ ] also capture the debug log in a file stored in `target` directory of the root of project. 

- [ ] update the documentation.

- [ ] verify and improve the content like test selections.

core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -12,6 +14,16 @@
 
     private Logger(java.util.logging.Logger logger) {
         this.jul = logger;
+        setLogLevel();
+    }
+
+    private void setLogLevel() {
```
bartoszmajsak: `enableDebugLogLevel`?

core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -145,6 +172,10 @@ public void finest(String msg, Object... args) {
         jul.finest(getFormattedMsg(msg, args));
     }
 
+    public Boolean isDebugEnabled() {
+        return Boolean.valueOf(System.getProperty(SMART_TESTING_DEBUG, "false"));
```
bartoszmajsak: As @MatousJobanek pointed out in another PR, if a property is not set it will return `null` `Boolean.valueOf` will make it `false`. But I believe this verbosity here clearly explains what we expect as a default

core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -86,6 +98,21 @@ public void warn(String msg, Object... args) {
...
+     * "DEBUG: Smart-Testing - ", if debug mode is enabled.
+     *
+     * @param msg
+     *     The string message (or a key in the message catalog)
+     * @param args
+     *     arguments to the message
+     */
+    public void debug(String msg, Object... args) {
+        if (isDebugEnabled()) {
+            System.out.println(getFormattedMsg("DEBUG", msg, args));
```
lordofthejars: Shouldn't be this a call to `logger` instead of `sysout`?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ModifiedPomExporter.java
```diff
@@ -0,0 +1,38 @@
...
+            throw new RuntimeException("Failed writing updated pom file: " + model.getPomFile().getAbsolutePath(), e);
+        }
+    }
+
+    static void copyModifiedPomInTarget(StringWriter modifiedPom, File project) throws IOException {
+        Path path = Paths.get(project.toPath() + File.separator + "target", "smart-testing");
+        if (!Files.exists(path)) {
+            Files.createDirectories(path);
+        }
+        final Path target = Paths.get(path + File.separator + "modifiedPom.xml");
```
bartoszmajsak: Let's name it `smart-testing-pom.xml`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java
```diff
@@ -26,6 +29,7 @@ private void addStrategies(Model model) {
         final StrategyDependencyResolver strategyDependencyResolver = new StrategyDependencyResolver();
         model.addDependency(smartTestingProviderDependency());
         final Map<String, Dependency> dependencies = strategyDependencyResolver.resolveDependencies();
+        logger.debug("System Properties Applied: %s ", dependencies);
```
bartoszmajsak: Maybe `System Properties Applied: %s` -> `Added Dependencies: %s`?

core/src/main/java/org/arquillian/smart/testing/Logger.java
bartoszmajsak: This is our logger and I remember @MatousJobanek did that by purpose to go to sysout too.

core/src/main/java/org/arquillian/smart/testing/Logger.java
bartoszmajsak: If I remember correctly we also wanted to have it enabled when user run the build with `mvn -X`

lordofthejars: Yes but I think it is better as you said to have in this way here to be clear what is the default.

bartoszmajsak: Yeah, default is fine. But what about `-X` flag?

lordofthejars: Do you think is possible to get it from there?

hemanik: @bartoszmajsak, it doesn't work for the `-X` flag at the moment. `-X` is not set as system or user property. It is a maven option and I am not sure how I can get this value from the build.

In the test-bed, we can set it with the `builderOptions`, but I don't know how to retrieve it from the 
maven configurer.

bartoszmajsak: After a quick look into Maven Core source code I can think about one way of doing it... Which in fact  reveals that we can rethink our logging approach. In the `SmartTestingMavenConfigurer` you can inject Maven logger as a field:

```
@Requirement
private org.codehaus.plexus.logging.Logger log;
```

and this gives you access to `log.isDebugEnabled()` which is true when `-X` is set.

lordofthejars: The thing is that then we log `core` elements with one logger and maven things with another,  not bad but just noticing it

bartoszmajsak: The question is however - how to propagate this information further down the execution.

bartoszmajsak: > The thing is that then we log core elements with one logger and maven things with another, not bad but just noticing it

Sure, rethinking part is open for discussion. The access to this information (if we are in debug mode in Maven) is another thing.

lordofthejars: Maybe we can just use this logger instance to detect if `-X` is set

bartoszmajsak: On the other hand, this got me thinking. As `-X` is a command line option which is a JVM thing.

```java
RuntimeMXBean runtimeMxBean = ManagementFactory.getRuntimeMXBean();
List<String> arguments = runtimeMxBean.getInputArguments()
```
and this list contains `-Xdebug` element. That's all what we need.

bartoszmajsak: Ok, it seems that `-X` != `-Xdebug`. Makes me wonder how Mave reads this flag. @MatousJobanek any idea?

MatousJobanek: The maven `org.codehaus.plexus.logging.Logger` is available for the plugins, but it is not available nor used in any `SurefireProvider` implementation. In the surefire providers, they are using their own implementation (based on the information if the execution is in a fork or not). But in any case, they log on stdout. This is also why I have implemented it using stdout and stderr.
There is a way to get the instance of the Surefire logger from the `ProviderParameters`, but this will work only for maven environment, not for IDEs.

lordofthejars: We need an instance of `CliRequest`

MatousJobanek:
@lordofthejars @bartoszmajsak @hemanik 
As for the `-X` retrieval. In the Maven extension, we should probably use the Maven logger provided by DI:
```
@Requirement
private Logger logger;
```
this logger also provides information about the set level: 
`logger.getThreshold()`

On the side of Surefire provider, we can get the information from `ProviderProperties`:
```
mainCliOptions = {ArrayList@579}  size = 6
 0 = {CommandLineOption@892} "LOGGING_LEVEL_ERROR"
 1 = {CommandLineOption@893} "LOGGING_LEVEL_WARN"
 2 = {CommandLineOption@894} "LOGGING_LEVEL_INFO"
 3 = {CommandLineOption@895} "LOGGING_LEVEL_DEBUG"
 4 = {CommandLineOption@896} "REACTOR_FAIL_FAST"
 5 = {CommandLineOption@897} "SHOW_ERRORS"
```

hemanik:
Thanks @MatousJobanek, we can get the logging level for debug from `session.getRequest().getLoggingLevel()` if this returns 0, debug is set.

MatousJobanek:
Yes @hemanik that should work as well - great find. 
What I also wanted to point out (by my comment) is that in the maven extension we should use the logger provided by Maven - just to keep the format that is used in Maven. WDYT?

hemanik:
@MatousJobanek, last night I updated the the PR with `-X` flag using `session.getRequest().getLoggingLevel()` and it works fine.

> What I also wanted to point out (by my comment) is that in the maven extension we should use the logger provided by Maven - just to keep the format that is used in Maven. WDYT?

I personally like the idea of keeping the format consistent with that used in Maven, but then I am not sure how we can integrate it with our smart testing logger. Do you propose that we replace smart testing logger with this one ? 
And then how do we pass an instance of this across the code base, if we keep both the smart testing and maven logger ?

bartoszmajsak:
Let's open separated issue for logger thing. @MatousJobanek could you do it?

@hemani do we also print out some debug info on the provider side? if so we have to make sure that we are able to handle `-X` on this side too, as @MatousJobanek hinted https://github.com/arquillian/smart-testing/pull/131#issuecomment-327835513

core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -12,6 +13,7 @@
 
     private Logger(java.util.logging.Logger logger) {
         this.jul = logger;
+        new LoggerConfigurator(jul);
```
bartoszmajsak: This looks quite strange. On one side we create an instance of `LoggerConfigurator` but then refer to it through global state / statics?

core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -86,6 +88,21 @@ public void warn(String msg, Object... args) {
     }
 
     /**
+     * Will format the given message with the given arguments and prints it on standard output with the prefix:
```
bartoszmajsak: I am not sure if we really need this javadoc here. WDYT @MatousJobanek?

core/src/main/java/org/arquillian/smart/testing/LoggerConfigurator.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private String getLogFile() {
+        if (debugLogFile == null || debugLogFile.isEmpty()) {
+            return DEFAULT_SMART_TESTING_LOG_FILE;
+        }
+        return debugLogFile;
+    }
+
+    private static String getDefaultLogFile() {
+        final File logDirectory = new File("target/smart-testing-debug.log");
```
bartoszmajsak: I would rather expect to have one log file for the whole build and stored in the root of the project rather than several files scattered around in `target`s.

docs/usage.adoc
```diff
@@ -46,6 +46,13 @@ If you only want to use smart testing on `surefire` or `failsafe` plugin definit
 
 To do it you just need to set `smart.testing.apply.to` with either `surefire` or `failsafe` to just enable in one of them.
 
+=== Debug Mode
+
+In order to get diagnostic information related to Smart testing extension, you can enable the debug mode by either
+setting the system property `smart.testing.debug` or by setting the mvn debug output `(-X)` flag along with smart testing.
+
+Debug mode by default writes smart testing debug level logs to the console.
```
bartoszmajsak: What about the poms we are creating? I think it's the most important information to share in the documentation.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/DebugModeSmartTestingFunctionalTest.java
```diff
@@ -0,0 +1,113 @@
...
+        final TestResults actualTestResults = projectBuilder
+                .options()
+                    .withSystemProperties(COMMIT, "HEAD", PREVIOUS_COMMIT, "HEAD~", SMART_TESTING_DEBUG, "true", "skipITs", "true")
+                .configure()
+            .run();
+
+        // then
+        String projectMavenLog = project.getMavenLog();
+        assertThat(projectMavenLog).contains("DEBUG: Smart-Testing");
+        assertThatAllBuiltSubmodulesHaveReportsIncluded(projectBuilder.getBuiltProject(), "smart-testing/smart-testing-pom.xml");
```
bartoszmajsak: These are not really reports from the functionality point of view - these are alternated poms.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ModifiedPomExporter.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.maven.model.io.xpp3.MavenXpp3Writer;
+import org.arquillian.smart.testing.Logger;
+
+class ModifiedPomExporter {
+
+    static Logger logger = Logger.getLogger(ModifiedPomExporter.class);
+
+    private static String SMART_TESTING_POM_FILE = "smart-testing-pom.xml";
+
+    static void showPom(Model model) {
```
bartoszmajsak: Why is it still called `showPom` - it was called this way in the first version because it was printing it out to the console among other logs, but this name doesn't really reflect what is happening under the hood now.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java
```diff
@@ -6,12 +6,15 @@
...
 import org.apache.maven.model.Plugin;
 import org.arquillian.smart.testing.Configuration;
+import org.arquillian.smart.testing.Logger;
 import org.arquillian.smart.testing.mvn.ext.ApplicablePlugins;
 
 public class DependencyResolver {
 
     private final Configuration configuration;
 
+    private Logger logger = Logger.getLogger(DependencyResolver.class);
```
bartoszmajsak: Why do we need this logger here if we don't log anything from this class anymore?

core/src/main/java/org/arquillian/smart/testing/Logger.java
MatousJobanek: I'm fine with the javadoc - it describes what it does.
What I can see as a problem is that the log is logged only when `smart.testing.debug=true` is used. This can be confusing as `FINE` and `FINEST` are logged anyway and the visibility can be changed by the JUL log level. IMHO `DEBUG` should be higher than `FINE` or `FINEST`. Our problem is that we mix two or three things together. When we have the `DEBUG` implemented, then I would get rid of JUL and the `FINE` and `FINEST` WDYT?

core/src/main/java/org/arquillian/smart/testing/LoggerConfigurator.java
bartoszmajsak: Le'ts park it as part of this PR

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/DebugModeSmartTestingFunctionalTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+        project
+            .applyAsCommits("Single method body modification - sysout",
+                "Inlined variable in a method");
+
+        // when
+        ProjectBuilder projectBuilder = project.build("config/impl-base");
+        final TestResults actualTestResults = projectBuilder
+                .options()
+                    .withSystemProperties(COMMIT, "HEAD", PREVIOUS_COMMIT, "HEAD~", SMART_TESTING_DEBUG, "true", "skipITs", "true")
```
MatousJobanek: Are there any integration tests when we use: `"skipITs", "true"`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ModifiedPomExporter.java
```diff
@@ -0,0 +1,40 @@
...
+        try (StringWriter pomOut = new StringWriter()) {
+            new MavenXpp3Writer().write(pomOut, model);
+            copyModifiedPomInTarget(pomOut, model.getProjectDirectory());
+        } catch (IOException e) {
+            throw new RuntimeException("Failed writing updated pom file: " + model.getPomFile().getAbsolutePath(), e);
+        }
+    }
+
+    static void copyModifiedPomInTarget(StringWriter modifiedPom, File project) throws IOException {
```
MatousJobanek: This method also doesn't copy anything, it just prints the content of the modified pom to the file. It hasn't been stored anywhere, right?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -69,6 +80,12 @@ public void afterSessionEnd(MavenSession session) throws MavenExecutionException
             return;
         }
 
+        if (enableDebugLogLevel()) {
```
MatousJobanek: having `enableMavenDebugLogLevel(true)` for setting the level up and `enableDebugLogLevel()` for figuring out if it is set is very confusing to my taste. Please, change the second one - something like: `isDebugLogLevelEnabled`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -36,6 +40,13 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
 
         configuration = Configuration.load();
 
+        if (session.getRequest().getLoggingLevel() == 0) {
+            enableMavenDebugLogLevel(true);
+            logger.debug("Maven build run in debug log level.");
```
MatousJobanek: Do we really need this information?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -36,6 +40,13 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
...
+        }
+
+        logger.debug("Applied user properties: %s", session.getUserProperties());
```
MatousJobanek: Same for this line - I don't really think that we need this. This is nothing related only to smart testing - it contains all properties in general. Instead of this, you could log only the "smart testing" properties specified by user

hemanik: This returns all the properties set by the user in the mvn request. I feel it is useful to have scm properties and other properties set in the request like `skipTests` etc. that are not included in "smart testing" properties but influence our extension.
Eg: `DEBUG: Smart-Testing - Applied user properties: {smart.testing.version=0.0.1-SNAPSHOT, surefire.exitTimeout=-1, surefire.timeout=0, smart.testing.mode=ordering, skipITs=true, smart.testing=affected, scm.range.tail=HEAD~, skipTests=false, scm.range.head=HEAD, smart.testing.debug=true}
`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
bartoszmajsak: -1 from my side. I think we should remove it

core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -1,19 +1,22 @@
...
 public class Logger {
 
     private final java.util.logging.Logger jul;
 
     private static final String PREFIX = "%s: Smart-Testing - ";
 
+    private boolean mavenDebugLogLevel = false;
+
     private Logger(java.util.logging.Logger logger) {
         this.jul = logger;
```
bartoszmajsak: If everything goes to `System.out` now why do we still need instance of the logger?

core/src/main/resources/logging.properties
```diff
@@ -0,0 +1,14 @@
+handlers= java.util.logging.ConsoleHandler
```
bartoszmajsak: Not needed anymore.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ModifiedPomExporter.java
```diff
@@ -0,0 +1,40 @@
...
+    static void exportModifiedPom(Model model) {
+        try (StringWriter pomOut = new StringWriter()) {
+            new MavenXpp3Writer().write(pomOut, model);
+            writeModifiedPomToTarget(pomOut, model.getProjectDirectory());
+        } catch (IOException e) {
+            throw new RuntimeException("Failed writing updated pom file: " + model.getPomFile().getAbsolutePath(), e);
+        }
+    }
+
+    static void writeModifiedPomToTarget(StringWriter modifiedPom, File project) throws IOException {
```
bartoszmajsak: This can be `private`, right?

core/src/main/java/org/arquillian/smart/testing/Logger.java
```diff
@@ -1,67 +1,25 @@
...
-     *     The string message (or a key in the message catalog)
-     * @param args
-     *     arguments to the message
-     */
-    public void log(Level level, String msg, Object... args) {
-        if (jul.isLoggable(level)) {
-            jul.log(level, getFormattedMsg(msg, args));
+    public static Logger getLogger() {
+        if (logger == null) {
+            logger = new Logger();
```
bartoszmajsak: Let's just return new instance for each and every call to this factory method. We don't have to share the instance across the classes.

bartoszmajsak:
I will push few small changes and then merge with `master`.

PR #134, opened 2017-09-08
The surefire reports are automatically copied in our maven extension

### LINKED ISSUE #47: Seamlessly store test results from the build execution
`failed` strategy relies on historical build execution. At this point we only support local experience. It has an assumption that reports from previous test run are stored in the `.report` folder. We don't have however any automated way of getting those reports there other than user doing it manually.

This task should automate this process in the `surefire-provider`. Every test result file should be moved to this folder.  We should also purge the folder **after** we gather test results data from the previous build, as we will write new ones. I'm open here to other ideas though, as this is suboptimal. Maybe subfolder per build would be better?

We should also implement it such that plugging in publisher of test reports to the "hub" would be a matter of adding it as a dependency. Good candidate for SPI?

MatousJobanek:
Previously, when the `failed` strategy was used, a user had to move/copy the surefire reports from previous test execution to the `${project.directory}/.reports` directory. 
In this PR the "copy" process is automatically done in maven extension when the project is read. Files that are copied are only those that end with `.xml` suffix - as we are interested only in XML files. When the build is finished, all `${project.directory}/.reports` are automatically removed.

Fixes #47


mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -88,7 +100,62 @@ private void calculateChanges() {
...
-        session.getAllProjects().forEach(mavenProject -> mavenProjectConfigurator.configureTestRunner(mavenProject.getModel()));
+        session.getAllProjects().forEach(mavenProject -> {
+            mavenProjectConfigurator.configureTestRunner(mavenProject.getModel());
+            if (isFailedStrategyUsed()) {
+                copySurefireReports(mavenProject.getModel());
+            }
+        });
+    }
+
+    private void purgeReports(MavenSession session) {
```
bartoszmajsak: Ultimately let's move this logic to its own class.

bartoszmajsak:
I don't think we need more than "status" label, the rest is in the related issue itself. If there is no issue then maybe :)

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
MatousJobanek: Sure, I'll move all the logic related to surefire reports to a separated class `SurefireReportStorage` - I didn't make it on Friday :-)

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/failed/ReportsVerifier.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import org.arquillian.smart.testing.ftest.testbed.project.Project;
+import org.assertj.core.api.Assertions;
+
+import static org.arquillian.smart.testing.spi.TestResult.TEMP_REPORT_DIR;
+
+public class ReportsVerifier {
+
+    static void assertNoReportsDirectoryPresent(Project project) throws IOException {
```
dipak-pawar: How about `reportsDirectoryDoesNotExists`?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -86,9 +91,18 @@ private void calculateChanges() {
         changeStorage.store(changes);
     }
 
-    private void configureExtension(MavenSession session, Configuration configuration) {
+    private void configureExtensionPrepareReports(MavenSession session, Configuration configuration) {
```
dipak-pawar: Can we give better name? This name is ambiguous

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorageTest.java
```diff
@@ -0,0 +1,110 @@
...
+        // when
+        assertThat(reportsDir).exists();
+        SurefireReportStorage.purgeReports(mavenSession);
+
+        // then
+        assertThat(reportsDir).doesNotExist();
+    }
+
+    @Test
+    public void should_not_create_nor_copy_anything() throws IOException {
```
dipak-pawar: How about `should_not_create_nor_copy_anything_if_surefire_reports_dir_does_not_exist`?

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorageTest.java
```diff
@@ -0,0 +1,110 @@
...
+        expectedReports.put("second-report.xml", createDummyFile(surefireReportsDir, "second-report.xml"));
+        expectedReports.put("third-report.xml", createDummyFile(surefireReportsDir, "third-report.xml"));
+        createDummyFile(surefireReportsDir, "not-xml-report");
+        return expectedReports;
+    }
+
+    private long createDummyFile(File directory, String fileName) throws IOException {
+        File file = new File(directory, fileName);
+        RandomAccessFile raf = new RandomAccessFile(file, "rw");
+        raf.setLength(fileName.length());
```
dipak-pawar: Why do we need to `setLength` here?

dipak-pawar:
Greak work @MatousJobanek

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/failed/ReportsVerifier.java
bartoszmajsak: Let's leave it like it is for now. I will create a separated issue to clean up the assertions using `assertj` as we brought similar static methods in other tests.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
bartoszmajsak: What would you suggest @dipak-pawar ?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorage.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private static void copyReportsDirectory(Model model, File surefireReportsDir) {
+        File reportsDir = new File(model.getProjectDirectory(), TEMP_REPORT_DIR);
+        logger.fine("Copying surefire report directory from [%s] to [%s]", surefireReportsDir, reportsDir);
+
+        if (!reportsDir.exists()) {
+            try {
+                Files.createDirectory(reportsDir.toPath());
+            } catch (IOException e) {
+                logger.severe("There occurred an error when the directory %s was being created: %s", reportsDir,
```
bartoszmajsak: Shouldn't we fail here too? Especially that later we do `.forEach(file -> copyReportFile(file, reportsDir))`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorage.java
```diff
@@ -0,0 +1,77 @@
...
+
+        Arrays.stream(surefireReportsDir.listFiles())
+            .filter(file -> file.isFile() && file.getName().endsWith(".xml"))
+            .forEach(file -> copyReportFile(file, reportsDir));
+    }
+
+    private static void copyReportFile(File src, File destDir) {
+        File destination = new File(destDir, src.getName());
+        try {
+            FileUtils.copyFile(src, destination);
```
bartoszmajsak: I know we have it available on the cp because it's maven extension but can we set us free from those utils which might get change in Mvn as I assume it's their internal thing?

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorageTest.java
```diff
@@ -0,0 +1,110 @@
...
+        // given
+        SurefireReportStorage.copySurefireReports(project);
+        MavenSession mavenSession = mock(MavenSession.class);
+        MavenProject mavenProject = mock(MavenProject.class);
+        when(mavenProject.getModel()).thenReturn(project);
+        when(mavenSession.getAllProjects()).thenReturn(Arrays.asList(new MavenProject[] {mavenProject}));
+        File reportsDir = new File(projectDir, TEMP_REPORT_DIR);
+
+        // when
+        assertThat(reportsDir).exists();
```
bartoszmajsak: Do we really need to assert here?

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorageTest.java
bartoszmajsak: What would you suggest @dipak-pawar ?

dipak-pawar: I am thinking appending/adding some text instead of setting length for a file which doesn't have any content inside.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
bartoszmajsak: I think it can stay like it was before. It's clear enough from the content what is going on.

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorageTest.java
bartoszmajsak: I'm just wondering why we need RAF at all here?

MatousJobanek: The idea is creating dummy but not empty file by using RAF and setting the length of the file. Yeah, I could put the string there, that is another option. From my point of view, this seemed to be faster, simpler and easier, that's why I used this approach...

bartoszmajsak: Fair enough, at the end of the day, it's testing dummy so we can live with that :)

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
MatousJobanek: Ok, if you're fine with the original name `configureExtension` also for copying the reports, why not. 
For me, configure extension was related to modifying pom.xml files and similar, not preparing reports - this is why I changed the name. But I agree that the name is weird - this is always one of my weaknesses :-D.

bartoszmajsak: I think it was ambiguous already. But that is fine - this method takes care of configuring extension and within it we have detailed steps explaining what really is the "configuration" all about.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorage.java
MatousJobanek: I wouldn't fail, but return - good catch, I missed that.

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorageTest.java
MatousJobanek: argh... zombie

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorage.java
MatousJobanek: Yeah, I didn't want to add another dependency such as `commons-io` as this was already on CP and I was too lazy to implement it on my own, but I didn't check which scope is the dependency with.

bartoszmajsak:
Any idea why Travis is not building it?

docs/configuration.adoc
```diff
@@ -108,11 +108,7 @@ WARNING: At this moment, this strategy does not work with Java 9.
 
 `Failed` strategy just gets all tests that failed from previous executions and mark them as *important* tests to run first (_ordering_) or not filtered (_selecting_).
 
-This strategy uses the _JUnit_ XML _de-facto_ https://github.com/apache/maven-surefire/blob/master/maven-surefire-plugin/src/site/resources/xsd/surefire-test-report.xsd[report] for reading past executions.
+This strategy uses the _JUnit_ XML _de-facto_ https://github.com/apache/maven-surefire/blob/master/maven-surefire-plugin/src/site/resources/xsd/surefire-test-report.xsd[report] for reading past executions. All reports are automatically copied by the maven extension to a temp directory `${project.directory}/.reports` and when the build is finished the directory is removed.
```
bartoszmajsak: Let's make it `.smart-testing/execution/reports`

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/failed/ReportsVerifier.java
bartoszmajsak: Related issue #140

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SurefireReportStorage.java
```diff
@@ -0,0 +1,92 @@
...
+    private static final Logger logger = Logger.getLogger(SurefireReportStorage.class);
+
+    static void copySurefireReports(Model model) {
+        Build build = model.getBuild();
+        if (build != null && build.getDirectory() != null) {
+            File targetDir = new File(build.getDirectory());
+            if (targetDir.exists() && targetDir.isDirectory()) {
+                File[] surefireReports =
+                    targetDir.listFiles(file -> file.isDirectory() && SUREFIRE_REPORTS_DIR_NAME.equals(file.getName()));
+                if (surefireReports.length > 0) {
```
bartoszmajsak: Can it be more than 1?

MatousJobanek: well, probably not. But it doesn't matter I guess - I just need to check if there is any, which means that the array is not empty

docs/configuration.adoc
MatousJobanek: yeah, my memory is not good :-(

MatousJobanek: btw. I'd like to create a separated issue/PR for this... to add also the file containing changes there.
I'll do it today morning

PR #145, opened 2017-09-13
Introduced LocalChanges class that takes care of all files stored locally

### LINKED ISSUE #143: All working files (surefire reports, changes) should be placed inside one directory
Currently, we have several dot files stored in the main project directory. All of them should be placed inside one dot directory: `.smart-testing`. The directory should have subdirectories based on the type of the file. For working files intended to be used during test execution, there should be subdirectory `execution`.  For reporting information `reporting`

MatousJobanek:
The directory, where all files are all files are stored during test execution, is `${project.directory}/.smart-testing` and contains two subdirectories - `execution` and `reporting`.

* `execution` directory is intended to be used during test execution for "working files" such as surefire reports and file containing scm changes. 

* `reporting` directory should contain files/directories related to reports and statistics. These files should be also stored when the build is finished.

When the build is finished, the `reporting` directory is copied to `${project.directory}/target/smart-testing/` directory and the whole  `${project.directory}/.smart-testing` is removed.

Fixes #143


core/src/main/java/org/arquillian/smart/testing/hub/storage/local/LocalChangeStorage.java
```diff
@@ -0,0 +1,75 @@
...
+        this.currentDirectory = ".";
+    }
+
+    public LocalChangeStorage(String currentDirectory) {
+        this.currentDirectory = currentDirectory;
+    }
+
+    @Override
+    public void store(Collection<Change> changes) {
+        StringBuffer fileContent = new StringBuffer();
```
lordofthejars: Stupid thing but you can use `StringBuilder` it is fastest to use rather than `StringBuffer`

core/src/main/java/org/arquillian/smart/testing/hub/storage/local/LocalChangeStorage.java
```diff
@@ -0,0 +1,75 @@
...
+        changes.forEach(change -> fileContent.append(change.write()).append("\n"));
```
lordofthejars: Maybe change to `System.lineSeparator()`

PR #148, opened 2017-09-14
feat(docs): Creates Asciidoctor constant extension

lordofthejars:

Creates Asciidoctor constant extension. This allows you to add source files setting the named value of the constant. Then this extension finds the value and sets in docs. For example:

```
// In Java source file
public static final String DEFAULT_COLOR = "blue";

// How to refer to value from docs
const::src/main/java/org/superbiz/Color.java[name="DEFAULT_COLOR"]
```

Also you can set an special tag as comment so you can rename the variable name in the future without worrying about the docs

```
// const::DEFAULT_COLOR[]
public static final STRING MY_DEFAULT_COLOR = "blue";

const::src/main/java/org/superbiz/Color.java[tag="DEFAULT_COLOR"]
```

- Creates an extension to deal with constants

#### Additional comments

This extension now just works with String constants, for now it is enough but it is something that when needed should be added.


bartoszmajsak:
Awesome! Can we have it in `docs/asciidoctor/lib` though?

bartoszmajsak:
Could you already apply it on our docs?

lordofthejars:
> Could you already apply it on our docs? 

I was waiting for merge and then update the docs in different issue or do it in different commit when extension has been reviewed

> Can we have it in docs/asciidoctor/lib though?

I thought the same but I understand that `docs` folder is just for documentation and not for "code" but I have no strong opinion about it

bartoszmajsak:
So maybe `.asciidoctor` to align with other folders we have which clearly indicate the hold some configuration for toolset we have in place?

lordofthejars:
Sounds good

PR #150, opened 2017-09-14
Update documentation to use constant extension

### LINKED ISSUE #149: Update documentation to use const extension
Update documentation to use const extension.

Related to https://github.com/arquillian/smart-testing/pull/148

lordofthejars:

Update documentation to use Asciidoctor constant extension instead of hardcoding the constants on documentation.

Fixes #149

### DIFF
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -8,9 +8,11 @@ to run first the important tests and then the rest. This is know as `ordering` m
 `ordering`:: `ordering` mode as its name suggests orders the test execution plan so important tests are executed first
 and then the rest.
 
-`selecting`:: `selecting` mode just selects the important tests and execute them, skipping the rest of the tests. This is the default mode.
+`selecting`:: `selecting` mode just selects the important tests and execute them, skipping the rest of the tests.
 
-In order to define the mode use `smart.testing.mode` Java system property with either one.
+The `const:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="DEFAULT_MODE"]` is the **default** mode.
+
+In order to define the mode use `const:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING_MODE"]` Java system property with either one.
 
 === Strategies
 
@@ -20,7 +22,7 @@ But how do we know which tests are important and which ones not?
 There are several strategies that you can choose from which determine what are the *important* tests.
 Currently we have following strategies in place: `new`, `changed`, `affected` and `failed`.
 
-To set them you need to set Java system property `smart.testing` to one or more strategies in comma-separated value form.
+To set them you need to set Java system property `const:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING"]` to one or more strategies in comma-separated value form.
 
 Subsequent sections dive deeper into each and every one of them.
 
@@ -34,16 +36,16 @@ are marked as important ones.
 
 This strategy can be configured in several ways through Java system properties:
 
-* `scm.range.head` and `scm.range.tail` are used to set the range of commits you want to inspect for new tests.
+* `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="COMMIT"]` and `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]` are used to set the range of commits you want to inspect for new tests.
 The values can be commit ids or using `HEAD` notation.
-For example: `-Dscm.range.head=HEAD -Dscm.range.tail=HEAD~`
+For example: `-Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="COMMIT"]=HEAD -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]=HEAD~`
 By default if not specified, default value is `HEAD`.
-* `scm.last.changes` can be used to set the number of commits from `HEAD` that you want to inspect.
-For example `-Dscm.last.changes=3` will be equivalent as `-Dscm.range.head=HEAD -Dscm.range.tail=HEAD~~`.
+* `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]` can be used to set the number of commits from `HEAD` that you want to inspect.
+For example `-Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]=3` will be equivalent as `-Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="COMMIT"]=HEAD -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]=HEAD~~`.
 
 TIP: Those properties might be in particular handy when used in the CI builds.
 For example in Jenkins when using https://wiki.jenkins.io/display/JENKINS/Git+Plugin[Git Plugin] you can configure your build as follows:
-`$ mvn clean verify -Dsmart.testing=new, affected -Dscm.range.head=GIT_COMMIT -Dscm.range.tail=GIT_PREVIOUS_COMMIT`
+`$ mvn clean verify -Dconst:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING"]=new, affected -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="COMMIT"]=GIT_COMMIT -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="PREVIOUS_COMMIT"]=GIT_PREVIOUS_COMMIT`
 
 Currently not committed changes (those that are _added_ and _untracked_) are considered as important tests as well.
 This effectively means that if you don't specify any commit range, only these information is considered.
@@ -55,7 +57,7 @@ instead of *new* ones.
 
 In this strategy not committed changes (those that are _modified_ or _changed_) are considered as important tests as well.
 
-TIP: Usually `new` and `changed` strategies are used together `-Dsmart.testing=new, changed`.
+TIP: Usually `new` and `changed` strategies are used together `-Dconst:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING"]=new, changed`.
 
 ==== Affected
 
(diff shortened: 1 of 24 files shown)


docs/configuration.adoc
```diff
@@ -34,16 +36,16 @@ are marked as important ones.
 
 This strategy can be configured in several ways through Java system properties:
 
-* `scm.range.head` and `scm.range.tail` are used to set the range of commits you want to inspect for new tests.
+* `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="COMMIT"]` and `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]` are used to set the range of commits you want to inspect for new tests.
```
MatousJobanek: the second property seems to be wrong: `ScmRunnerProperties#LAST_COMMITS = "scm.last.changes";`
there should used be `PREVIOUS_COMMIT` variable

MatousJobanek: https://github.com/arquillian/smart-testing/blob/master/core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java

docs/configuration.adoc
```diff
@@ -34,16 +36,16 @@ are marked as important ones.
...
 The values can be commit ids or using `HEAD` notation.
-For example: `-Dscm.range.head=HEAD -Dscm.range.tail=HEAD~`
+For example: `-Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="COMMIT"]=HEAD -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]=HEAD~`
 By default if not specified, default value is `HEAD`.
-* `scm.last.changes` can be used to set the number of commits from `HEAD` that you want to inspect.
-For example `-Dscm.last.changes=3` will be equivalent as `-Dscm.range.head=HEAD -Dscm.range.tail=HEAD~~`.
+* `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]` can be used to set the number of commits from `HEAD` that you want to inspect.
+For example `-Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]=3` will be equivalent as `-Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="COMMIT"]=HEAD -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]=HEAD~~`.
```
MatousJobanek: Here is the same mistake. 
Maybe, it would be worth changing the variable names to something that reflects the content. eg:
```java
    public static final String SCM_RANGE_TAIL = "scm.range.tail";
    public static final String SCM_RANGE_HEAD = "scm.range.head";
    public static final String SCM_LAST_CHANGES = "scm.last.changes";
```
instead of:
```java
    public static final String PREVIOUS_COMMIT = "scm.range.tail";
    public static final String COMMIT = "scm.range.head";
    public static final String LAST_COMMITS = "scm.last.changes";
```

lordofthejars: I have not checked yet but you mean that the value is correct, what happened is that you get confused because of the variables names right?

docs/configuration.adoc
lordofthejars: Fixed.

docs/configuration.adoc
```diff
@@ -20,7 +22,7 @@ But how do we know which tests are important and which ones not?
 There are several strategies that you can choose from which determine what are the *important* tests.
 Currently we have following strategies in place: `new`, `changed`, `affected` and `failed`.
 
-To set them you need to set Java system property `smart.testing` to one or more strategies in comma-separated value form.
+To set them you need to set Java system property `const:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING"]` to one or more strategies in comma-separated value form.
```
MatousJobanek: When I tried to generate it, then the resulting page contains `smart.testing.report.dir` in all cases when `[name="SMART_TESTING"]` is used. It seems that it takes the first variable that starts with the given `"SMART_TESTING"` string. When I moved this variable to the first place inside of the java class, then it is generated correctly.

docs/configuration.adoc
lordofthejars: Should we rename the variable names to this PR? I think it should be another one, since this was just a PR for updating docs, I think this improvement (that should be done) should be a separate issue.

docs/configuration.adoc
MatousJobanek: Here is another problem with the generation process - the first variable is property as: `scm.range.tail`, so the extension/generator matches the first variable that contains the given string `"COMMIT"` which is in this case: `PREVIOUS_COMMIT = "scm.range.tail"`

docs/configuration.adoc
MatousJobanek: it's more than that: https://github.com/arquillian/smart-testing/pull/150#discussion_r139092538

docs/configuration.adoc
lordofthejars: humm probably an error on my regexp :) Thanks for noticing it

docs/configuration.adoc
```diff
@@ -109,7 +111,7 @@ WARNING: At this moment, this strategy does not work with Java 9.
 `Failed` strategy just gets all tests that failed from previous executions and mark them as *important* tests to run first (_ordering_) or not filtered (_selecting_).
 
 This strategy uses the _JUnit_ XML https://github.com/apache/maven-surefire/blob/master/maven-surefire-plugin/src/site/resources/xsd/surefire-test-report.xsd[report] for reading past executions.
-All reports from previous local build are automatically copied by the maven extension to a temp directory `${project.directory}/.smart-testing/execution/reports` and when the build is finished the directory is removed.
+All reports from previous local build are automatically copied by the maven extension to a temp directory `${project.directory}/const:core/src/main/java/org/arquillian/smart/testing/hub/storage/local/LocalStorage.java[name="SMART_TESTING_WORKING_DIRECTORY_NAME"]/const:core/src/main/java/org/arquillian/smart/testing/hub/storage/local/LocalStorage.java[name="EXECUTION_SUBDIRECTORY"]/const:core/src/main/java/org/arquillian/smart/testing/hub/storage/local/LocalStorage.java[name="REPORTING_SUBDIRECTORY"]` and when the build is finished the directory is removed.
```
MatousJobanek: the variable `REPORTING_SUBDIRECTORY` is wrong - see my description here: https://github.com/arquillian/smart-testing/pull/145
there should be: 
`core/src/main/java/org/arquillian/smart/testing/spi/TestResult.java[name="TEMP_REPORT_DIR"]`

docs/usage.adoc
```diff
@@ -10,17 +10,17 @@ Let's see some examples:
...
+`$ mvn clean test -Dconst:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING_MODE"]=selecting -Dconst:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING"]="new, changed"`
 
 *You want to run all tests but given priority to the latest tests added or modified*
 
-`$ mvn clean test -Dsmart.testing.mode=ordering -Dscm.last.changes=1 -Dsmart.testing="new, changed"`
+`$ mvn clean test -Dconst:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING_MODE"]=ordering -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]=1 -Dconst:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING"]="new, changed"`
 
-TIP: `smart.testing.mode` is not mandatory in case of `ordering` because it is the **default** mode.
+TIP: `const:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING_MODE"]` is not mandatory because `const:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="DEFAULT_MODE"]` is the **default** mode.
```
MatousJobanek: You have changed the wrong/old line:

> smart.testing.mode is not mandatory in case of ordering because it is the default mode.

to:

> smart.testing.mode is not mandatory because selecting is the default mode.

but in the description of this command is:

> You want to run all tests but given priority to the latest tests added or modified

In other words, the user wants to run *all* tests, so the mode is mandatory and has to contain `ordering`

docs/configuration.adoc
lordofthejars: YEs fixing it right now, but the rename task I think it should be another PR

docs/usage.adoc
lordofthejars: Don't get what you mean

docs/refcard.adoc
```diff
@@ -42,35 +42,35 @@ a|`surefire`, `failsafe`
...
 |Set the number of commits from `HEAD` that you want to inspect
-a|`0`
+a|`const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="DEFAULT_LAST_COMMITS"]`
 a|`new`, `changed`, `affected`
 
-a|`smart.testing.affected.inclusions`
+a|`const:strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java[name="SMART_TESTING_AFFECTED_INCLUSIONS"]`
 |Set classes to be included for scanning
 a|
-a|`affected`
```
MatousJobanek: Here and in next three cases, you removed the information about which strategy is the variable applicable for.

docs/configuration.adoc
lordofthejars: YEs same problem as in other places, that when regexp thing

docs/refcard.adoc
lordofthejars: ouch

docs/configuration.adoc
MatousJobanek: yeah, can be :-)

docs/usage.adoc
MatousJobanek: In the description of the command is:

You want to run **all** tests....

which means that you should specify the ordering mode inside of the command. 
This part was probably written at that time when we had `ordering` mode as the default one. When we changed it to `selecting` we probably forgot to change also the documentation...

MatousJobanek: and yeah, the preceding example:

> You want to run only tests that you’ve just added or modified locally
`$ mvn clean test -Dsmart.testing.mode=selecting -Dsmart.testing.report.dir="new, changed"`

 could be changed as well. The mode is not necessary here as selecting is the default one :-)

MatousJobanek: And the next one as well: 

> You want to run only tests that validates new or modified business classes locally
`$ mvn clean test -Dsmart.testing.mode=selecting -Dsmart.testing.report.dir="affected"`

MatousJobanek:
Good job @lordofthejars

PR #152, opened 2017-09-14
Integration with maven plexus logger.

### LINKED ISSUE #133: Align ST Logger also with Maven logging format
In ST we use our own [Logger](https://github.com/arquillian/smart-testing/blob/master/core/src/main/java/org/arquillian/smart/testing/Logger.java) that reflects the way of logging used in Surefire providers. In the original surefire providers, they are using their own implementations (based on the information if the execution is using a fork or not). But in any case, they log on stdout. This is also why I have implemented it using stdout and stderr.

The current state is fine in case of logging in the phase of provider execution - before/during/after test invocation. However, there is also Maven extension that logs much earlier and should reflect the way of logging that is used by Maven. For this, Maven provides [injection](http://maven.apache.org/ref/3.1.1/maven-embedder/logging.html) of `org.codehaus.plexus.logging.Logger`. 

Our Logger should reflect the phase when it is used and should use the correct way of logging.

hemanik:

The new version makes the smart-testing logger more generic providing the option of creating/using a new custom loggers

Created two interfaces:
* `Logger` - representing a logger that will log on console
* `LoggerFactory` - takes care of creating instances of the logger class

There are (so far) three implementations of the interfaces - the Default one and one for MavenExtension thingy and one for SurefireProvider.
* `DefaultLogger` - has the same behavior as the original `Logger`
* `MavenExtensionLogger` - is just a delegator of the injected maven-plexus-logger
* `SurefireProviderLogger` - is just a delegator of the `ConsoleLogger`

then there is `Log` class that is kind of registry that stores the latest `LoggerFactory` that has been set to be used.
Calling: 
`Log.setLoggerFactory(new MavenExtensionLoggerFactory(mavenLogger, configuration));`
I'm saying that from now on, everytime when I call `Log.getLogger();` the `MavenExtensionLoggerFactory` should be used for creating new instances of the `Logger`

Similar thing I need to do in the `SurefireProvider` to set the `DefaultLoggerFactory`.

Fixes #133


mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -28,7 +29,8 @@
     hint = "smart-testing")
 class SmartTestingMavenConfigurer extends AbstractMavenLifecycleParticipant {
 
-    private static final Logger logger = Logger.getLogger();
+    @Requirement
+    public static org.codehaus.plexus.logging.Logger LOGGER;
```
hemanik: @MatousJobanek @bartoszmajsak, the injected Logger always returns a null value. I am unable to detect the issue. Is there something basic that I am missing here for the DI to work ?

bartoszmajsak: It cannot be `static` in this case. Injection works with instance fields.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -6,37 +6,38 @@
...
         this.dependencyResolver = new DependencyResolver(configuration);
+        this.logger = logger;
     }
 
     void configureTestRunner(Model model) {
         final List<Plugin> effectiveTestRunnerPluginConfigurations = getEffectivePlugins(model);
 
         if (!effectiveTestRunnerPluginConfigurations.isEmpty()) {
-            logger.debug("Enabling Smart Testing %s for plugin %s in %s module", ExtensionVersion.version().toString(),
+            logger.debug(format("Enabling Smart Testing %s for plugin %s in %s module", ExtensionVersion.version().toString(),
```
hemanik: @bartoszmajsak @MatousJobanek, what I was trying to say was that these debug logs for the plexus logger would only be enabled in the event `-X` flag is set, otherwise it won't.

IMO, this is also logging information related to smart testing extension, and should be available, when just `smart.testing.debug` property is set.

WDYT ?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/DebugModeSmartTestingFunctionalTest.java
```diff
@@ -80,7 +80,7 @@ public void should_show_debug_logs_when_maven_is_executed_in_debug_mode() throws
 
         // then
         String projectMavenLog = project.getMavenLog();
-        assertThat(projectMavenLog).contains(DEBUG_LOGS);
+        //assertThat(projectMavenLog).contains(DEBUG_LOGS);
```
bartoszmajsak: Why is it commented out?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -6,37 +6,38 @@
...
 
+import static java.lang.String.format;
 import static org.arquillian.smart.testing.mvn.ext.MavenPropertyResolver.isSkipITs;
 
 class MavenProjectConfigurator {
 
     private static final Version MINIMUM_VERSION = Version.from("2.19.1");
 
-    private static final Logger logger = Logger.getLogger();
+    private final org.codehaus.plexus.logging.Logger logger;
```
bartoszmajsak: We can import it to not have fully qualified package name scattered around.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ModifiedPomExporter.java
```diff
@@ -8,24 +8,23 @@
...
             new MavenXpp3Writer().write(pomOut, model);
-            writeModifiedPomToTarget(pomOut, model.getProjectDirectory());
+            writeModifiedPomToTarget(pomOut, model.getProjectDirectory(), logger);
         } catch (IOException e) {
             throw new RuntimeException("Failed writing updated pom file: " + model.getPomFile().getAbsolutePath(), e);
         }
     }
 
-    private static void writeModifiedPomToTarget(StringWriter modifiedPom, File project) throws IOException {
+    private static void writeModifiedPomToTarget(StringWriter modifiedPom, File project, org.codehaus.plexus.logging.Logger logger) throws IOException {
```
bartoszmajsak: If we cannot inject it, can we pass is as constructor parameter instead?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -28,7 +29,8 @@
     hint = "smart-testing")
 class SmartTestingMavenConfigurer extends AbstractMavenLifecycleParticipant {
 
-    private static final Logger logger = Logger.getLogger();
+    @Requirement
+    private org.codehaus.plexus.logging.Logger LOGGER;
```
bartoszmajsak: If it's not static then it shouldn't be uppercase. Also we can import it as usual, no need for fully qualified package name here.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/DebugModeSmartTestingFunctionalTest.java
hemanik: Because the assertion was failing, we seem to no longer get the smart-testing debug logs, most of which have been replaced by plexus logger but few other ones are also not displayed any longer, which is strange.

MatousJobanek:
Next steps that have to be done
- [ ] Create another Logger implementation for the provider side - this one will delegate `ConsoleLogger` provided by Surefire in `ProviderParameters` - just call `bootParams.getConsoleLogger()` in `SmartTestingSurefireProvider` to get the instance. 
- [ ] The `ConsoleLogger` has only one method `info`, so print everything using the info.
- [ ] If the `ConsoleLogger` does not add any prefix - so add the relevant prefixes (WARN, INFO, ERROR, DEBUG)
- [ ] Check and modify the strings based on the fact if `ConsoleLogger` adds a new line at the end of the message
- [ ] Modify the current javadoc  or add a new one
- [ ] Check if this new Logger logic works in all possible fork-configurations - see `SurefireForksConfigurationTest` 
- [ ] Add test for the `Log` loggic

core/src/main/java/org/arquillian/smart/testing/logger/Logger.java
```diff
@@ -3,8 +3,7 @@
 public interface Logger {
 
     /**
-     * Will format the given message with the given arguments and prints it on standard output with the prefix:
-     * "[INFO] [Smart Testing Extension]"
+     * Will format the given message with the given arguments and print it on standard output.
```
MatousJobanek: > ...and print it on standard output.

On the interface level, we don't know where the information will be printed. I would just say that the message will be logged at the info level. analogically for the following methods.

core/src/main/java/org/arquillian/smart/testing/logger/Logger.java
```diff
@@ -25,8 +23,7 @@
     void warn(String msg, Object... args);
 
     /**
-     * Will format the given message with the given arguments and prints it on standard output with the prefix:
-     * "[DEBUG] [Smart Testing Extension]", if debug mode is enabled.
+     * Will format the given message with the given arguments and print it on standard output, if debug mode is enabled.
      *
      * @param msg
      *     The string message (or a key in the message catalog)
```
MatousJobanek: > or a key in the message catalog

I'm not sure what is meant by that...

core/src/test/java/org/arquillian/smart/testing/logger/LogTest.java
```diff
@@ -0,0 +1,29 @@
...
+    @Test
+    public void should_return_logger_instance_of_set_logger_factory() throws Exception {
+        // given
+        Log.setLoggerFactory(new DummyLoggerFactory());
+
+        //when
+        final Logger logger = Log.getLogger();
+
+        // then
+        assertThat(logger).isInstanceOf(new DummyLoggerFactory().getLogger().getClass());
```
MatousJobanek: Try to reset it back to the `DefaultLoggerFactory` and verify also that

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/logger/SurefireProviderLoggerFactory.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public Logger getLogger() {
+        return new SurefireProviderLogger(logger);
+    }
+
+    private class SurefireProviderLogger implements Logger {
+
+        private static final String PREFIX = "%s: Smart Testing Extension - ";
```
MatousJobanek: How about logging in the provider phase with the prefix `%s: Smart Testing Provider - `? Wouldn't that make more sense to have in maven extension phase: `[Smart Testing Extension]` and in provider with the relevant string? 
Or does it look just weird and confusing? WDYT @lordofthejars @bartoszmajsak 
And yeah, the default logger would contain only  `%s: Smart Testing - ` prefix. I'm also thinking in other contexts - Che or other IDEs

bartoszmajsak: Didn't we set it to `[]` thingy before for everything? Why is it now changed?

bartoszmajsak: We should have it unified for all so it's easy to grep etc.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/logger/SurefireProviderLoggerFactory.java
```diff
@@ -0,0 +1,63 @@
...
+        @Override
+        public void error(String msg, Object... args) {
+            consoleLogger.info(getFormattedMsg("ERROR", msg, args));
+        }
+
+        private String getFormattedMsg(String level, String msg, Object... args) {
+            if (args != null && args.length > 0) {
+                msg = format(msg, args);
+            }
+            return format(PREFIX, level) + msg + System.lineSeparator();
```
MatousJobanek: I executed the `SurefireForksConfigurationTest` and noticed that in some cases (eg. `test_with_multiple_forks_not_reusing_forks`) there are empty lines - there are two line separators.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/logger/SurefireProviderLoggerFactory.java
MatousJobanek: Yeah, you set it to `[ ]` because we wanted to make it unified with maven extension/plugin logging. Now, we have the possibility to have different prefixes in different stages of the maven build. 
When you take a look at the logging produced by surefire logic/providers, there is no `[ ]` present in the log. 
We should keep it aligned also with the logging of the particular stage.

bartoszmajsak: I am not really sure if that is so important to have it aligned with how surefire logs stuff, but maybe it makes sense from the point of view of seeing which side of the build it is - the build engine / extension vs plugin. If so, then let's make it this way.

MatousJobanek: If you want to keep it as unified as possible, then I would remove brackets from the prefix at all. I mean from maven extension part `[Smart Testing Extension]`. When you check the maven log, there is not such a line containing multiple strings with brackets in sequence.
so in the maven extension, the log would start with:
`[INFO] Smart Testing Extension - `
and for provider side:
`INFO: Smart Testing Extension - `

MatousJobanek: > I am not really sure if that is so important to have it aligned with how surefire logs stuff

because you don't log from maven extension but from the surefire side. For my eyes, it is better having one way of logging in particular phases - maven extension(or plugin) logs vs surefire logs...

MatousJobanek: as for the content of the prefix, yeah I agree that it would be good having it unified - to grep it easily.

bartoszmajsak: I think we are on the same page here, no need to quote the first part of my previous comment, where I agree with unification based on the follow-up reasoning which you just rephrased.

MatousJobanek: So, do we agree on: 
* `[INFO] Smart Testing Extension -` for maven extension logs
* `INFO: Smart Testing Extension -` for both the default logger and surefire provider logger

bartoszmajsak: Took us few extra comments, but YES we do :D

agree

MatousJobanek: I love your gifs :-D

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/logger/SurefireProviderLoggerFactory.java
```diff
@@ -23,7 +23,7 @@ public Logger getLogger() {
 
     private class SurefireProviderLogger implements Logger {
 
-        private static final String PREFIX = "%s: Smart Testing Extension - ";
+        private static final String PREFIX = "%s: Smart Testing Provider - ";
```
MatousJobanek: Please see this comment: https://github.com/arquillian/smart-testing/pull/152#discussion_r140217015

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/logger/SurefireProviderLoggerFactory.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        private String getFormattedMsg(String level, String msg, Object... args) {
+            if (args != null && args.length > 0) {
+                msg = format(msg, args);
+            }
+            if (consoleLogger.getClass().getName().equals("org.apache.maven.surefire.booter.ForkingRunListener")) {
+                return format(PREFIX, level) + msg + System.lineSeparator();
+            }
+            return format(PREFIX, level) + msg;
```
MatousJobanek: I would use StringBuffer for it

core/src/main/java/org/arquillian/smart/testing/logger/Logger.java
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @param msg
+     *     Message to print
+     * @param args
+     *     arguments to use for formatting the given message
+     */
+    void info(String msg, Object... args);
+
+    /**
+     * Will format the given message with the given arguments and log it warn level.
```
MatousJobanek: missing "at"

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SurefireForksConfigurationTest.java
```diff
@@ -24,7 +24,7 @@
 public class SurefireForksConfigurationTest {
 
     private static final String MAVEN_LOGS = "[Smart Testing Extension]";
-    private static final String SUREFIRE_PROVIDER_LOGS = "Smart Testing Extension -";
+    private static final String SUREFIRE_PROVIDER_LOGS = "Smart Testing Provider -";
```
MatousJobanek: Check also the prefix - `INFO` or `[INFO]`

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
```diff
@@ -15,8 +15,8 @@
 
 public class SkipTestExecutionFunctionalTest {
 
-    private static final String SMART_TESTING_EXTENSION_DISABLED = "[Smart Testing Extension] Smart Testing is disabled.";
-    private static final String SMART_TESTING_EXTENSION_ENABLED = "[Smart Testing Extension] Enabling extension.";
+    private static final String SMART_TESTING_EXTENSION_DISABLED = "Smart Testing Extension - Smart Testing is disabled.";
+    private static final String SMART_TESTING_EXTENSION_ENABLED = "Smart Testing Extension - Enabling extension.";
```
MatousJobanek: check also the prefix `[INFO]`

MatousJobanek:
In the functional tests, you check the presence of the prefix `Smart Testing Extension -` in the Maven log. It would be great to check also that the correct prefix - with and without `[]` -  is used in the correct phase. 
It is mainly important in the `SurefireForksConfigurationTest.java` class as there are used forks.

MatousJobanek:
Please, also check if it is necessary to reset `MavenExtensionLogger` in the [SmartTestingMavenConfigurer#afterSessionEnd](https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java#L79)

hemanik:
> Please, also check whether it is necessary to reset MavenExtensionLogger in the SmartTestingMavenConfigurer#afterSessionEnd or not

@MatousJobanek, `afterSessionEnd` is the last part to get executed and after checking it I don't feel the necessity to reset MavenExtensionLogger since there is no logging done after this point.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/DebugModeSmartTestingFunctionalTest.java
```diff
@@ -46,12 +48,14 @@ public void should_show_debug_logs_when_smart_testing_is_executed_in_debug_mode(
         final TestResults actualTestResults = projectBuilder
                 .options()
                     .withSystemProperties(SCM_RANGE_HEAD, "HEAD", SCM_RANGE_TAIL, "HEAD~", SMART_TESTING_DEBUG, "true")
+                    .logBuildOutput()
```
MatousJobanek: this is not necessary

core/src/main/java/org/arquillian/smart/testing/impl/TestStrategyApplierImpl.java
```diff
@@ -12,16 +12,16 @@
...
 import org.arquillian.smart.testing.api.TestStrategyApplier;
 import org.arquillian.smart.testing.configuration.Configuration;
+import org.arquillian.smart.testing.logger.Log;
 import org.arquillian.smart.testing.report.SmartTestingReportGenerator;
 import org.arquillian.smart.testing.spi.TestExecutionPlanner;
 
 class TestStrategyApplierImpl implements TestStrategyApplier {
 
-    private static final Logger logger = Logger.getLogger();
+    private static final org.arquillian.smart.testing.logger.Logger logger = Log.getLogger();
```
MatousJobanek: I guess that this doesn't have to be fully qualified

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/DebugModeSmartTestingFunctionalTest.java
```diff
@@ -46,12 +48,14 @@ public void should_show_debug_logs_when_smart_testing_is_executed_in_debug_mode(
...
                 .configure()
             .run();
 
         // then
         String projectMavenLog = project.getMavenLog();
-        assertThat(projectMavenLog).contains(DEBUG_LOGS);
+        assertThat(projectMavenLog).contains(PROVIDER_DEBUG_LOGS);
```
MatousJobanek: both provider and maven extension logs should be present

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/DebugModeSmartTestingFunctionalTest.java
```diff
@@ -46,12 +48,14 @@ public void should_show_debug_logs_when_smart_testing_is_executed_in_debug_mode(
...
+        assertThat(projectMavenLog).contains("Applied user properties");
```
MatousJobanek: if you want to check a concrete string, you should also use the related prefix

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SurefireForksConfigurationTest.java
```diff
@@ -81,7 +84,10 @@ private void verifyTestSuiteExecution(String... systemPropertiesPairs){
                 .withSystemProperties(SMART_TESTING_REPORT_ENABLE, "true")
             .configure()
             .run();
+
         // then
+        String projectMavenLog = project.getMavenLog();
+        assertThat(projectMavenLog).contains(SUREFIRE_PROVIDER_LOGS);
```
MatousJobanek: check also maven extension logs. and in this particular case use the smart testing debug mode and check specific strings when the extension is loaded, during the surefire provider phase and again in maven extension when the session is ended

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -14,15 +14,16 @@
...
+import org.arquillian.smart.testing.logger.Logger;
 import org.arquillian.smart.testing.RunMode;
 import org.arquillian.smart.testing.hub.storage.local.LocalStorage;
 import org.arquillian.smart.testing.hub.storage.local.LocalStorageFileAction;
 import org.yaml.snakeyaml.Yaml;
 
 public class Configuration implements ConfigurationSection {
 
-    private static final Logger logger = Logger.getLogger();
+    private static final Logger logger = new DefaultLoggerFactory(false).getLogger();
```
MatousJobanek: why the default logger?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/logger/MavenExtensionLoggerFactory.java
```diff
@@ -0,0 +1,71 @@
...
+
+        @Override
+        public void error(String msg, Object... args) {
+            logger.error(getFormattedMsg(msg, args));
+        }
+
+        @Override
+        public boolean isDebug() {
+            return logger.isDebugEnabled() || (configuration != null && configuration.isDebug())
+                   || Boolean.valueOf(System.getProperty(Configuration.SMART_TESTING_DEBUG));
```
dipak-pawar: We don't need to check this here as configuration is already overwriting it if `smart.testing.debug` is set using system property.

core/src/main/java/org/arquillian/smart/testing/logger/DefaultLoggerFactory.java
```diff
@@ -0,0 +1,104 @@
+package org.arquillian.smart.testing.logger;
+
+import org.arquillian.smart.testing.configuration.Configuration;
+
+import static java.lang.String.format;
+
+public class DefaultLoggerFactory implements LoggerFactory {
+
+    private boolean isDebugLogLevel = false;
```
dipak-pawar: Do we really need to define default value? If you haven't define it will take `false` as default

dipak-pawar:
@hemanik Can you check my comments? 

Also I have a feeling we should update document about logging format. What is `[INFO] Smart Testing Extension - ` & `INFO: Smart Testing Extension -` So user can easily understand why ST is printing debug logs in two formats. WDYT @MatousJobanek ?

core/src/main/java/org/arquillian/smart/testing/logger/DefaultLoggerFactory.java
MatousJobanek: From my point of view, it's kind of transparency. But you are right - can be removed.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/logger/MavenExtensionLoggerFactory.java
MatousJobanek: It is necessary to check it here. We set the `MavenExtensionLoggerFactory` before we create the configuration object: https://github.com/hemanik/smart-testing/blob/be0c6c9deba19d50f075fdc0f40a658d72c63372/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java#L52 and we don't pass the configuration as a parameter, which means that it is null. So we can only check the system property.
Then, when the configuration instance is retrieved, we can reset the factory with a new instance that contains also the `Configuration`: https://github.com/hemanik/smart-testing/blob/be0c6c9deba19d50f075fdc0f40a658d72c63372/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java#L55

MatousJobanek:
As for the documentation, I don't see it as essential, but a short description can be provided... 
I wouldn't explain the two formats because we just want to align the logging format with the correct phase (Maven/Surefire). The only thing I would put there is a note that all ST related logs can be found with the prefix `Smart Testing Extension -`.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/logger/MavenExtensionLoggerFactory.java
dipak-pawar: But in this case if user is not setting system property & he is defining `debug: true` in config file. then in this case we should enable debug options, but we are not enabling it because configuration object is not created before setting `MavenExtensionLoggerFactory`

dipak-pawar:
@MatousJobanek +1 for short description

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/logger/MavenExtensionLoggerFactory.java
MatousJobanek: Yeah, exactly. This is a chicken-egg problem and there is no better solution for it.
BTW, currently, there is no debug log in the `Configuration` class - only info & warn, so you don't have to care about it - we don't miss any information ;-)

dipak-pawar: This is fine for now. But we should add comment about this why we did this way, so any new developer can understand it easily

MatousJobanek:
Great job @hemanik 👍 
Thank you 🥇

Issue #151, opened 2017-09-14
Create an Asciidoctor extension that adds a copy to clipboard button


lordofthejars:
Create an Asciidoctor extension that adds a copy to clipboard button

PR #153, opened 2017-09-15
Implements copy to clipboard Asciidoctor extension

lordofthejars:
Implements copy to clipboard Asciidoctor extension.

Fixes: #151


.asciidoctor/lib/copy-to-clipboard-inline-macro/extension.rb
```diff
@@ -0,0 +1,73 @@
...
+class ClipboardAssetsDocinfoProcessor < Extensions::DocinfoProcessor
+  use_dsl
+  at_location :footer
+
+  def process doc
+
+    clipboard = %(
+      var clipboard = new Clipboard('.cpybtn');
+      clipboard.on('success', function(e) {
+        console.log(e);
```
bartoszmajsak: Do we really have to log to console? Just wondering what's the usefulness of it

lordofthejars: I read in clipboard website. What I suspect is that it is important also to notice if something has been copied as well, so anyone can validate if the thing you are copying is the one it is on web page. For this reason I decided to left.

bartoszmajsak: You only see it when you have web tools opened in the browser, so normally you won't see it anyway.

lordofthejars: yes, but for example I see that might help in case any user opens an issue about copy to clipboard feature. We can always say, check console and tell me what's there or something like this. I see some small benefits on having it there.

MatousJobanek:
If I understand it correctly - this is not generated anywhere so far, right? This PR only brings the extension to be used in the future.
I just wanted to see what it looks like...

.asciidoctor/lib/copy-to-clipboard-inline-macro/extension.rb
bartoszmajsak: Ok, then maybe instead of having the same logging content let's differentiate if it was a `success` or a `failure`? e.g. use some prefixes so we know from where this log is coming, as well as `console.error` for the faulty path.

On top of that not all the versions of browsers are supporting logging - see http://caniuse.com/#feat=console-basic, but I'm not sure if we should care at this stage.

lordofthejars:
@MatousJobanek You are right. I'll send another PR with the updates

Issue #158, opened 2017-09-18
There is not need to have report name/directory customizable


MatousJobanek:
Following properties are not really needed:
  * `smart.testing.report.dir`
  * `smart.testing.report.name`

In addition, the property customizing a report directory can be very confusing.
Both of them should be removed and used the default names. 

If there is a use-case/request where the customization functionality is needed, we can bring it back.

PR #161, opened 2017-09-19
Moved report to target/smart-testing/reporting/report.xml

MatousJobanek:
Changes:
* changed storage directory to: `target/smart-testing/reporting/`
* changed report name to: `report.xml` - IMHO the original `smart-testing-report.xml` had duplicated information that is contained in the directory where the report is stored.
* for storing the report file, I'm using `LocalStorage` logic proposed by: https://github.com/arquillian/smart-testing/pull/159
* removed properties: `smart.testing.report.dir` & `smart.testing.report.name`

Changes proposed by this PR is contained in the commit: https://github.com/arquillian/smart-testing/pull/161/commits/253fb4cd3fce1ac987911b35f8c8076124345cb1

fixes: https://github.com/arquillian/smart-testing/issues/158

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/Configuration.java
+++ b/core/src/main/java/org/arquillian/smart/testing/Configuration.java
@@ -5,9 +5,6 @@ public class Configuration {
     public static final String DEFAULT_MODE = "selecting";
     public static final String DEFAULT_STRATEGIES = "";
     public static final String ENABLE_REPORT_PROPERTY = "smart.testing.report.enable";
-    public static final String DEFAULT_REPORT_FILE_NAME = "smart-testing-report.xml";
-    public static final String SMART_TESTING_REPORT_DIR = "smart.testing.report.dir";
-    public static final String SMART_TESTING_REPORT_NAME = "smart.testing.report.name";
 
     public static final String SMART_TESTING = "smart.testing";
     public static final String SMART_TESTING_MODE = "smart.testing.mode";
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/arquillian/smart/testing/hub/storage/local/DuringExecutionLocalStorage.java
```diff
@@ -0,0 +1,44 @@
+package org.arquillian.smart.testing.hub.storage.local;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+
+public class DuringExecutionLocalStorage extends AfterExecutionLocalStorage {
+
+    public static final String SMART_TESTING_WORKING_DIRECTORY_NAME = ".smart-testing";
+    public static final String TEMPORARY_SUBDIRECTORY = "temporary";
```
bartoszmajsak: Let's move it to `Configuration` class as this will be anyway reworked to file-based solution. Having it centralized it will let us move to the file-based approach easier and not forget important bits.

MatousJobanek: Personally, into the `Configuration` class I would put those constants that are used there (or parameters that can be configured or constants containing defaults values and similar). 
If you want to make the constants centralized, then let's create `Constants` class that will contain everything.
These directory names are static values and cannot be configured anyhow.

bartoszmajsak: Right, this is something we don't need to expose to end-users. Fine with me.

bartoszmajsak:
Can you merge with `master` so that the doc issue is resolved?

alien-ike:
just testing

alien-ike:
just testing

PR #162, opened 2017-09-19
fix(#160): handles RENAME for changed strategy

### LINKED ISSUE #160: Should not fail If committed changes has deleted file
##### Issue Overview
Currently we are only looking for `ADD` & `MODIFY` [ChangeType](https://github.com/arquillian/smart-testing/blob/master/core/src/main/java/org/arquillian/smart/testing/scm/ChangeType.java#L5) 

While extracting committed changes [here](https://github.com/arquillian/smart-testing/blob/master/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java#L145) we should look for `DELETE` OR `RANAMED` ChangeType & filter it later if not interested.  

##### Expected Behaviour
ST should work fine with deleted test file.

##### Current Behaviour
ST is failing with exception 
```java
org.apache.maven.InternalErrorException: Internal error: java.lang.IllegalArgumentException: No enum constant org.arquillian.smart.testing.scm.ChangeType.DELETE
```

##### Steps To Reproduce
1. Clone this reproducible [project](https://github.com/dipak-pawar/ST_reproducer)
2. Run tests using cmd
``` bash
mvn clean test -Dsmart.testing=new,changed,affected -Dscm.range.head=HEAD -Dscm.range.tail=HEAD~3
```
Note: I haven't tested this for `RENAMED` ChangeType. Please test this as well. I believe this will also fail.

bartoszmajsak:

We missed mapping all change types from diffs to our representation. 

On top of that we don't consider RENAMEs currently (we treat them as DELETE and ADD) - so this will fail because we don't map `DELETE` change type. If we would have this mapping it will fall into "new" strategy though.

- covers mapping between jgit change types and ours
- consistency unit test ensuring we will cover all when upgrading to newer versions (unlikely to change though)
- recognizes `RENAME`s properly

Fixes #160


core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
```diff
@@ -88,6 +89,26 @@ public boolean isApplicable() {
...
+            final CanonicalTreeParser oldTree = new CanonicalTreeParser();
+            oldTree.reset(reader, oldHead);
+            final CanonicalTreeParser newTree = new CanonicalTreeParser();
+            newTree.reset(reader, newHead);
+
+            final List<DiffEntry> commitDiffs = git.diff()
+                    .setNewTree(newTree)
+                    .setOldTree(oldTree)
+                .call();
+            return transformToChangeSet(findRenames(commitDiffs), repoRoot);
```
lordofthejars: Aren't we just transforming the renames and not the other changes like add or modify?

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
```diff
@@ -121,24 +142,13 @@ public boolean isApplicable() {
...
-            oldTree.reset(reader, oldHead);
-            final CanonicalTreeParser newTree = new CanonicalTreeParser();
-            newTree.reset(reader, newHead);
-
-            final List<DiffEntry> commitDiffs = git.diff().setNewTree(newTree).setOldTree(oldTree).call();
-            return extract(commitDiffs, repoRoot);
-        } catch (IOException | GitAPIException e) {
-            throw new IllegalStateException(e);
-        }
+    private List<DiffEntry> findRenames(List<DiffEntry> commitDiffs) throws IOException {
```
dipak-pawar: Is it only finding Renamed DiffEntry or it's finding renamed and others(Added, Modified) as well?

bartoszmajsak: As I commented on #160 - by default jgit sees it as `DELETE` and `ADD` - this checks if they are actually renames.

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
bartoszmajsak: No, we transform all. `findRenames` will merge `ADD` and `DELETE` entries which are in fact `RENAME`s.

bartoszmajsak: Would be more understandable with different name? What would you suggest?

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
dipak-pawar: ok But what about other entries like `Modify` we should consider that as well right?

core/src/main/java/org/arquillian/smart/testing/scm/ChangeType.java
```diff
@@ -2,6 +2,19 @@
...
+    MODIFY,
+
+    /** Delete an existing file from the project */
+    DELETE,
+
+    /** Rename an existing file to a new location */
+    RENAME,
+
+    /** Copy an existing file to a new location, keeping the original */
+    COPY;
```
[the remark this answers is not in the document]
bartoszmajsak: I'm not sure really if we need it at this point. You can find 1-1 copies, but do we really care now of finding duplicates? I'm not sure if that should be our responsibility to find duplicated tests.

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
lordofthejars: Yes I think @dipak-pawar was confused as well. But I am not sure about another name, maybe `mergeRenames`?

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
bartoszmajsak: See https://github.com/arquillian/smart-testing/pull/162#discussion_r139666555

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
bartoszmajsak: I was thinking the same, but it actually merges adds and deletes to renames based on similarity score.

lordofthejars: Then maybe add a comment saying this?

bartoszmajsak: That's exactly what I'm doing right now

core/src/main/java/org/arquillian/smart/testing/scm/ChangeType.java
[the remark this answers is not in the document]
dipak-pawar: I believe you won't get any duplicate test file with same path. If it is on different path then it will be treated as `New`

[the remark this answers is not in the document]
bartoszmajsak: The same path? Not sure what you mean by that.

Have a look here https://git-scm.com/docs/git-status#_changed_tracked_entries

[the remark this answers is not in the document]
bartoszmajsak: Maybe something to investigate further, but not as a part of this PR.

PR #164, opened 2017-09-19
docs: introduces script for generating docs

### LINKED ISSUE #163: Simplify documentation generation process
As the `asciidoc` command for generating docs is getting lengthier and lengthier let's wrap it in the script which can be used by both us (and contributors) as well as on CI server.

bartoszmajsak:
Provides a convenient script for generating docs so it can be used by contributors and on CI server.

Fixes #163

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -25,7 +25,7 @@ ifndef::generated-doc[]
 To read complete documentation head over to http://arquillian.org/smart-testing/
 
 You can also generate it locally. All you need to have in place is [docker](https://docs.docker.com/engine/installation/).
-Simply execute `$ ./.asciidoctor/generate.sh` and you are set.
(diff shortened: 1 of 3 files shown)


README.adoc
```diff
@@ -23,6 +23,10 @@ We created *Smart Testing* to give you the fastest possible feedback loop when i
 
 ifndef::generated-doc[]
 To read complete documentation head over to http://arquillian.org/smart-testing/
+
+You can also generate it locally. All you need to have in place is [docker](https://docs.docker.com/engine/installation/).
+Simply execute `$ ./.asciidoctor/generate.sh` and you are set.
```
lordofthejars: Add copy to clipboard extension?

bartoszmajsak: It will not work, as this part is only rendered on GH.

lordofthejars: ah true :)

Issue #175, opened 2017-09-20
Shaded extension doesn't work with older versions of Maven


bartoszmajsak:
##### Issue Overview

Following the guide, I copied shaded extension (`maven-lifecycle-extension-0.0.1-shaded.jar`) into `M2_HOME/lib/ext`, but when running the build nothing happens (no sign of extension in the logs).

##### Expected Behaviour

The extension is working after installing it by [following the guide](http://arquillian.org/smart-testing/#_maven_3_1_x).

##### Current Behaviour

No signs of extension kicking in (nothing in the logs).

##### Steps To Reproduce

1. Change Maven to version pre `3.3.x`
2. Download extension from [Maven Central](http://central.maven.org/maven2/org/arquillian/smart/testing/maven-lifecycle-extension/0.0.1/maven-lifecycle-extension-0.0.1-shaded.jar)
3. Copy to `M2_HOME/lib/ext`
4. Run the build using e.g. `mvn clean package -Dsmart.testing=new,changed -Dsmart.testing.mode=selecting`

##### Additional Information

```
Apache Maven 3.2.5 (12a6b3acb947671f09b81f49094c53f426d8cea1; 2014-12-14T18:29:23+01:00)
Maven home: /usr/share/maven/apache-maven-3.2.5
Java version: 1.8.0_144, vendor: Oracle Corporation
Java home: /usr/java/jdk1.8.0_144/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.12.9-300.fc26.x86_64", arch: "amd64", family: "unix"
arquillian-core [mvn --version]
```

bartoszmajsak:
So we have to decide, either we treat it as high prio and ship fix soon enough, or we revise the docs.

lordofthejars:
I think that the easiest and most used way by users would be to just drop support for versions prior to 3.3.X for now.

MatousJobanek:
Do you have any idea what could be the problem? It can be just a simple bug, but can be also a little bit more complex. How about time-box it for several hours and if it seems to be more complex, then I would drop support for the older version...

bartoszmajsak:
My idea for dropping is temporary - we just release `0.0.2` with all the fixes and say that older versions are not supported yet. And then we can timebox to analyze. WDYT?

lordofthejars:
:+1:

bartoszmajsak:
https://github.com/jcgay/maven-profiler

PR #176, opened 2017-09-21
feat(#146): provides installation script

### LINKED ISSUE #146: Improve quick start experience (installation script)
We should provide one-liner installation which will setup Maven extension in the given project, something like:

```shell
curl -sS http://arquillian.org/install-smart-testing.sh | bash
```

which will automate the whole process which we now describe in the [docs](http://arquillian.org/smart-testing/#_maven_extension). 

Following cases to handle:

1. `.mvn/extension.xml` does not exist (or the folder itself)
2. `.mvn/extension.xml` has other extensions defined (we have to append to XML properly)
3. Detect if we can use extension mechanism, otherwise rollback to [pre Maven `3.3.x` version](http://arquillian.org/smart-testing/#_maven_3_1_x)

Additionally, we can specify a concrete version or a flag (release,snapshot) and script will take of defining a proper version of the extension.

```shell
curl -sS http://arquillian.org/install-smart-testing.sh | bash -s -- --release
```

```shell
curl -sS http://arquillian.org/install-smart-testing.sh | bash -s -- --snapshot
```

(assuming that we will publish snapshots on Bintray for example)

```shell
curl -sS http://arquillian.org/install-smart-testing.sh | bash -s -- --version=0.0.2
```

bartoszmajsak:

Installation script which simplifies getting started experience with our tool. 

`curl -sSL https://git.io/v5jD6 | bash` 

notice that the `git.io` URL in the docs is different - it points to `master` which will be working only after we merge this PR. See the docs for more options or 

1. Creates  `.mvn/extension.xml` with ST configured to latest (using Maven Central metadata) if the extension file does not exist
2. In case `.mvn/extension.xml` has other extensions defined ours is added to it 
3. If our extension is already defined and has older version we ask to overwrite and do it automatically
4. Detects if we can use extension mechanism, otherwise rollback to [pre Maven `3.3.x` version](http://arquillian.org/smart-testing/#_maven_3_1_x) - before merging we have to decide on #175  (for old way of installation we don't do any of the overwrites/updates as in points above - it's just a simple copy)

Tested only on Fedora, would be good to try on OSX before merging.

- bash installation script
- updated docs

Fixes #146

### DIFF
--- a/docs/installation.adoc
+++ b/docs/installation.adoc
@@ -5,7 +5,7 @@
 Ok, just execute following snippet and you are all set:
 
 [[installing]]
-`curl -sS https://git.io/v5jMS | bash` copyToClipboard:installing[]
+`curl -sSL https://git.io/v5jy6 | bash` copyToClipboard:installing[]
 
 You can also add these flags:
 
@@ -15,7 +15,7 @@ You can also add these flags:
 
 For example:
 
-`curl -sS https://git.io/v5jMS | bash -- --latest`
+`curl -sSL https://git.io/v5jy6 | bash -s -- --latest`
 
 This script will automatically do what we described below. If you are curious keep on reading, but you may skip it too.
 
--- a/docs/installation.adoc
+++ b/docs/installation.adoc
@@ -4,8 +4,8 @@
 
 Ok, just execute following snippet and you are all set:
 
-[[installing]]
-`curl -sSL https://git.io/v5jy6 | bash` copyToClipboard:installing[]
+[[installing-release]]
+`curl -sSL https://git.io/v5jy6 | bash` copyToClipboard:installing-release[]
 
 You can also add these flags:
 
@@ -15,7 +15,8 @@ You can also add these flags:
 
 For example:
 
-`curl -sSL https://git.io/v5jy6 | bash -s -- --latest`
+[[installing-latest]]
+`curl -sSL https://git.io/v5jy6 | bash -s +++--+++ --latest` copyToClipboard:installing-latest[]
 
 This script will automatically do what we described below. If you are curious keep on reading, but you may skip it too.
 
--- a/docs/installation.adoc
+++ b/docs/installation.adoc
@@ -9,8 +9,7 @@ Ok, just execute following snippet and you are all set:
 
 You can also add these flags:
 
-  * `-l|--latest` latest published artifact in Maven Central
-  * `-r|--release` latest release (that's the default)
+  * `-l|--latest` latest published artifact in Maven Central (that's the default)
   * `-v|--version=` - specific version to install
 
 For example:
--- a/docs/installation.adoc
+++ b/docs/installation.adoc
@@ -1,14 +1,33 @@
 == Installation
 
+=== Don't make me think
+
+Ok, just execute following snippet and you are all set:
+
+[[installing]]
+`curl -sS https://git.io/v5jMS | bash` copyToClipboard:installing[]
+
+You can also add these flags:
+
+  * `-l|--latest` latest published artifact in Maven Central
+  * `-r|--release` latest release (that's the default)
+  * `-v|--version=` - specific version to install
+
+For example:
+
+`curl -sS https://git.io/v5jMS | bash -- --latest`
+
+This script will automatically do what we described below. If you are curious keep on reading, but you may skip it too.
+
 === Maven Extension
 
-*Smart Testing* is a Maven extension, so needs to be installed as is.
+*Smart Testing* is a Maven extension, so depending on the version of Maven you have to follow slightly different approach of installing it.
 
-==== Maven >= 3.1.X
+==== Maven above 3.1.X
 
-Get Smart Testing _shaded_ copy from https://bintray.com/arquillian/Arquillian/smart-testing[Bintray] and copy it to `M2_HOME/lib/ext`
+Get Smart Testing Extension _shaded jar_ from http://central.maven.org/maven2/org/arquillian/smart/testing/maven-lifecycle-extension/[Maven Central] and copy it to `M2_HOME/lib/ext`.
 
-==== Maven >= 3.3.X
+==== Maven above 3.3.X
 
 You can still use the process described at <<Maven >= 3.1.X>> or use the new _core extension configuration mechanism_ by
 creating folder called `.mvn` in the root of your project and create inside it an `extensions.xml` file
--- a/docs/installation.adoc
+++ b/docs/installation.adoc
@@ -20,6 +20,8 @@ For example:
 
 This script will automatically do what we described below. If you are curious keep on reading, but you may skip it too.
 
+IMPORTANT: This script requires `xmllint`, so make sure you have it installed.
+
 === Maven Extension
 
 *Smart Testing* is a Maven extension, so depending on the version of Maven you have to follow slightly different approach of installing it.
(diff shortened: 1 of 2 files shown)


bartoszmajsak:
I will record `asciinema` once this is merged so we use proper `git.io` link in the recording.

install.sh
```diff
@@ -0,0 +1,142 @@
...
+      $0 ~ "<groupId>" groupId "</groupId>" {
+        sub("<version>.*</version>", "<version>" version "</version>")
+      }
+      {printf "%s", $0 RS}' | sed '$ d' > .mvn/extensions-new.xml
+    mv .mvn/extensions-new.xml .mvn/extensions.xml
+}
+
+## MAIN LOGIC
+
+MVN_VERSION=$(mvn --version | head -n1 | cut -d' ' -f3)
```
lordofthejars: If `mvn` is not in `PATH` then I am not sure what kind of error is going to be printed. You could get the result int code and check if succeeded or not.

install.sh
```diff
@@ -0,0 +1,142 @@
...
+                        ;;
+                esac
+            else
+                echo " which is the latest stable version."
+            fi
+        fi
+    fi
+
+    mv .mvn/extensions.xml .mvn/extensions-unformatted.xml
+    xmllint --format .mvn/extensions-unformatted.xml > .mvn/extensions.xml
```
lordofthejars: Seems `xmllint` is installed in MacOs and Linux but is this true in all cases?

MatousJobanek:
- [ ] add `.smart-testing` directory into the .gitignore file
I know that the directory is purged when the build is ended, but I can imagine some cases when the build process stops and the purge is not performed.

lordofthejars:
@MatousJobanek Currently when JVM stops (crash or whatever) the file is removed. So it is not bad at all adding `.gitignore` entry too but I think it is something that rarely might occur.

install.sh
bartoszmajsak: Right, I forgot to clarify it in the docs.

- [ ] Mention `xmllint` as a prerequisite in the docs

install.sh
bartoszmajsak: Right, this could be checked too

- [ ] check if `mvn` and `xmllint` is installed and report error

bartoszmajsak:
@lordofthejars I would be careful with the "whatever" part of JVM stop statement. Shutdown hooks are not always executed, from the JavaDoc.

> In rare circumstances the virtual machine may abort, that is, stop running without shutting down cleanly. This occurs when the virtual machine is terminated externally, for example with the SIGKILL signal on Unix or the TerminateProcess call on Microsoft Windows. The virtual machine may also abort if a native method goes awry by, for example, corrupting internal data structures or attempting to access nonexistent memory. If the virtual machine aborts then no guarantee can be made about whether or not any shutdown hooks will be run.

lordofthejars:
@bartoszmajsak yes I know, but doing this .gitignore thing means checking if file exists or not, then add this entry, just wondering if it has sense this effort or not :)

bartoszmajsak:
> @bartoszmajsak yes I know, but doing this .gitignore thing means checking if file exists or not, then add this entry, just wondering if it has sense this effort or not :)

Done

docs/installation.adoc
```diff
@@ -1,14 +1,36 @@
...
+
+Ok, just execute following snippet and you are all set:
+
+[[installing-release]]
+`curl -sSL https://git.io/v5jy6 | bash` copyToClipboard:installing-release[]
+
+You can also add these flags:
+
+  * `-l|--latest` latest published artifact in Maven Central
+  * `-r|--release` latest release (that's the default)
```
MatousJobanek: what is the difference between: ` latest published artifact in Maven Central` and  `latest release`?
For me, they seem to be same

bartoszmajsak: you could have betas etc

bartoszmajsak: See here http://central.maven.org/maven2/org/arquillian/smart/testing/smart-testing-parent/maven-metadata.xml

bartoszmajsak: But maybe this is not really needed - but I was thinking we can at some point have betas or even make this link configurable so it can point to local repo or on-premise nexus - just a flexibility.

bartoszmajsak: But yeah, looking at other projects in practice there is no distinction http://central.maven.org/maven2/io/netty/netty-parent/maven-metadata.xml.

bartoszmajsak: Shall we drop one? `release` or `latest`?

MatousJobanek: Nice monologue :-D
betas and alphas are releases in maven central as well, so that wouldn't be clear at all.
It would make sense in case you would like to take the latest commit (state) from the GH project, but this is not necessary I think.
So, I would drop the `release` and use just `latest`

bartoszmajsak: it seems to only be working from plugins

install.sh
```diff
@@ -0,0 +1,157 @@
...
+        case "$response" in
+            [nN][oO]|[nN])
+                ;;
+            *)
+                if [ ! -f .gitignore ]; then
+                    touch ./.gitignore
+                fi
+                cat .gitignore | grep -q '.smart-testing' && EXISTS=1 || EXISTS=0
+                if [ ${EXISTS} == 0 ]; then
+                    echo "# Smart Testing Exclusions\n.smart-testing" >> ./.gitignore
```
MatousJobanek: the resulting string in .gitignore is: `# Smart Testing Exclusions\n.smart-testing`
the command should be: `echo -e "# Smart Testing Exclusions\n.smart-testing" >> ./.gitignore`

install.sh
```diff
@@ -0,0 +1,157 @@
...
+            ;;
+        *)
+            ;;
+    esac
+}
+
+function install_extension() {
+
+    # Needs to be without new lines, otherwise my sed skills below will fail badly :\
+    EXTENSION="<extension><groupId>org.arquillian.smart.testing</groupId><artifactId>maven-lifecycle-extension</artifactId><version>${VERSION}</version></extension>"
```
MatousJobanek: simple:
`EXTENSION="<extension>\\n<groupId>org.arquillian.smart.testing</groupId>\\n<artifactId>maven-lifecycle-extension</artifactId>\\n<version>${VERSION}</version>\\n</extension>"` should work I think.

then, the xmllint could be used for the final formatting only if it is available...

install.sh
```diff
@@ -0,0 +1,157 @@
...
+
+    if [ ! -f .mvn/extensions.xml ]; then
+        echo "<?xml version=\"1.0\" encoding=\"UTF-8\"?>
+            <extensions>
+              ${EXTENSION}
+            </extensions>" > .mvn/extensions.xml
```
MatousJobanek: if the `.mvn` directory does not exist, then this line will fail. You need to check the existence of the directory as well

install.sh
bartoszmajsak: Strange. I will check again.

install.sh
bartoszmajsak: I would like to leave it like it is with the lint

install.sh
bartoszmajsak: Right, I ran it only against the repo with existing `.mvn` directory

install.sh
bartoszmajsak: I meant visual new lines in the editor not `\n` BTW

PR #180, opened 2017-09-21
docs(#177): Add Jenkins integration section

### LINKED ISSUE #177: Add section about how to use it in Jenkins
##### Issue Overview

Add a section explaining how to use Smart Testing in Jenkins Freestyle and Jenkins Pipeline

lordofthejars:
Add Jenkins integration section

Fixes #177


dipak-pawar:
@lordofthejars how about adding `ST with CI server` heading? then we can give more lights on what we provide with `scm.range.head`, `scm.range.tail`, `scm.last.chages`& How user can use it. So user can try to configure ST depending on his CI server. So user can understand it's not only about Jenkins, he can use it for any CI server. 

Maybe we can add Jenkins as example here. WDYT?

lordofthejars:
Well it is not necessary to explain the flags again, we have already done in configuration section, and in case of CIs the `scm.last.changes` is not useful at all.

I thought the same about the title bust since the majority of people is using Jenkins and so far the only real problem users might find is when they use Jenkins Pipeline. For this reason I just added this section.

bartoszmajsak:
> how about adding ST with CI server heading?

As soon as we add travis for example this would make sense. For the moment it's fine like that.

bartoszmajsak:
Reworked few bits and merging upstream.

PR #181, opened 2017-09-21
fix(NPE in test) Added missing mock in SmartTestingProviderTest

### LINKED ISSUE #179: NPE in SmartTestingProviderTests when executing from the IDE
##### Issue Overview

Smart testing provider tests return NPE when provider invocation not set or set to null.
 
##### Expected Behaviour

System property `basedir` should not be null.

##### Current Behaviour

System property `basedir` is not set.

##### Steps To Reproduce
1. Execute SmartTestingProviderTest `when_get_suites_is_called_then_same_list_of_classes_will_be_returned` and `test_when_invoke_is_called_with_null` from Intellij

MatousJobanek:
Fixes #179


bartoszmajsak:
Thanks!

PR #183, opened 2017-09-21
Introducing Yaml configuration file

### LINKED ISSUE #113: Introduce YAML configuration file
In order to simplify user experience, we should introduce configuration file and use system properties as special-cases overwrite strategy instead of default approach as we are doing now.

This task is to map all generic configuration we have in place to a YAML stored in the root of the project named `smart-testing.yaml`. 

- [x] file becomes a prerequisite/enabler of our extension if `smart.testing.strategies` property not used
- [x] keep the same defaults as with properties
- [ ] update documentation
- [ ] update `Test Bed` to follow new approach

dipak-pawar:
Added support for configuration file at parent location of multi-module project.

- Introducing `smart-testing.yml`
```yml
mode: ordering
strategies: new, changed, affected
applyTo: surefire
debug: true
disable: false
report:
    enable: true
scm:
    range:
      head: HEAD
      tail: HEAD~3
```
- Introducing `Configuration.loadPrecalculated()` method to create configuration dumped at `.smart-testing/temporary/smart-testing.yml`
- Modified test-bed to use config file
 - Introducing fluent api to create config file using configuration object.
 ```java
 final Configuration configuration = Configuration.builder()
                .mode(SELECTING)
                .strategies(AFFECTED.getName())
                .scm()
                    .range()
                        .head("HEAD").tail("HEAD~")
                        .build()
                    .build()
                .build();
                    .withConfiguration(configuration)
                .createConfigFile()
            .enable();
```

Fixes #113

Note: Documentation needs to update.

### CODE REVIEW

**Comment by bartoszmajsak** [comment]
Thanks @dipak-pawar, I will go over it tomorrow morning (so in 12h from now ;)).

As you will be working on updating the documentation, which ideally should come as part of this PR, don't forget to cross-check all the constants which we are referring from there: https://github.com/arquillian/smart-testing/pull/183/files#diff-a1f054f1bd53a1e8081bd95bb2ce9447R5

Also we could think about using sample yaml config file for two purposes

  * for testing if we can unmarshall properly
  * the same file can be used as a living documentation - we can use tags to import it directly to the documentation, so when we introduce changes in the code and thus the file, we will have the documentation up-to-date. See how we leverage it for `Test Bed` for example. Including [`GitClone`](https://github.com/arquillian/smart-testing/blame/39c3e2cdf5e378953708488e7bbd346983031826/docs/developing.adoc#L20) directly from the [source](https://github.com/arquillian/smart-testing/blob/57e29f0d7d967ac07749a4d918d2e5ec564fbedb/functional-tests/git-rules/src/main/java/org/arquillian/smart/testing/rules/git/GitClone.java#L13-L15)

You can also use [footnotes](https://github.com/arquillian/smart-testing/blame/master/docs/developing.adoc#L83) to explain easily particular bits of the code.

**Comment by bartoszmajsak** [comment]
```java
Configuration configuration = Configuration.builder()
               .mode(SELECTING)
               .strategies(AFFECTED.getName())
               .scm(Scm.builder().lastChanges("2").build())
           .build();
```

how about `scm()` returning builder instead of accepting it as a parameter? Would make it a bit more fluent.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,343 @@
...
+
+    public boolean isApplyToDefined() {
+        return this.applyTo != null;
+    }
+
+    public static Builder builder() {
+        return new Configuration.Builder();
+    }
+
+    public static class Builder {
```
MatousJobanek: If I understand the purpose of the builders correctly, they are used only for test purposes. If so, then I would move them to the `TestBed` implementation to keep the class clear and to not expose test logic from the normal class.

dipak-pawar: @MatousJobanek If we move it to TestBed then we can't use it in unit test for [ConfigurationTest](https://github.com/arquillian/smart-testing/pull/183/files#diff-09137b69519fe19b2f7c61862edeba9b) & [ConfigurationUsingPropertyTest](https://github.com/arquillian/smart-testing/pull/183/files#diff-ebaf055f2f7537b003d0b8833f51d445). Also as we are keeping it here it's quite easy to reference things in future if we want to modify it.

dipak-pawar:
> how about scm() returning builder instead of accepting it as a parameter? Would make it a bit more fluent.

:+1: 

> **Passing strategies**
`.strategies(AFFECTED.getName())` why not AFFECTED anymore? Or is it just a mistake in the description?

No it's not mistake in the description. This is how I implemented it. I kn ow it's little weird.
We can add it.But the problem is we are using `Strategy` class from test-bed. & we have builder in `core` module. Either we have to Strategy to `core` or the other way around. But I don't want to use another way. as I won't use builder to test anything in core model if I move all builders to test-bed.

Also I am not sure it is good way to move `Strategy` class to `core`. So I implemented this way.

bartoszmajsak:
> Also I am not sure it is good way to move Strategy class to core. So I implemented this way.

Right, we keep strategies based on strings, not types, as we might want to open it up for new ones, when types would be pretty much blocking it.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
MatousJobanek: I'm still not a fan of this. If it is a test logic, then I would move it to the test-bed module or make it as hidden as possible (package protected). 
If you want to open the API to the world, then you should also verify that the builder works properly - create a test for the builder. 
If you moved it to the test-bed, then in Core you could use different approach of creating the object:
* reading a file
* using the set methods (that you have there anyway)
* configuring the object by providing `Properties` object - that would make more sense exposing it than the static builder API

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,279 @@
...
+    public static Configuration load() {
+        final File parent = Paths.get("").toAbsolutePath().toFile();
+        final File[] files = parent.listFiles((dir, name) -> name.equals(SMART_TESTING_YML) || name.equals(SMART_TESTING_YAML));
+
+        Map<String, Object> yamlConfiguration = new LinkedHashMap<>();
+
+        if (files != null) {
+            if (files.length == 0) {
+                logger.info(
+                    "Config file `" + SMART_TESTING_YAML + "` OR `" + SMART_TESTING_YAML + "` is not found. Using system properties to load "
```
MatousJobanek: The usage of the same constant `SMART_TESTING_YAML` is correct?

MatousJobanek: I guess that one of them should be `SMART_TESTING_YML`

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,279 @@
...
+    public Scm getScm() {
+        return scm;
+    }
+
+    public void setScm(Scm scm) {
+        this.scm = scm;
+    }
+
+    public static Configuration load() {
+        final File parent = Paths.get("").toAbsolutePath().toFile();
```
MatousJobanek: I would be great to have option to define the root dir as a parameter of the method `load()`

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,279 @@
...
+
+        configuration.mode = RunMode.valueOf(overWriteSystemProperty(yamlConfiguration, "mode", SMART_TESTING_MODE, DEFAULT_MODE).toUpperCase());
+
+        final String strategies = overWriteSystemProperty(yamlConfiguration, "strategies", SMART_TESTING, DEFAULT_STRATEGIES);
+
+        if (containsAnyStrategy(strategies)) {
+            configuration.strategies = strategies.split("\\s*,\\s*");
+        }
+
+        configuration.disable = Boolean.valueOf(overWriteSystemProperty(yamlConfiguration, "disable", SMART_TESTING_DISABLE, "false"));
```
MatousJobanek: The default values `false` are not necessary for the `Boolean.valueOf`

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,279 @@
...
+        logger.warn("Using configuration from " + SMART_TESTING_YML);
+
+        return Arrays.stream(files)
+            .filter(file -> file.getName().equals(SMART_TESTING_YML))
+            .map(File::toPath)
+            .findFirst()
+            .get();
+    }
+
+    private static Configuration parseConfiguration(Map<String, Object> yamlConfiguration) {
```
MatousJobanek: This whole method is a little bit chaotic to my taste. I'm slightly afraid of adding additional props and also what it will look like when we support also configuration per strategy or when the size of parameters will grow. 
Have you considered my comment: https://github.com/arquillian/smart-testing/issues/113#issuecomment-328831162
Similar approach could be also useful for you - having one class "Configuration" and go through all variables stored in this class, based on the type and name read relative parameters from the file and store inside of the class instance. If the variable is something different than array/list/string/int/boolean and similar (in other words it is some additional object such as `Report`), then take this new class and do the same stuff also for this one... 
You know, some kind of generic engine that takes the yaml file and any Object and based on the structure read and fill corresponding values. 
Then, adding/removing parameters would be as easy as modifying the class structure.
Something similar I do for browser options in Drone: https://github.com/arquillian/arquillian-extension-drone/blob/master/drone-webdriver/src/main/java/org/jboss/arquillian/drone/webdriver/factory/CapabilitiesOptionsMapper.java
I think that you could find similar code also in Arquillian Core or other project...

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/SmartTestingProviderTest.java
```diff
@@ -101,6 +113,17 @@ public void test_when_invoke_is_called_with_whole_set_of_classes() throws Except
...
 
+    @BeforeClass
+    public static void dumpConfiguration() {
+        Configuration.load().dump();
+    }
+
+    @AfterClass
+    public static void purgeConfiguration() {
+        final Path tempDirPath = Paths.get(Paths.get("").toAbsolutePath().toString(), SMART_TESTING_WORKING_DIRECTORY_NAME);
+        FileSystemOperations.deleteDirectory(tempDirPath, true);
```
MatousJobanek: wouldn't be better to use `TemporaryFolder` rule?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -54,17 +59,38 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
...
             calculateChanges();
             Runtime.getRuntime().addShutdownHook(new Thread(() -> purgeLocalStorage(session)));
         } else {
             logStrategiesNotDefined();
         }
     }
 
+    private void storeConfiguration(MavenSession session) {
+        final Path path = configuration.dump();
+        session.getAllProjects().forEach(mavenProject -> copyConfigurationFile(mavenProject.getModel(), path.toFile()));
```
MatousJobanek: We already go through all projects in `configureExtension(session, configuration);` - don't loop twice

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,279 @@
...
+        try(FileReader fileReader = new FileReader(configFile)) {
+            final Yaml yaml = new Yaml();
+            return yaml.loadAs(fileReader, Configuration.class);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to load configuration from file " + configFile.getPath(), e);
+        }
+    }
+
+    public Path dump() {
+        final LocalStorageFileAction configFile = new LocalStorage(Paths.get("").toFile())
```
MatousJobanek: setting path to the root dir should be done using a parameter in `dump()`;

MatousJobanek: The same can be applied also to the other methods

MatousJobanek: mainly the `loadPrecalculated()` one - otherwise it won't work in the surefire forked mode

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,279 @@
...
+        final File parent = Paths.get("").toAbsolutePath().toFile();
+        final File configFile = new LocalStorage(parent).duringExecution().temporary().file(SMART_TESTING_YML).getFile();
+
+        return loadConfigurationFromFile(configFile);
+    }
+
+    static Configuration loadConfigurationFromFile(File configFile) {
+        try(FileReader fileReader = new FileReader(configFile)) {
+            final Yaml yaml = new Yaml();
+            return yaml.loadAs(fileReader, Configuration.class);
```
MatousJobanek: now, I have noticed this line... Why don't you use this method also for loading the yaml file configured by the user? 
related to my comment: https://github.com/arquillian/smart-testing/pull/183/files#r140521368

core/src/main/java/org/arquillian/smart/testing/hub/storage/local/FileSystemOperations.java
```diff
@@ -62,7 +62,7 @@ static void deleteFile(Path path, boolean catchException) {
         }
     }
 
-    static void deleteDirectory(Path directory, boolean catchException) {
+    public static void deleteDirectory(Path directory, boolean catchException) {
```
MatousJobanek: doesn't have to be public - see my comment: https://github.com/arquillian/smart-testing/pull/183/files#r140523705

core/src/test/java/org/arquillian/smart/testing/impl/TestStrategyApplierUsingPropertyTest.java
```diff
@@ -135,4 +140,9 @@ private TestExecutionPlannerLoader prepareLoader(final Set<Class<?>> testsToRun,
     private Set<Class<?>> createTestsToRun(Class... testClasses) {
         return new LinkedHashSet<>(Arrays.<Class<?>>asList(testClasses));
     }
+
+    private String getBaseDir(Class<?> aClass) {
```
MatousJobanek: do we really need this?

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
dipak-pawar: Here we are not using `false` as second parameter to `Boolean.valueof`, where as we are using it for `overwriteSystemPreoperty` method

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/SmartTestingProviderTest.java
dipak-pawar: We can't use it `TemporaryFolder` Rule because it has it's own format for naming `given_name_time_stamp`. We need to dump configuration to `${projectdir}/.smart-testing/temporary/smart-testing.yml` So `Configuration.loadPrecalculated(File projectDir)` method can load configuration from that predefined location.

MatousJobanek: setting the path to the project root directory should solve this problem - see my comment: https://github.com/arquillian/smart-testing/pull/183#discussion_r140525783

core/src/test/java/org/arquillian/smart/testing/impl/TestStrategyApplierUsingPropertyTest.java
dipak-pawar: yes we need this, otherwise we have to hard code this value to "core" to work it from `IDEA` & terminal.

lordofthejars:
Are we far from merging this?

dipak-pawar:
> Are we far from merging this?

We are not too far from merging. Needs final review from @MatousJobanek

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,267 @@
...
+            .filter(file -> file.getName().equals(SMART_TESTING_YML))
+            .map(File::toPath)
+            .findFirst()
+            .get();
+    }
+
+    private static Configuration parseConfiguration(Map<String, Object> yamlConfiguration) {
+        final Configuration configuration = ObjectMapper.mapToObject(Configuration.class, yamlConfiguration);
+        ObjectMerger<Configuration> configurationMerger =
+            new ObjectMerger<>(Configuration.fromSystemProperties(), configuration, Configuration.withDefaultValues());
```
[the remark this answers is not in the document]
dipak-pawar: yes that's what we are doing in `ObjectMerger.merge()`

dipak-pawar:
> One general question, is configuration information written as report?

no it's not. but as this story is too big, I think it's good to create small task for it, after we merge this.

core/src/main/java/org/arquillian/smart/testing/ObjectMapper.java
```diff
@@ -0,0 +1,133 @@
...
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class ObjectMapper {
+
+    public static <T> T mapToObject(Class<T> aClass, Map<String, Object> map) {
```
MatousJobanek: Do we really need to have it public? 
This is used only by the `Configuration` class, so I would move it to the `configuration` package and make it package protected. If we need it also in some other cases, then we can open it to the world.

core/src/main/java/org/arquillian/smart/testing/ObjectMerger.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+public class ObjectMerger<T> {
+
+    private T first;
+    private T second;
+    private T third;
+
+    public ObjectMerger(T first, T second, T third) {
```
MatousJobanek: Same as for `ObjectMapper` - do not expose our implementation details if not necessary

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    public static Configuration load(File projectDir) {
+        final File[] files =
+            projectDir.listFiles((dir, name) -> name.equals(SMART_TESTING_YML) || name.equals(SMART_TESTING_YAML));
+
+        Map<String, Object> yamlConfiguration = new LinkedHashMap<>();
+
+        if (files != null) {
+            if (files.length == 0) {
```
MatousJobanek: these two `if`s should be merged, to handle (log) also the case when `files` is `null`

core/src/main/java/org/arquillian/smart/testing/ObjectMapper.java
```diff
@@ -0,0 +1,133 @@
...
+
+    private static Object convert(Method method, Object mappedValue) {
+        Class<?> parameterType = method.getParameterTypes()[0];
+        if (parameterType.isArray()) {
+            return handleArray(parameterType.getComponentType(), mappedValue);
+        } else if (parameterType.isEnum()) {
+            return handleEnum(method, mappedValue);
+        } else if (parameterType.isAssignableFrom(List.class)) {
+            return handleList(method, mappedValue);
+        } else if (!parameterType.isAssignableFrom(Map.class) && Map.class.isAssignableFrom(mappedValue.getClass())) {
```
MatousJobanek: This if should check other configuration objects such as `Report` and `Scm`. Personally, I would create an interface (eg. `ConfigurationSection`) that all configuration classes would implement it... Then you could check if the parameter type is assignable from the interface.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        if (System.getProperty(SMART_TESTING_DEBUG) != null) {
+            configuration.debug = Boolean.valueOf(System.getProperty(SMART_TESTING_DEBUG));
+        }
+
+        configuration.applyTo = System.getProperty(SMART_TESTING_APPLY_TO);
+
+        configuration.report = Report.fromSystemProperties();
+        configuration.scm = Scm.fromSystemProperties();
```
MatousJobanek: To be honest, I still don't like this tightly coupled way of merging the configurations. I just cannot imagine the solution for separated configurations for each strategy. These strategy related configs should be part of the strategy modules and this doesn't allow it. 
I would go for the solution I was proposing you some time ago - having kind of list of objects that contains all necessary information: param name, system property, default value. Then, based on the type (implementing an interface), you could call the method that returns the list of the mentioned objects. This would simplify the logic of getting the system properties (making it generic for all parameters), you could get rid of the `ObjectMapper` class and also simplify this [if](https://github.com/arquillian/smart-testing/pull/183/files#diff-550791242a7d1b9809c8c5ed3e9ac5ecR60). At the same time you could reuse the [convertToType](https://github.com/arquillian/smart-testing/pull/183/files#diff-550791242a7d1b9809c8c5ed3e9ac5ecR94) method for getting the system properties. 
We can have a bjn tomorrow and I can show you what I'm talking about. Some pair programming would make sense as well - I have already some bits implemented.

core/src/main/java/org/arquillian/smart/testing/configuration/Range.java
```diff
@@ -0,0 +1,24 @@
+package org.arquillian.smart.testing.configuration;
+
+public class Range {
```
MatousJobanek: Why do we have a separated class for the `Range`? What is the difference between having `Scm` + `Range` and only `ScmRange`?

core/src/test/resources/dumped-smart-testing.yml
```diff
@@ -0,0 +1,9 @@
+!!org.arquillian.smart.testing.configuration.Configuration
```
MatousJobanek: the directory ` core/src/test/resources` is getting to be chaotic - move the config files to separated subdirectory.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
MatousJobanek: here is the implementation I'm talking about. https://github.com/MatousJobanek/smart-testing/commit/17f94b647ff0feb8b211d8fee9e8d99972d83e62 You can take a look so you know what I have in my mind. It needs some refactoring, removing old implementation classes and methods (such as `fromSystemProperties`) and changes in test classes, but it should work. 
I believe that it will make it simpler and easier to use. 
In addition, I haven't found (in your code) the solution for reading the list of parameters & it's default methods - for Che integration we were talking about. Having the list of configuration items for every configuration section should make it easier to implement.
Let's discuss it in the morning

core/src/main/java/org/arquillian/smart/testing/configuration/Range.java
dipak-pawar: The main idea to keep both classes separated to keep same structure in system properties  (`scm.range.head` & `scm.range.tail`) & in yaml
```yaml
scm:
    range:
      head: HEAD
      tail: HEAD~3
```

Otherwise we have to give option in config file as 
```yaml
scmRange:
      head: HEAD
      tail: HEAD~3
```

Also for this second option, we need to think about `lastChanges` option, as it's defined as `scm.lastChanges` as system prioperty for now. So either we need to change it to `scmRange.lastChanges` if we decided to go with second option

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
dipak-pawar: you can find solution for reading the list of parameters in [SchemaGenerator](https://github.com/arquillian/smart-testing/pull/183/files#diff-01b186fdde2ad7c30adc1cfbeb7b08f9)

MatousJobanek: Yeah, I missed that class. Probably because it is package protected and as I need to call it from Che logic I didn't pay much attention to it.
Is there some way of getting default values?

core/src/main/java/org/arquillian/smart/testing/api/SmartTesting.java
```diff
@@ -23,7 +24,7 @@
      * @return An instance of {@link ConfiguredSmartTesting} class
      */
     public static ConfiguredSmartTesting with(TestVerifier testVerifier) {
-        return with(testVerifier, Configuration.load());
+        return with(testVerifier, Configuration.loadPrecalculated(Paths.get("").toFile()));
```
MatousJobanek: As a default behavior, I would load the file from main directory - the original one. `Configuration.load(Paths.get("").toFile())`
The precalculated is Maven(surefire) specific.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
dipak-pawar: You mean default values for each property. for this we need to keep some map with `key` as `property(class's instance field)` & `value` as `default value` for that field.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,234 @@
...
+            .map(File::toPath)
+            .findFirst()
+            .get();
+    }
+
+    private static Configuration parseConfiguration(Map<String, Object> yamlConfiguration) {
+        return ObjectMapper.mapToObject(Configuration.class, yamlConfiguration);
+    }
+
+    private static boolean containsAnyStrategy(String strategies) {
```
MatousJobanek: not used

core/src/main/java/org/arquillian/smart/testing/api/SmartTesting.java
dipak-pawar: No this `with` method is not used anywhere. We are using with `(TestVerifier testVerifier, Configuration configuration)`. You can look [here](https://github.com/arquillian/smart-testing/pull/183/files/99430c1dc61dcd34884d69229d0eaf08c30e1210#diff-e92614f0bf12fbca05272c1456ecba67R67). I am not sure about `che` for this. Is it used or not?

Also one more thing, as with default pointing to `rootDir` other submodule won't have this configuration file in it's respective directory, So it won't work

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -0,0 +1,177 @@
...
+
+        List<ConfigurationItem> configItems = instance.registerConfigurationItems();
+
+        Arrays.stream(aClass.getMethods()).filter(ObjectMapper::isSetter)
+            .forEach(method -> invokeMethodWithMappedValue(configItems, method, instance, map));
+
+        return instance;
+    }
+
+    private static <T> void invokeMethodWithMappedValue(List<ConfigurationItem> configItems, Method method, T instance,
```
MatousJobanek: maybe break this method not to be so long - something like separating: https://gist.github.com/MatousJobanek/9e79ba640b3b98361c2afca283aa867a

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -0,0 +1,177 @@
...
+            return handleEnum(method, mappedValue);
+        } else if (parameterType.isAssignableFrom(List.class)) {
+            return handleList(method, mappedValue);
+        } else if (ConfigurationSection.class.isAssignableFrom(parameterType)) {
+            if (parameterType.isAssignableFrom(mappedValue.getClass())) {
+                return mappedValue;
+            }
+            return mapToObject((Class<ConfigurationSection>) parameterType, (Map<String, Object>) mappedValue);
+        } else if (parameterType.isAssignableFrom(mappedValue.getClass())) {
+            return mappedValue;
```
MatousJobanek: now looking at it - moving the 
```java
} else if (parameterType.isAssignableFrom(mappedValue.getClass())) {
            return mappedValue;
```
on the line 93 would eliminate the nested if on the lines 94-96

core/src/main/java/org/arquillian/smart/testing/configuration/Scm.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static org.arquillian.smart.testing.scm.ScmRunnerProperties.DEFAULT_LAST_COMMITS;
+import static org.arquillian.smart.testing.scm.ScmRunnerProperties.HEAD;
+import static org.arquillian.smart.testing.scm.ScmRunnerProperties.SCM_LAST_CHANGES;
+
+public class Scm implements ConfigurationSection {
+
+    private Range range;
+
+    public void setLastChanges(String lastChanges) {
```
MatousJobanek: not used

dipak-pawar: It's used by `ObjectMapper`

core/src/main/java/org/arquillian/smart/testing/api/SmartTesting.java
MatousJobanek: It had been used somewhere in ST, but not now obviously. 
BTW, the API is not intended to be used for one parent module expecting that it will invoke all submodules.
Nevermind, leave it as it is.

core/src/main/java/org/arquillian/smart/testing/configuration/Scm.java
MatousJobanek: now I got your point... interesting idea - I didn't think about it in this way. 👍

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -0,0 +1,234 @@
...
+import org.arquillian.smart.testing.hub.storage.local.LocalStorage;
+import org.arquillian.smart.testing.hub.storage.local.LocalStorageFileAction;
+import org.yaml.snakeyaml.Yaml;
+
+public class Configuration implements ConfigurationSection {
+
+    private static final Logger logger = Logger.getLogger();
+
+    public static final String DEFAULT_MODE = "selecting";
+    public static final String DEFAULT_STRATEGIES = "";
```
MatousJobanek: both variables `DEFAULT_MODE` and `DEFAULT_STRATEGIES` are not used

dipak-pawar: `DEFAULT_MODE` is used in documentation. `DEFAULT_STRATEGIES` are used anywhere

core/src/main/java/org/arquillian/smart/testing/configuration/Scm.java
MatousJobanek: add the annotation `@SuppressWarnings("unused")`

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
dipak-pawar: no they both covers different scenarios. If you run `ObjectMapperTest` you will get clear idea why I did this way

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationItem.java
```diff
@@ -0,0 +1,40 @@
+package org.arquillian.smart.testing.configuration;
+
+public class ConfigurationItem {
+
+    private final String paramName;
+    private String systemProperty;
+    private Object defaultValue;
+
+    public ConfigurationItem(String paramName, String systemProperty, Object defaultValue) {
```
MatousJobanek: the constructors would deserve Javadoc - mainly describing the `paramName` - that it needs to match the set methods

core/src/main/java/org/arquillian/smart/testing/configuration/Report.java
```diff
@@ -0,0 +1,48 @@
...
+
+import static org.arquillian.smart.testing.configuration.Configuration.SMART_TESTING_REPORT_ENABLE;
+import static org.arquillian.smart.testing.report.SmartTestingReportGenerator.REPORT_FILE_NAME;
+import static org.arquillian.smart.testing.report.SmartTestingReportGenerator.TARGET;
+
+public class Report implements ConfigurationSection {
+
+    private boolean enable;
+    private String name = REPORT_FILE_NAME;
+    private String dir = TARGET;
```
MatousJobanek: if the defaults are in `registerConfigurationItems` then they don't have to be here... just a detail...

core/src/main/java/org/arquillian/smart/testing/SchemaGenerator.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private void addClassType(Class<?> aClass, Map<String, Object> properties) {
+        properties.put(TYPE, aClass.getSimpleName().toLowerCase());
+    }
+
+    private void addClassDetails(Class<?> aClass, Map<String, Object> map) {
+        map.put(TYPE, "object");
+        map.put("javatype", aClass.getName());
+
+        final List<Field> list = Arrays.stream(aClass.getDeclaredFields())
```
MatousJobanek: here you get the information about used params from the declared fields, but when we load the config, we use setter methods. Eg the `setLastChanges` won't be covered here.

core/src/main/java/org/arquillian/smart/testing/SchemaGenerator.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+class SchemaGenerator {
```
MatousJobanek: if the SchemaGenerator is related to configuration, then move it there - into the same package

core/src/test/resources/configuration/smart-testing-1.yml
```diff
@@ -0,0 +1,12 @@
+mode: ordering
```
MatousJobanek: I would welcome some better suffix then just `1`

docs/configuration.adoc
```diff
@@ -57,15 +57,15 @@ instead of *new* ones.
...
 
 ==== Affected
 
 `Affected` strategy uses a different approach to choose what are the *important* tests to run first (_ordering_) or filtered (_selecting_).
 This strategy also relies on _SCM_ information but in this case it retrieves any *new* or *modified* business class
 between commits range and local changes as well.
 
-TIP: `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="COMMIT"]`, `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="PREVIOUS_COMMIT"]` and `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]` Java system properties are valid as well.
+TIP: `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="SCM_RANGE_HEAD"]`, `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="SCM_RANGE_TAIL"]` and `const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="DEFAULT_LAST_COMMITS"]` Java system properties are valid as well.
```
MatousJobanek: I guess that the `DEFAULT_LAST_COMMITS` is wrong - there should be the system property used...

docs/usage.adoc
```diff
@@ -8,22 +8,22 @@ or `maven-failsafe-plugin`.
...
+`mvn clean test -Dconst:core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java[name="SMART_TESTING"]="new, changed"` `mvn clean test -Dsmart.testing="new, changed"`  copyToClipboard:selectingChanged[]
 
 *You want to run all tests but given priority to the latest tests added or modified*
 
 [[orderingChanged]]
-`mvn clean test -Dconst:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING_MODE"]=ordering -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="LAST_COMMITS"]=1 -Dconst:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING"]="new, changed"`  copyToClipboard:orderingChanged[]
+`mvn clean test -Dconst:core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java[name="SMART_TESTING_MODE"]=ordering -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="DEFAULT_LAST_COMMITS"]=1 -Dconst:core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java[name="SMART_TESTING"]="new, changed"`  copyToClipboard:orderingChanged[]
```
MatousJobanek: same here with the `DEFAULT_LAST_COMMITS`

docs/jenkins.adoc
```diff
@@ -59,7 +59,7 @@ pipeline {
             steps {
                 script {
                     def scmVars = checkout scm
-                    sh "mvn -Dconst:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING"]='new, affected' -Dconst:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING_MODE"]=ordering -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="COMMIT"]=${scmVars.GIT_COMMIT} -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="PREVIOUS_COMMIT"]=${scmVars.GIT_PREVIOUS_COMMIT} test"
+                    sh "mvn -Dconst:core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java[name="SMART_TESTING"]='new, affected' -Dconst:core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java[name="SMART_TESTING_MODE"]=ordering -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="SCM_RANGE_HEAD"]=${scmVars.GIT_COMMIT} -Dconst:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="SCM_RANGE_TAIL"]=${scmVars.GIT_PREVIOUS_COMMI} test"
```
MatousJobanek: missing `T` in `${scmVars.GIT_PREVIOUS_COMMI}`

docs/refcard.adoc
```diff
@@ -42,17 +42,17 @@ a|`surefire`, `failsafe`
 |===
 |Property | Description | Default | Applicable Strategies
 
-a|`const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="PREVIOUS_COMMIT"]`
+a|`const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="SCM_RANGE_TAIL"]`
 |Set first commit id for inspecting changes
 a|`const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="HEAD"]~const:core/src/main/java/org/arquillian/smart/testing/scm/ScmRunnerProperties.java[name="DEFAULT_LAST_COMMITS"]`
```
MatousJobanek: again `DEFAULT_LAST_COMMITS`

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
MatousJobanek: I changed it to:
```java
        if (parameterType.isArray()) {
            return handleArray(parameterType.getComponentType(), mappedValue);
        } else if (parameterType.isEnum()) {
            return handleEnum(method, mappedValue);
        } else if (parameterType.isAssignableFrom(List.class)) {
            return handleList(method, mappedValue);
        } else if (parameterType.isAssignableFrom(mappedValue.getClass())) {
            return mappedValue;
        } else if (ConfigurationSection.class.isAssignableFrom(parameterType)) {
            return mapToObject((Class<ConfigurationSection>) parameterType, (Map<String, Object>) mappedValue);
        } else {
            return convertToType(parameterType, mappedValue.toString());
        }
```
and the tests are green

core/src/test/java/org/arquillian/smart/testing/impl/TestStrategyApplierUsingPropertyTest.java
```diff
@@ -135,4 +140,9 @@ private TestExecutionPlannerLoader prepareLoader(final Set<Class<?>> testsToRun,
     private Set<Class<?>> createTestsToRun(Class... testClasses) {
         return new LinkedHashSet<>(Arrays.<Class<?>>asList(testClasses));
     }
+
+    private String getBaseDir() {
```
MatousJobanek: you told me:

> yes we need this, otherwise we have to hard code this value to "core" to work it from IDEA & terminal.

I tried it with these options:
* `System.getProperty("user.dir")`
* `new File(".").getAbsolutePath()`
* removing the `in(getBaseDir())` completely

and in all cases the tests are green on both command line and IDE

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
MatousJobanek: As for the `SchemaGenerator` - maybe it would be good to remove it from this PR and finish it in a separated PR - we can do it together.
This PR is huge enough, so let's finish the main logic first.
I'll create an issue for it

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
MatousJobanek: then use the `DEFAULT_MODE` also here https://github.com/arquillian/smart-testing/pull/183/files#diff-348861c21efd28f5cc99117b2c67302aR110

docs/configfile.adoc
```diff
@@ -0,0 +1,52 @@
+== Configuration file
+
+You can use smart-testing.y(a)ml file which is present at `${project.dir}/smart-testing.yml`.
```
MatousJobanek: `smart-testing.y(a)ml` this doesn't look obvious in the doc. Specify both types or just write:

You can use smart-testing yaml file at `${project.dir}/smart-testing.yml` or with alternative suffix `${project.dir}/smart-testing.yaml`

MatousJobanek: now I have noticed that it is mentioned here https://github.com/arquillian/smart-testing/pull/183/files#diff-1dfff5bec990830d634a922b323a0661R18 again

docs/configfile.adoc
```diff
@@ -0,0 +1,52 @@
...
+copyToClipboard:config-file[]
+
+However you can overwrite all configuration options using system properties supported by Smart Testing. You can look <<_reference_card, references>> for all supported system properties.
+
+=== Configuration file reference
+The `const:core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java[name="SMART_TESTING_YML"]` file is a `YAML` file defining required configuration to configure Smart Testing.
+
+NOTE: You can use either a .yml or .yaml extension for this file.
+
+==== strategies
```
MatousJobanek: instead of creating a new titli for every param I would use the same approach of explaining parameters as in cube config: http://arquillian.org/arquillian-cube/#_configuration

MatousJobanek: you know, referencing from the config file example using numbers

core/src/test/java/org/arquillian/smart/testing/impl/TestStrategyApplierUsingPropertyTest.java
dipak-pawar: I know why it's passing because you added fix for this few days ago in #181

MatousJobanek:
Good job @dipak-pawar 👍 
Thank you for your great work 🥇

PR #186, opened 2017-09-22
fix(#165): TestVerifier is executed only in Java files

### LINKED ISSUE #165: Affected strategy analyze .class files when target dir not ignored
##### Issue Overview

Affected strategy analyze `.class` files when target dir not ignored from git (`.gitignore`) and this causes an exception: 

```
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
[ERROR] Unable to analyze source file /Users/alex/git/stdemo/target/classes/org/superbiz/A.class
[ERROR] 	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:665)
```
##### Expected Behaviour

`.class` files should be ignored.

##### Current Behaviour

Exception is thrown.

##### Steps To Reproduce
1. Clone https://github.com/lordofthejars/stdemo
2. Edit `.gitignore` and comment `target/` line using `#`
3. Run `mvn clean test -Dsmart.testing="affected"`

lordofthejars:

TestVerifier is executed only in Java files

- `TestVerifier` methods return false if given path is not a Java file

Fixes #165

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/api/TestVerifier.java
+++ b/core/src/main/java/org/arquillian/smart/testing/api/TestVerifier.java
@@ -16,6 +16,11 @@ public interface TestVerifier {
      * @return Whether the given path to a resource represents a test class or not
      */
     default boolean isTest(Path resource) {
+
+        if (!isJavaFile(resource)) {
+            return false;
+        }
+
         final String className = new ClassNameExtractor().extractFullyQualifiedName(resource);
         return isTest(className);
     }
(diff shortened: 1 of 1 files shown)


lordofthejars:
Probably when we rewrite everything this would be a class, but for this release I would leave as is

bartoszmajsak:
But now we have interface with default methods which effectively makes it a class...

lordofthejars:
Well actually some tests are implementing this interface for testing purposes, so changing to class requires also a rewrite of the tests. Also if next week we work on rewriting this part, would means change the tests again so it is like in one week do the same task twice.

bartoszmajsak:
How long the "reimplementation" will take?

lordofthejars:
No idea, it depends on how far we want to go. Maybe two days between designing reimplementing and testing

bartoszmajsak:
Ok, let's open an issue describing our findings for refactoring and refer to #165 so we can have a trace of ongoing effort. Can you take care of it?

lordofthejars:
@bartoszmajsak I am doing it right now

lordofthejars:
This issue is just a workaround, the final solution should be implemented as described in https://github.com/arquillian/smart-testing/issues/187

PR #189, opened 2017-09-22
fix(#185): Adds compatibility to Mac OSx install script

### LINKED ISSUE #185: Full support for OSX in install.sh
Following cases still don't work after fd52992e62f650200a0ff409ee541beb4e30f19b:

  * Install given version (`-v` or `--version` flag) - combination of `awk` and `sed` in `override_version` function results in odd `xml` 
  * Prompt (`read`) is ignored and just executes

lordofthejars:

Adds compatibility to Mac OSx install script

- Don't use `read` prompt option but `echo`
- Uses just `sed` to update version instead of `awk` + `sed`

Fixes #185


install.sh
```diff
@@ -113,18 +115,15 @@ function install_extension() {
 }
 
 function override_version() {
-    cat .mvn/extensions.xml | awk -v groupId="org.arquillian.smart.testing" -v version=$1 -v RS="</extension>" '
-      $0 ~ "<groupId>" groupId "</groupId>" {
-        sub("<version>.*</version>", "<version>" version "</version>")
-      }
-      {printf "%s", $0 RS}' | sed '$ d' > .mvn/extensions-new.xml
+    sed -e "s/<version>.*<\/version>/<version>$1<\/version>/g" .mvn/extensions.xml > .mvn/extensions-new.xml
```
bartoszmajsak: What if you have more than one extension registered already? This won't work in such a case.

bartoszmajsak: We have to fix this use case prior to merge

install.sh
```diff
@@ -113,18 +115,15 @@ function install_extension() {
 }
 
 function override_version() {
-    cat .mvn/extensions.xml | awk -v groupId="org.arquillian.smart.testing" -v version=$1 -v RS="</extension>" '
-      $0 ~ "<groupId>" groupId "</groupId>" {
-        sub("<version>.*</version>", "<version>" version "</version>")
-      }
-      {printf "%s", $0 RS}' | sed '$ d' > .mvn/extensions-new.xml
+    xsltproc --stringparam version $1 updateversion.xslt .mvn/extensions.xml > .mvn/extensions-new.xml
```
bartoszmajsak: Add ` command_exists xsltproc` around line `150`

bartoszmajsak:
Two issues:

  * `read` does not work under Linux now (exactly the same issue as in OSX with the current one which is on `master`
  * overwriting existing version of the extension (e.g. `0.0.1`) fails with

```shell
╰─ curl -sSL https://git.io/vdL33 | bash -s -- -v 2.0.0
Installing extension in .mvn/extensions.xml
Smart Testing Extension already registered with version 1.0.0 - overwriting with 2.0.0.
warning: failed to load external entity "updateversion.xslt"
cannot parse updateversion.xslt
Updated Smart Testing Extension to 2.0.0.
.mvn/extensions-unformatted.xml:1: parser error : Document is empty
```

PR #204, opened 2017-10-03
feat: Introduced additonal parallelism to the test suite execution

### LINKED ISSUE #126: Build time improvement
As we are adding more and more functional tests our build is taking quite a bit of time which will make it harder to have fast feedback loops (either locally or on the CI server). It bumped from ~4:30min to above 7 on my local machine:

```
[INFO] Reactor Summary:                                                                                                                       
[INFO]                                                                                                                                        
[INFO] Smart testing ...................................... SUCCESS [  0.727 s]                                                               
[INFO] smart-testing-core ................................. SUCCESS [ 13.491 s]                                                               
[INFO] junit-test-result-parser ........................... SUCCESS [  1.728 s]                                                               
[INFO] smart-testing-surefire-provider .................... SUCCESS [  7.337 s]                                                               
[INFO] smart-testing-strategy-changed ..................... SUCCESS [  4.464 s]                                                               
[INFO] smart-testing-strategy-affected .................... SUCCESS [  9.435 s]                                                               
[INFO] smart-testing-strategy-failed ...................... SUCCESS [  2.511 s]                                                               
[INFO] smart-testing-maven-lifecycle-extension ............ SUCCESS [  4.022 s]                                                               
[INFO] smart-testing-test-bed ............................. SUCCESS [07:17 min]                                                               
[INFO] ------------------------------------------------------------------------                                                               
[INFO] BUILD SUCCESS               
[INFO] ------------------------------------------------------------------------                                                               
[INFO] Total time: 07:49 min (Wall Clock)                              
[INFO] Finished at: 2017-09-05T22:17:52+02:00                          
[INFO] Final Memory: 40M/1385M     
[INFO] ------------------------------------------------------------------------ 
```

We should review our Test Bed tests in order to:

  * minimize redundant/overlapping tests
  * identify those which we could test in isolation (unit tests)
  * narrow the builds for only relevant modules, e.g. `.build("config/impl-base")` where necessary

MatousJobanek:
* the test-bed build uses Maven parallelism with 50 threads per cpu core
* the same is used for build invocation on Travis (could be used locally)
* the Travis is building in three virtual machines at once:
  * one for unit tests
  * one for functional tests - to separate unit tests and functional tests I had to introduce two profiles inside of the pom files
  * one for generating documentation - no to do it in the previous builds/vm twice
* speeding up Java startup using properties: -XX:+TieredCompilation -XX:TieredStopAtLevel=1

Fixes #126

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,18 +17,13 @@ cache:
   directories:
     - $HOME/.m2
 
-env:
-  - TEST_SUITE=functional TEST_BED_M2_HOME=/home/travis/.arquillian/mvn/apache-maven-3.3.9 GENERATE_DOC=1
-  - TEST_SUITE=units GENERATE_DOC=1
-  - GENERATE_DOC=0
-
 before_install:
   - BRANCH=${TRAVIS_PULL_REQUEST_BRANCH:-$TRAVIS_BRANCH}
   - MODIFIED_DOCS=$(git diff --name-only $TRAVIS_COMMIT_RANGE | grep -E 'README.adoc|^docs/.*.adoc$' | wc -l)
   - COMMITS_WITH_FORCED_GEN=$(git log --format=%B  $TRAVIS_COMMIT_RANGE | grep -i "\[gen docs\]" | wc -l)
   - git describe --tags --exact-match HEAD >/dev/null 2>&1 && TAGGED=0 || TAGGED=1
-  - '[ $TAGGED -eq 0 ] || [ $COMMITS_WITH_FORCED_GEN -gt 0 ] || [ $MODIFIED_DOCS -ge 1 ] && [ $GENERATE_DOC -eq 0 ] && GENERATE_DOC=0 || GENERATE_DOC=1'
-  - '[ "${BRANCH}" == "master" ] && [ "${TRAVIS_REPO_SLUG}" == "arquillian/smart-testing" ] || [ ! -z "${TRAVIS_PULL_REQUEST// }"  -a "${TRAVIS_PULL_REQUEST}" != "false" ] && [ $GENERATE_DOC -eq 0 ] && GENERATE_DOC=0 || GENERATE_DOC=1'
+  - '[ $TAGGED -eq 0 ] || [ $COMMITS_WITH_FORCED_GEN -gt 0 ] || [ $MODIFIED_DOCS -ge 1 ] && GENERATE_DOC=0 || GENERATE_DOC=1'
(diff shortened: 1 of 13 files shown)


lordofthejars:
So the idea is to merge both PRs right? And thank you.

MatousJobanek:
yup - they are different solutions => separated PRs.
Yours is important as well - to minimize the test suite

MatousJobanek:
Yeah, I'm not a huge fan of profiles either so I completely understand your point/feeling. 
On the other hand, I would really welcome a simple way of executing only unit tests - just an example of the practical usage on the command line.
In addition, there is no change of the usage - you can still use the original standard command of building/testing the whole project.
Anyway, I'll rebase it with the latest changes, compare the time and if there is a improving in the build time then I'll add the explanation and finish the PR.

- [ ] rebase 
- [ ] compare the build time
- [ ] add documentation explanation

bartoszmajsak:
> I would really welcome a simple way of executing only unit tests - just an example of the practical usage on the command line.

`mvn test` won't work here?

MatousJobanek:
> mvn test won't work here?

nope, that will execute the whole test suite

MatousJobanek:
This brings an interesting question. How about removing the profiles and configure the failsafe plugin for functional tests? Then it would be possible to run only unit tests using `mvn clean test`, we could still have separated builds in Travis and we would get rid of the profiles...

MatousJobanek:
The build times are pretty weird - in my Travis account, the build [takes 15:41](https://travis-ci.org/MatousJobanek/smart-testing/builds/287471335) but here on [arquillian's 19:17](https://travis-ci.org/arquillian/smart-testing/builds/287471368)
Similarly look also the Maven builds of the project:
on my travis [account 13:53](https://travis-ci.org/MatousJobanek/smart-testing/builds/287471335#L4055)
on Arquillian's [account 17:26](https://travis-ci.org/arquillian/smart-testing/builds/287471368#L4062)
The differences are quite huge and seem to be stable. Any idea what could be the cause? How about your accounts? Are they also faster? @bartoszmajsak @lordofthejars @hemanik @dipak-pawar

bartoszmajsak:
> This brings an interesting question. How about removing the profiles and configure the failsafe plugin for functional tests? Then it would be possible to run only unit tests using mvn clean test, we could still have separated builds in Travis and we would get rid of the profiles...

That's exactly what was the bottom line of my short question :)

bartoszmajsak:
> The build times are pretty weir

Could that be that there are other projects under arq organization which had running builds at the same time? Might be that we have resources constraints per organization on Travis.

bartoszmajsak:
What do we do with this work @MatousJobanek ?

MatousJobanek:
I'll park it for a while and return back to it when I have time

bartoszmajsak:
I re-opened and added "on hold" label so we won't miss it.

PR #205, opened 2017-10-04
Remove redundant code & fix string concatenation

dipak-pawar:

- Removes redundant code for ST configuration in pom.xml for initial approach.
- Corrected string concatenation to log correct message.

### DIFF
--- a/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
+++ b/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
@@ -1,7 +1,6 @@
 package org.arquillian.smart.testing.mvn.ext;
 
 import java.util.List;
-import java.util.StringJoiner;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


MatousJobanek:
@dipak-pawar could you please rebase it with latest changes from the master branch?

lordofthejars:
@MatousJobanek it seems it is still mergable

dipak-pawar:
@lordofthejars I rebased it. So it's mergeable

lordofthejars:
@dipak-pawar Ok now I only see the original commit. But anyway, when Matous gives the ok, let's merge

MatousJobanek:
Thank you @dipak-pawar

bartoszmajsak:
I have a feeling this concatenation was already fixed before. Was it lost in some other merge or is it a different place where we had a similar mistake?

dipak-pawar:
> I have a feeling this concatenation was already fixed before. Was it lost in some other merge or is it a different place where we had a similar mistake?

@bartoszmajsak yes I submitted #130 with this fix. But we decided to not go with that change which contains more than concatenation fix. So I closed that PR. Hence we have this change here.

PR #206, opened 2017-10-04
Adds information about surefire.skipAfterFailureCount

dipak-pawar:
Educate users about `skipAfterFailureCount` property to get fast feedback loop.

- Updated documentation about `skipAfterFailureCount`. Ref https://github.com/arquillian/smart-testing/issues/96#issuecomment-329310444

### DIFF
--- a/docs/configuration.adoc
+++ b/docs/configuration.adoc
@@ -14,6 +14,8 @@ The `const:core/src/main/java/org/arquillian/smart/testing/configuration/Configu
 
 In order to define the mode use `const:core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java[name="SMART_TESTING_MODE"]` Java system property with either one.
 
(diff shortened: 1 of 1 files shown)


docs/configuration.adoc
```diff
@@ -14,6 +14,8 @@ The `const:core/src/main/java/org/arquillian/smart/testing/Configuration.java[na
 
 In order to define the mode use `const:core/src/main/java/org/arquillian/smart/testing/Configuration.java[name="SMART_TESTING_MODE"]` Java system property with either one.
 
+IMPORTANT: To get fast feedback loop, you can tell *Smart Testing* to skip after `N` failures by setting system property `surefire.skipAfterFailureCount` to N. Read `http://maven.apache.org/surefire/maven-surefire-plugin/examples/skip-after-failure.html[skipAfterFailureCount]` for more information.
```
lordofthejars: I will not say "tell *Smart Testing* to skip ..." sentence basically because it seems that is something provided by us when the reality is provided by surefire. So what I would say is that this is something provided by surefire, then put an snippet of plugin configuration and warn the user that there are some limitations regarding reusing JVM. 

Basically what I am thinking is that if someone finds a bug on `skipAfterFailureCount` they will report as ST issue.

MatousJobanek:
Great explanation! Thanks @dipak-pawar  👍

PR #210, opened 2017-10-05
Now generating report if St or maven debug enabled.

### LINKED ISSUE #144: Should produce reports when debug mode enabled
##### Issue Overview

When we enable debug mode we should also produce Smart Testing [execution reports](http://arquillian.org/smart-testing/#_reports) during the execution.

dipak-pawar:
Generating report if `smart.testing.debug` OR `mvnDedug` is enabled.

Fixes #144


core/src/main/java/org/arquillian/smart/testing/api/SmartTesting.java
```diff
@@ -35,11 +35,12 @@ public static ConfiguredSmartTesting with(TestVerifier testVerifier) {
...
      * @param configuration
      *     An instance of {@link Configuration}
-     *
+     * @param isSurefireOrMavenDebug
+     *     A boolean to check surefire or maven's debug is enabled
      * @return An instance of {@link ConfiguredSmartTesting} class
      */
-    public static ConfiguredSmartTesting with(TestVerifier testVerifier, Configuration configuration) {
-        return new ConfiguredSmartTestingImpl(testVerifier, configuration);
+    public static ConfiguredSmartTesting with(TestVerifier testVerifier, Configuration configuration, boolean isSurefireOrMavenDebug) {
```
MatousJobanek: The API should be neither surefire nor maven specific. So just in general if debug is enabled

MatousJobanek: BTW, with the new Logger https://github.com/arquillian/smart-testing/pull/152 it should be possible to implement kind of `getThreshold` method that would return the logger level, so this parameter wouldn't be necessary. After all, we behave based on the logger level, so we should retrieve the information from the Logger.

MatousJobanek: So just wait till the Logger PR is merged - I believe that it should be tomorrow ;-)

PR #211, opened 2017-10-05
Enable configuring skipTests property from pom.xml

### LINKED ISSUE #142: Should not register extension if skipTests is configured in pom.xml
**pom.xml**

Properties:
```xml
<properties>
        <skipTests>true</skipTests>
</properties>
```
Surefire plugin configuration:
```xml
<plugin>
    <configuration>
      <skipTests>${skipTests}</skipTests>  
    </configuration>
</plugin>
```

With this configuration If user is trying to run tests with smart testing enabled. Then `ST` extension should not register.

Currently we are only looking for system properties `maven.test.skip` or `skipTests`. We should look for pom.xml configuration

hemanik:
Should disable registering smart testing extension when `skipTests` property is configured in pom.xml. 

`skipTests` property can be configured inside pom in multiple ways as follows: 

- In order to have it enabled by default, set it inside the properties tag.
``` java
<properties>
        <skipTests>true</skipTests>
</properties>
```
It can similarly be set for `maven.test.skip` and `skipITs`.  The properties set like this are configurable and can be changed at runtime using system properties.
Also, note setting `skipTests` or `maven.test.skip` skips execution of all tests.

- In order to set it for a particular project, set it as a configuration property of the plugin.
``` java
<project>
  [...]
  <build>
    <plugins>
      <plugin>
        <version>2.20.1</version>
        <configuration>
          <skipTests>true</skipTests>
        </configuration>
      </plugin>
    </plugins>
  </build>
  [...]
</project>
```
Different key-words that can be used to skip test execution are: `skipTests`, `skip` and `skipITs`.

For the scenario with both `surefire` and `failsafe` plugins configured the execution is described in the table below as verified against [project](https://github.com/pkainulainen/maven-examples/tree/master/integration-testing).
 
Surefire | Failsafe |  Tests executed in Verify Goal
------------ | ------------- | ------------- 
x | x | UT + IT
skip | x | IT 
x | skip | UT
skip | skip | No tests executed 

From the above observation it can be concluded that `failsafe` handles only execution of the integration tests. Hence, specifying either `skipTests`, `skip` or `skipITs` inside `failsafe` plugin all have the effect of skipping just the integration tests. 

If no integration tests are present and `surefire` is skipped, no tests are executed.

- Read skipTests property from pom.xml and disable extension if set.
- Adds new functional tests with related tags in Dogfooding repo.
- Read property in the test bed to skip test execution from embedded maven build.

NOTE:  The related description of tags set is present [here](https://github.com/arquillian/smart-testing-dogfood-repo/pull/4).

Fixes #142

### DIFF
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
@@ -3,13 +3,17 @@ package org.arquillian.smart.testing.ftest.testbed.project;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
+import java.io.FileReader;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.net.ServerSocket;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Properties;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.maven.model.Model;
@@ -29,6 +33,7 @@ public class BuildConfigurator {
     private static final int DEFAULT_DEBUG_PORT = 8000;
     private static final int DEFAULT_SUREFIRE_DEBUG_PORT = 5005;
 
+    private final Path root;
     private final ProjectBuilder projectBuilder;
     private final Map<String, String> systemProperties = new HashMap<>();
     private final Set<String> modulesToBeBuilt = new HashSet<>();
@@ -47,9 +52,10 @@ public class BuildConfigurator {
     private String mavenVersion;
     private Using usingInstallation;
 
-    BuildConfigurator(ProjectBuilder projectBuilder) {
+    BuildConfigurator(ProjectBuilder projectBuilder, Path root) {
         systemProperties.put("surefire.exitTimeout", "-1"); // see http://bit.ly/2vARQ5p
         systemProperties.put("surefire.timeout", "0"); // see http://bit.ly/2u7xCAH
+        this.root = root;
         this.projectBuilder = projectBuilder;
     }
 
@@ -243,7 +249,7 @@ public class BuildConfigurator {
     }
 
     boolean isSkipTestsEnabled() {
-        return skipTests;
+        return skipTests || Boolean.valueOf(getPropertiesFromPom().getProperty("skipTests"));
     }
 
     String getMavenOpts() {
@@ -280,6 +286,16 @@ public class BuildConfigurator {
 
     boolean isIgnoreBuildFailure() {
         return ignoreBuildFailure;
+    }
+
+    private Properties getPropertiesFromPom() {
+        final File effectivePom = Paths.get(root.toAbsolutePath().toString() + "/pom.xml").toFile();
+        try (FileReader reader = new FileReader(effectivePom)) {
+            Model model = new MavenXpp3Reader().read(reader);
+            return model.getProperties();
+        } catch(Exception ex){
+            throw new RuntimeException("Failed to read properties from file " + effectivePom, ex);
+        }
     }
 
     String getMavenVersion() {
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
@@ -9,17 +9,13 @@ import java.net.ServerSocket;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.maven.model.Model;
-import org.apache.maven.model.Plugin;
 import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
 import org.apache.maven.model.io.xpp3.MavenXpp3Writer;
-import org.codehaus.plexus.util.xml.Xpp3Dom;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
-import org.jboss.shrinkwrap.resolver.impl.maven.embedded.BuiltProjectImpl;
 
 import static java.lang.System.getProperty;
 import static java.util.Arrays.stream;
@@ -247,7 +243,7 @@ public class BuildConfigurator {
     }
 
     boolean isSkipTestsEnabled() {
-        return skipTests || isSkipTestsSetInPom();
+        return skipTests;
     }
 
     String getMavenOpts() {
@@ -256,34 +252,6 @@ public class BuildConfigurator {
 
     File getWorkingDirectory() {
         return workingDirectory;
-    }
-
-    private Boolean isSkipTestsSetInPom() {
-        Plugin surefirePlugin = getPlugin("maven-surefire-plugin");
-        Plugin failsafePlugin = getPlugin("maven-failsafe-plugin");
-
-        if (surefirePlugin != null && failsafePlugin == null) {
-            Xpp3Dom configuration = (Xpp3Dom) surefirePlugin.getConfiguration();
-            return isPropertySetInPluginConfiguration(configuration, "skipTests")
-                || isPropertySetInPluginConfiguration(configuration, "skip");
-        }
-        return isPropertyInPom("maven.test.skip") || isPropertyInPom("skipTests");
-    }
-
-    private boolean isPropertySetInPluginConfiguration(Xpp3Dom pluginConfiguration, String property) {
-        Xpp3Dom propertyKey = pluginConfiguration.getChild(property);
-        if (propertyKey == null) {
-            if (property.equals("skip")) {
-                return isPropertyInPom("maven.test.skip");
-            }
-            return isPropertyInPom(property);
-        }
-        return Boolean.valueOf(propertyKey.getValue());
-    }
-
-    private boolean isPropertyInPom(String property) {
-        String key = getBuiltProject().getModel().getProperties().getProperty(property);
-        return Boolean.valueOf(key);
     }
 
     private boolean isRemoteDebugEnabled() {
@@ -312,21 +280,6 @@ public class BuildConfigurator {
 
     boolean isIgnoreBuildFailure() {
         return ignoreBuildFailure;
-    }
-
-    private Plugin getPlugin(String artifactId) {
-        List<Plugin> pluginList = getBuiltProject().getModel().getBuild().getPlugins()
-            .stream()
-            .filter(plugin -> plugin.getArtifactId().contains(artifactId))
-            .collect(Collectors.toList());
-        if (!pluginList.isEmpty()) {
-            return pluginList.get(0);
-        }
-        return null;
-    }
-
-    private BuiltProjectImpl getBuiltProject() {
-        return new BuiltProjectImpl(projectBuilder.getRoot().toAbsolutePath().toString() + File.separator + "pom.xml");
     }
 
     String getMavenVersion() {
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
@@ -3,12 +3,10 @@ package org.arquillian.smart.testing.ftest.testbed.project;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
-import java.io.FileReader;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.net.ServerSocket;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -20,6 +18,8 @@ import org.apache.maven.model.Model;
 import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
 import org.apache.maven.model.io.xpp3.MavenXpp3Writer;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
+import org.codehaus.plexus.util.xml.Xpp3Dom;
+import org.jboss.shrinkwrap.resolver.impl.maven.embedded.BuiltProjectImpl;
 
 import static java.lang.System.getProperty;
 import static java.util.Arrays.stream;
@@ -249,7 +249,16 @@ public class BuildConfigurator {
     }
 
     boolean isSkipTestsEnabled() {
-        return skipTests || Boolean.valueOf(getPropertiesFromPom().getProperty("skipTests"));
+        return skipTests || isSkipTestsSetInPom();
+    }
+
+    private Boolean isSkipTestsSetInPom() {
+        return Boolean.valueOf(getPropertiesFromPom().getProperty("skipTests")) || isSkipTestSetInPluginConfiguration();
+    }
+
+    private Boolean isSkipTestSetInPluginConfiguration() {
+        Xpp3Dom skipTests = getSurefirePluginConfiguration().getChild("skipTests");
+        return skipTests != null && "true".equals(skipTests.getValue());
     }
 
     String getMavenOpts() {
@@ -289,13 +298,20 @@ public class BuildConfigurator {
     }
 
     private Properties getPropertiesFromPom() {
-        final File effectivePom = Paths.get(root.toAbsolutePath().toString() + "/pom.xml").toFile();
-        try (FileReader reader = new FileReader(effectivePom)) {
-            Model model = new MavenXpp3Reader().read(reader);
-            return model.getProperties();
-        } catch(Exception ex){
-            throw new RuntimeException("Failed to read properties from file " + effectivePom, ex);
-        }
+        return new BuiltProjectImpl(root.toAbsolutePath().toString() + File.separator + "pom.xml")
+            .getModel()
+            .getProperties();
+    }
+
+    private Xpp3Dom getSurefirePluginConfiguration() {
+        return new BuiltProjectImpl(root.toAbsolutePath().toString() + File.separator + "pom.xml")
+            .getModel()
+            .getBuild()
+            .getPlugins()
+            .stream()
+            .filter(plugin -> plugin.getArtifactId().contains("maven-surefire-plugin"))
+            .map(plugin -> (Xpp3Dom) plugin.getConfiguration())
+            .collect(Collectors.toList()).get(0);
     }
 
     String getMavenVersion() {
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
@@ -9,15 +9,16 @@ import java.net.ServerSocket;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
-import java.util.Properties;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.maven.model.Model;
+import org.apache.maven.model.Plugin;
 import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
 import org.apache.maven.model.io.xpp3.MavenXpp3Writer;
-import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
 import org.codehaus.plexus.util.xml.Xpp3Dom;
+import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
 import org.jboss.shrinkwrap.resolver.impl.maven.embedded.BuiltProjectImpl;
 
 import static java.lang.System.getProperty;
@@ -249,27 +250,40 @@ public class BuildConfigurator {
         return skipTests || isSkipTestsSetInPom();
     }
 
-    private Boolean isSkipTestsSetInPom() {
-        return Boolean.valueOf(getPropertiesFromPom().getProperty("skipTests"))
-            || isSkipTestPropertyConfigured() || isSkipPropertyConfigured();
-    }
-
-    private Boolean isSkipTestPropertyConfigured() {
-        Xpp3Dom skipTests = getSurefirePluginConfiguration().getChild("skipTests");
-        return skipTests != null && "true".equals(skipTests.getValue());
-    }
-
-    private Boolean isSkipPropertyConfigured() {
-        Xpp3Dom skip = getSurefirePluginConfiguration().getChild("skip");
-        return skip != null && "true".equals(skip.getValue());
-    }
-
     String getMavenOpts() {
         return mavenOpts;
     }
 
     File getWorkingDirectory() {
         return workingDirectory;
+    }
+
+    private Boolean isSkipTestsSetInPom() {
+        Plugin surefirePlugin = getPlugin("maven-surefire-plugin");
+        Plugin failsafePlugin = getPlugin("maven-failsafe-plugin");
+
+        if (surefirePlugin != null && failsafePlugin == null) {
+            Xpp3Dom configuration = (Xpp3Dom) surefirePlugin.getConfiguration();
+            return isPropertySetInPluginConfiguration(configuration, "skipTests")
+                || isPropertySetInPluginConfiguration(configuration, "skip");
+        }
+        return isPropertyInPom("maven.test.skip") || isPropertyInPom("skipTests");
+    }
+
+    private boolean isPropertySetInPluginConfiguration(Xpp3Dom pluginConfiguration, String property) {
+        Xpp3Dom propertyKey = pluginConfiguration.getChild(property);
+        if (propertyKey == null) {
+            if (property.equals("skip")) {
+                return isPropertyInPom("maven.test.skip");
+            }
+            return isPropertyInPom(property);
+        }
+        return Boolean.valueOf(propertyKey.getValue());
+    }
+
+    private boolean isPropertyInPom(String property) {
+        String key = getBuiltProject().getModel().getProperties().getProperty(property);
+        return Boolean.valueOf(key);
     }
 
     private boolean isRemoteDebugEnabled() {
@@ -300,19 +314,19 @@ public class BuildConfigurator {
         return ignoreBuildFailure;
     }
 
-    private Properties getPropertiesFromPom() {
-        return new BuiltProjectImpl(projectBuilder.getRoot().toAbsolutePath().toString() + File.separator + "pom.xml")
-            .getModel()
-            .getProperties();
+    private Plugin getPlugin(String artifactId) {
+        List<Plugin> pluginList = getBuiltProject().getModel().getBuild().getPlugins()
+            .stream()
+            .filter(plugin -> plugin.getArtifactId().contains(artifactId))
+            .collect(Collectors.toList());
+        if (!pluginList.isEmpty()) {
+            return pluginList.get(0);
+        }
+        return null;
     }
 
-    private Xpp3Dom getSurefirePluginConfiguration() {
-        return new BuiltProjectImpl(projectBuilder.getRoot().toAbsolutePath().toString() + File.separator + "pom.xml")
-            .getModel().getBuild().getPlugins()
-            .stream()
-            .filter(plugin -> plugin.getArtifactId().contains("maven-surefire-plugin"))
-            .map(plugin -> (Xpp3Dom) plugin.getConfiguration())
-            .collect(Collectors.toList()).get(0);
+    private BuiltProjectImpl getBuiltProject() {
+        return new BuiltProjectImpl(projectBuilder.getRoot().toAbsolutePath().toString() + File.separator + "pom.xml");
     }
 
     String getMavenVersion() {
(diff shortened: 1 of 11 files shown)


mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -130,15 +130,22 @@ private void configureExtension(MavenSession session, Configuration configuratio
         final MavenProjectConfigurator mavenProjectConfigurator = new MavenProjectConfigurator(configuration);
         final File dumpedConfigFile = configuration.dump(Paths.get("").toFile());
         session.getAllProjects().forEach(mavenProject -> {
-            mavenProjectConfigurator.configureTestRunner(mavenProject.getModel());
-            copyConfigurationFile(mavenProject.getModel(), dumpedConfigFile);
-            if (isFailedStrategyUsed()) {
-                SurefireReportStorage.copySurefireReports(mavenProject.getModel());
+            String skipTests = mavenProject.getProperties().getProperty("skipTests");
```
dipak-pawar: How are we covering following scenario with above code?

```xml
<plugin>
    <configuration>
      <skipTests>true</skipTests>  
    </configuration>
</plugin>
```

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/testbed/TestRepository.java
```diff
@@ -3,6 +3,6 @@
 public class TestRepository {
 
     public static String testRepository() {
-        return System.getProperty("test.bed.repo", "https://github.com/arquillian/smart-testing-dogfood-repo.git");
+        return System.getProperty("test.bed.repo", "https://github.com/hemanik/smart-testing-dogfood-repo.git");
```
dipak-pawar: I think you forget to change this

hemanik: No, this is on purpose. Will change it once the review is complete, because the dogfooding repo does not have the latest tag yet.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
```diff
@@ -218,7 +226,7 @@ boolean isQuietMode() {
     }
 
     boolean isSkipTestsEnabled() {
-        return skipTests;
+        return skipTests || Boolean.valueOf(getPropertiesFromPom().getProperty("skipTests"));
```
lordofthejars: It is stupid but can you move this `Boolean.valueOf` sentence to a method called something like `isSkipTestsSetInPom` or something like this. Just to make it simple to read the conditional.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -130,15 +130,22 @@ private void configureExtension(MavenSession session, Configuration configuratio
...
+            if (skipTests != null && skipTests.equals("true")) {
```
lordofthejars: Do `"true".equals(skipTests)` and you can remove the `!=null` condition. So `if ("true".equals(skipTests){}` returns `false` if skipTests is null.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
```diff
@@ -39,9 +46,10 @@
     private File workingDirectory;
     private String mavenOpts = "-Xms512m -Xmx1024m";
 
-    BuildConfigurator(ProjectBuilder projectBuilder) {
+    BuildConfigurator(ProjectBuilder projectBuilder, Path root) {
```
MatousJobanek: the `ProjectBuilder` contains the information about root directory, so it should be possible to get it using getter method.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
```diff
@@ -256,7 +264,17 @@ int getSurefireDebugPort() {
...
-    
+
+    private Properties getPropertiesFromPom() {
+        final File effectivePom = Paths.get(root.toAbsolutePath().toString() + "/pom.xml").toFile();
+        try (FileReader reader = new FileReader(effectivePom)) {
+            Model model = new MavenXpp3Reader().read(reader);
+            return model.getProperties();
+        } catch(Exception ex){
+            throw new RuntimeException("Failed to read properties from file " + effectivePom, ex);
+        }
```
MatousJobanek: this code should be enough for getting properties from pom file:
```java
return new BuiltProjectImpl(root.toAbsolutePath().toString() + File.separator + "pom.xml")
    .getModel()
    .getProperties();
```

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/testbed/TestRepository.java
MatousJobanek: the code contained inside of the tag is [here](https://github.com/hemanik/smart-testing-dogfood-repo/commit/eabc6b93f433e2cd8af0d9f0fbf3054382933ff5):
```
+    <!-- skipTests property -->
+    <skipTests>true</skipTests>			 
   </properties>
 
   <issueManagement>
@@ -79,7 +82,8 @@
     <plugins>
       <plugin>
-        <configuration>
+	<configuration>
+	  <skipTests>${skipTests}</skipTests>	
```
the property shouldn't be inside of the `<properties>` tag, but only in the surefire confiugration

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -8,8 +8,10 @@
 
     private static final Pattern TEST_CLASS_PATTERN = Pattern.compile("[^a-z0-9 ]", Pattern.CASE_INSENSITIVE);
 
+    private static boolean skipTests = false;
+
     static boolean isSkipTestExecutionSet() {
-        return isSkipTests() || isSkip();
+        return isSkipTests() || isSkip() || skipTests;
```
MatousJobanek: having `isSkipTests()` method and `skipTests` variable with `setSkipTests` method that take care of something different than the first one is pretty confusing

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
MatousJobanek: Yeah and don't forget that it is also possible to use just `skip`:
```xml
    <configuration>
      <skip>true</skip>  
    </configuration>
```

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/testbed/TestRepository.java
MatousJobanek: as mentioned here: https://github.com/arquillian/smart-testing/pull/211/files#r142849360

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
MatousJobanek: Yeah and keep in mind that it is also possible to use just `skip`:
```xml
    <configuration>
      <skip>true</skip>  
    </configuration>
```

hemanik:
@MatousJobanek , @lordofthejars, @dipak-pawar: Please review this.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
```diff
@@ -72,4 +78,97 @@ public void should_disable_smart_testing_and_execute_no_tests_when_test_executio
...
         assertThat(actualTestResults.accumulatedPerTestClass()).size().isEqualTo(0);
     }
+
+    @Test
+    public void should_disable_smart_testing_and_execute_no_tests_when_skip_test_is_set_as_default_in_pom() throws Exception {
+        // given
+        final Project project = testBed.getProject();
+
+        project
+            .applyAsCommits("Configures skipTests as default property in pom");
```
MatousJobanek: Taking from the commit: 
```xml
+    <skipTests>true</skipTests>
   </properties>
 
   <issueManagement>
@@ -80,6 +83,7 @@
       <plugin>
         <configuration>
+          <skipTests>${skipTests}</skipTests>
```
if you use it in this way, it will be still the property defined in the plugin. If you want to set it only using properties, then the line `<skipTests>${skipTests}</skipTests>` cannot be present there.
In other words - what you wrote there has the very same effect as this one: https://github.com/hemanik/smart-testing-dogfood-repo/commit/3a5380aa90a39abb6e0f58f0539353f808cf3411

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
```diff
@@ -72,4 +78,97 @@ public void should_disable_smart_testing_and_execute_no_tests_when_test_executio
...
+        assertThat(actualTestResults.accumulatedPerTestClass()).size().isEqualTo(20);
+    }
+
+    @Test
+    public void should_be_able_to_enable_tests_from_property_when_set_as_default_property_in_pom() throws Exception {
+        // given
+        final Project project = testBed.getProject();
+
+        project
+            .applyAsCommits("Configure skipITs as default property in pom");
```
MatousJobanek: same as for the surefire one:
```xml
+    <skipITs>true</skipITs>
   </properties>
 
   <issueManagement>
@@ -87,10 +90,31 @@
             <include>**/*TestCase.java</include>
             <include>**/*TestSuite.java</include>
           </includes>
+          <excludes>
+            <exclude>**/IT*.java</exclude>
+          </excludes>
           <useFile>true</useFile>
         </configuration>
       </plugin>
       <plugin>
+        <artifactId>maven-failsafe-plugin</artifactId>
+        <configuration>
+          <skipITs>${skipITs}</skipITs>
```
You just evaluate the property you have set in the properties part... The property can be also 
`<my.cool.skip>true</my.cool.skip>`
and evaluating it: 
`<skipITs>${my.cool.skip}</skipITs>`
would have the very same effect

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -1,32 +1,94 @@
...
+    private final Model model;
+    private final Plugin surefirePlugin;
+    private final Plugin failsafePlugin;
+
     private static final Pattern TEST_CLASS_PATTERN = Pattern.compile("[^a-z0-9 ]", Pattern.CASE_INSENSITIVE);
 
+    MavenPropertyResolver(Model model) {
+        this.model = model;
+        this.surefirePlugin = getPlugin("maven-surefire-plugin");
+        this.failsafePlugin = getPlugin("maven-failsafe-plugin");
```
MatousJobanek: Don't we have something similar in place already? https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ApplicablePlugins.java

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
```diff
@@ -243,7 +252,7 @@ boolean isQuietMode() {
     }
 
     boolean isSkipTestsEnabled() {
-        return skipTests;
+        return skipTests || isSkipTestsSetInPom();
```
MatousJobanek: relates to https://github.com/arquillian/smart-testing/pull/211/files#r144826045

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -1,32 +1,94 @@
...
+
+    private boolean isPropertySetInPluginConfiguration(Xpp3Dom pluginConfiguration, String property) {
+        Xpp3Dom propertyKey = pluginConfiguration.getChild(property);
+        if (propertyKey == null) {
+            if (property.equals("skip")) {
+                return isPropertyInPom("maven.test.skip");
+            }
+            return isPropertyInPom(property);
+        }
+        return "true".equals(propertyKey.getValue());
```
MatousJobanek: use `Boolean.valueOf`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -138,15 +138,21 @@ private void configureExtension(MavenSession session, Configuration configuratio
         final MavenProjectConfigurator mavenProjectConfigurator = new MavenProjectConfigurator(configuration);
         final File dumpedConfigFile = configuration.dump(Paths.get("").toFile());
         session.getAllProjects().forEach(mavenProject -> {
-            mavenProjectConfigurator.configureTestRunner(mavenProject.getModel());
-            copyConfigurationFile(mavenProject.getModel(), dumpedConfigFile);
-            if (isFailedStrategyUsed()) {
-                SurefireReportStorage.copySurefireReports(mavenProject.getModel());
+            MavenPropertyResolver mavenPropertyResolver = new MavenPropertyResolver(mavenProject.getModel());
+            if (mavenPropertyResolver.isSkipTestsSetInPom()) {
+                logger.info("Smart Testing is disabled. Reason: Test Execution has been skipped in %s module.",
```
MatousJobanek: This line is pretty confusing - the ST is not disabled in general - just for the module.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
hemanik: Updated the tag.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
hemanik: Updated the surefire tag and leaving this one as it is, because it also verifies another case, https://github.com/arquillian/smart-testing/pull/211/files#diff-d7b0ca86a6c38fc16017af7fe32281a1R149, i.e user can reset the property from system property when set in property section of the pom.xml.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
hemanik: For some reason, this doesn't work and we need to explicitly set the skip test property as done above.  

The issue is here, if I remove skipTest configuration from here, https://github.com/hemanik/smart-testing/blob/ec98c148c6e51d0249315319f1e9fbb68b68d08e/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java#L73, then `skipTests` property works without any changes being made explicitly but `skipITs` doesn't work at all.

If I configure `skipTest` method in the projectBuilder, then I also need to handle the case, when it is set in pom explicitly.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
```diff
@@ -101,6 +97,10 @@ public BuiltProject getBuiltProject() {
         return builtProject;
     }
 
+    Path getRoot() {
+        return root;
+    }
+
```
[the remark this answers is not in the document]
hemanik: That isn't working. Explanation in the related comment.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
hemanik: That is what I actually wanted it to convey. My thought on this was, if we set the skip test in the pom for a particular module for instance - `core/impl-base` only, then ST should not be disabled for the whole multi-module project in general but only for the particular module for which the tests are skipped.  
WDYT ?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
MatousJobanek: Then change the property name to contain something else, otherwise, you cannot verify it correctly.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
MatousJobanek: > if we set the skip test in the pom for a particular module for instance - core/impl-base only, then ST should not be disabled for the whole multi-module project in general but only for the particular module for which the tests are skipped.

Yes, this is exactly what I would expect. In addition, if the `skipTests` is set in the parent pom file, then it can be still overridden in any of the submodules, so it has to be checked for all modules separately... Also, the log message should correspond to what it is actually doing - just disabling ST for the particular module.
As for the log - on this [line](https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java#L35) we log the information that the ST is activated for the module - there we have the information which modules ST is activated for. So here we can log it at `DEBUG` log level, that it is disabled. 
WDYT?

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
MatousJobanek: > but `skipITs` doesn't work at all.

What do you mean by that? I removed it, then executed [this test](https://github.com/arquillian/smart-testing/pull/211/files#diff-d7b0ca86a6c38fc16017af7fe32281a1R127) and it was green...

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
hemanik: I get your point to rename property name to something like `<my.cool.skip>true</my.cool.skip>` but I wanted to stick with the official documentation, so I was trying to test the case `Skipping By Default` as per [this](http://maven.apache.org/surefire/maven-failsafe-plugin/examples/skipping-tests.html).
Having the property set as `<my.cool.skip>true</my.cool.skip>` in my opinion was almost similar to having it set inside the plugin configuration, so I excluded this case.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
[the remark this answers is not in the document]
hemanik: Updated as discussed, removed conflicting test as it is as per design.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
hemanik: Updated.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -1,32 +1,97 @@
...
+    }
+
+    static boolean isSpecificTestClassSet() {
+        String testClasses = System.getProperty("test");
+        return testClasses != null && !containsPattern(testClasses);
     }
 
-    static boolean isSkipITs() {
-        return Boolean.valueOf(System.getProperty("skipITs", "false"));
+    static private boolean isSkipTests() {
```
dipak-pawar: It'll be good to follow java specification for [this].(https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.1.1)
`private static boolean isSkipTests()`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -1,32 +1,97 @@
...
+        return Boolean.valueOf(System.getProperty("skipTests", "false"));
```
dipak-pawar: no need to explicitly mention "false". We can make it something like `Boolean.valueOf(System.getProperty("skipTests"));`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -1,32 +1,97 @@
...
+        List<Plugin> pluginList = model.getBuild()
+            .getPlugins()
+            .stream()
+            .filter(plugin -> plugin.getArtifactId().equals(applicablePlugin.getArtifactId()))
+            .collect(Collectors.toList());
+        if (!pluginList.isEmpty()) {
+            return pluginList.get(0);
+        }
+        return null;
     }
```
dipak-pawar: How about 
```java
return model.getBuild()
            .getPlugins()
            .stream()
            .filter(plugin -> plugin.getArtifactId().equals(applicablePlugin.getArtifactId()))
            .findFirst()
            .orElse(null);
```

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -1,32 +1,97 @@
...
-    static private boolean isSkipTests() {
-        return Boolean.valueOf(System.getProperty("skipTests", "false"));
+    boolean isSkipITs() {
+        return Boolean.valueOf(System.getProperty("skipITs", "false")) || isSkipITsSetInPom();
+    }
+
+    boolean isSkipTestsSetInPom() {
+        if (surefirePlugin != null && failsafePlugin == null) {
+            Xpp3Dom surefirePluginConfiguration = (Xpp3Dom) surefirePlugin.getConfiguration();
+            return isPropertySetInPluginConfiguration(surefirePluginConfiguration, "skip") ||
```
dipak-pawar: we'll get `nullpointer` if `surefire` doesn't have any configuration. We should check for `surefirePluginConfiguration != null`

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
```diff
@@ -66,6 +69,9 @@ public void should_disable_smart_testing_and_execute_no_tests_when_test_executio
                 .inMode(ORDERING)
             .enable();
 
+        project
+            .applyAsLocalChanges("Enable both surefire and failsafe plugin with sample integration test");
+
```
MatousJobanek: to verify it correctly, there should be the "verify" goal used in the `run()` command - to execute also failsafe tests

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/SkipTestExecutionFunctionalTest.java
```diff
@@ -101,4 +107,103 @@ public void should_not_install_extension_when_only_clean_goal_is_used() throws E
...
+        final TestResults actualTestResults = project.build(CORE_MODULES).run("clean", "verify");
+
+        // then
+        String capturedMavenLog = project.getMavenLog();
+        softly.assertThat(capturedMavenLog).contains(SMART_TESTING_EXTENSION_ENABLED);
+        softly.assertThat(actualTestResults.accumulatedPerTestClass()).size().isEqualTo(20);
+    }
+
+    @Test
+    public void should_be_able_to_enable_tests_from_property_when_set_as_default_property_in_pom() throws Exception {
```
MatousJobanek: I don't think that this test is really needed - you just verify the behavior of Maven - creating the effective pom files.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
```diff
@@ -56,11 +56,8 @@ public void should_only_execute_new_tests_related_to_single_local_change_using_f
                 .createConfigFile()
             .enable();
 
-        project
-            .applyAsCommits("Disable surefire and enable just failsafe plugin");
-
         final Collection<TestResult> expectedTestResults = project
-            .applyAsLocalChanges("Adds new unit test");
+            .applyAsLocalChanges("Skip surefire and enable just failsafe plugin with sample integration test");
```
MatousJobanek: both this change and the `LocalChangesNewTestsSelectionExecutionFunctionalTest.java` one is not necessary - it did the same thing, you just bring another tag for it doin't the same in one commit. 
I would keep it as it was

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -88,7 +87,8 @@ void configureTestRunner(Model model) {
 
         return testRunnerPluginConfigurations.stream()
             .filter(
-                testRunnerPlugin -> !(testRunnerPlugin.getArtifactId().equals("maven-failsafe-plugin") && isSkipITs()))
+                testRunnerPlugin -> !(testRunnerPlugin.getArtifactId().equals("maven-failsafe-plugin")
+                    && new MavenPropertyResolver(model).isSkipITs()))
```
MatousJobanek: Instead of using the string `"maven-failsafe-plugin"` you could use `ApplicablePlugins.FAILSAFE.hasSameArtifactId(testRunnerPlugin.getArtifactId())`

In additon, it took me some time to understand what the if 
`!(testRunnerPlugin.getArtifactId().equals("maven-failsafe-plugin")
&& new MavenPropertyResolver(model).isSkipITs())`
represents. move either the if or the whole filter stream to a seperated method. eg:
`removeFailsafePluginIfSkipped(List<Plugin> testRunnerPluginConfigurations, Model model)`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenPropertyResolver.java
```diff
@@ -3,32 +3,101 @@
...
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import org.apache.maven.model.Model;
+import org.apache.maven.model.Plugin;
+import org.codehaus.plexus.util.xml.Xpp3Dom;
+
+import static org.arquillian.smart.testing.mvn.ext.ApplicablePlugins.FAILSAFE;
+import static org.arquillian.smart.testing.mvn.ext.ApplicablePlugins.SUREFIRE;
 
 class MavenPropertyResolver {
```
MatousJobanek: Having this whole change of the class `MavenPropertyResolver`, then it would make sense to move all the static methods to the `SkipInstallationChecker` and make then non-static. Then you would have this class separated only for methods operating with the `Model` and plugins. Also the name of the class could change then, for example to `SkipModuleChecker`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -139,19 +139,26 @@ private void calculateChanges(File projectDirectory, Configuration configuration
...
-            copyConfigurationFile(mavenProject.getModel(), dumpedConfigFile);
-            if (isFailedStrategyUsed()) {
-                SurefireReportStorage.copySurefireReports(mavenProject.getModel());
+            MavenPropertyResolver mavenPropertyResolver = new MavenPropertyResolver(mavenProject.getModel());
+            if (mavenPropertyResolver.isSkipTestsSetInPom()) {
+                logger.debug("Disabling Smart Testing %s in %s module. Reason: Test Execution has been skipped.",
+                    ExtensionVersion.version().toString(), mavenProject.getArtifactId());
+
+            } else {
+                logger.info("Enabling extension.");
```
MatousJobanek: This will be printed for every module, which is not expected. Bring it back as the extension is enabled anyway, then it just goes through the modules and enables/disables for every module separately

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
MatousJobanek: Btw, in the case when the failsafe plugin is removed, then [this if](https://github.com/arquillian/smart-testing/pull/211/files#diff-5a46d71d79f00752a476439e3186179eR38) will return `false`, but the message about disabling ST for the module is not printed.

MatousJobanek: And yeah, shouldn't we do the same thing also for the surefire plugin? otherwise it will configure the ST for surefire no matter what is set there - in the case when both surefire and failsafe plugins are specified and both of them are skipped.

PR #172, opened 2017-09-20
Suggests closest matching strategy when mistyped

bartoszmajsak:
#### Short description of what this resolves:

With #171 we introduced an exception informing about not existing strategy. This PR brings a suggestion which one could have been the intended one by still failing the build early, but hinting the user with

```
[ERROR] [Smart Testing Extension] Unable to find strategy [chaenged]. Did you mean [changed]?                                                                                                                                                                                              
[ERROR] [Smart Testing Extension] Unable to find strategy [neew]. Did you mean [new]?                                                                                                                                                                                                      
[ERROR] Internal error: java.lang.IllegalStateException: Unknown strategies (see above). Please refer to http://arquillian.org/smart-testing/#
_strategies for the list of available strategies. -> [Help 1]                                                                                 
org.apache.maven.InternalErrorException: Internal error: java.lang.IllegalStateException: Unknown strategies (see above). Please refer to http
://arquillian.org/smart-testing/#_strategies for the list of available strategies.                                                            
        at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:122)                                                                       
        at org.apache.maven.cli.MavenCli.execute(MavenCli.java:993)    
        at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:345)     
        at org.apache.maven.cli.MavenCli.main(MavenCli.java:191)       
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) 
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)                                                      
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)                                              
        at java.lang.reflect.Method.invoke(Method.java:498)            
        at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced(Launcher.java:289)                                                
        at org.codehaus.plexus.classworlds.launcher.Launcher.launch(Launcher.java:229)                                                        
        at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode(Launcher.java:415)                                              
        at org.codehaus.plexus.classworlds.launcher.Launcher.main(Launcher.java:356)            
```

Fixes #170

PR #212, opened 2017-10-06
feat(#92): Allows user registers his own strategies

### LINKED ISSUE #92: Configurable list of strategies
Currently, we do not allow registering custom strategies, but we have a mechanism in place which does mapping between a strategy name and corresponding dependency. We should make it configurable through a property pointing to the file, e.g. `smart.testing.custom.strategies`. This should work for both relative and absolute paths.

This task is about passing this configuration to [`StrategyDependencyResolver`](https://github.com/arquillian/smart-testing/blob/0b961dfba1dfc71ecc55be0b74feed6c5a517ea3/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java#L31) so we can open it up for new strategies.

lordofthejars:

Allows user registers his own strategies using `smart.testing.custom.strategies` property

- `StrategyDependencyResolver` to load custom strategies

IMPORTANT: Not to be merged until #172 is merged.

Fixes #92


MatousJobanek:
I don't understand your comment:

> IMPORTANT: Not to be merged until #172 is merged.

that PR had been already merged

lordofthejars:
The PR is the #212 sorry for the confusion.

core/src/test/resources/configuration/smart-testing.yml
```diff
@@ -9,3 +9,4 @@ scm:
     range:
       head: HEAD #<!--7-->
       tail: HEAD~2 #<!--8-->
+customStrategiesDefinition: ./custom.properties #<!--9-->
```
MatousJobanek: Referencing the file when the content is very simple doesn't make sense to me. 
I would rather use: 
```yaml
customStrategiesDefinition:
    - smart.testing.strategy.cool=my.company.smart.testing:strategy-cool:1.0.0
```
The external property file seems to me just too complicated

docs/registering-strategies.adoc
```diff
@@ -0,0 +1,87 @@
...
+Finally you need to create a `.properties` file where you set the strategy name as key and the Maven coordinates as value.
+This file is required so *Smart Testing* can automatically register the artifacts in build tool script.
+
+As an example of registering previous strategy:
+
+[[custom-strategy]]
+[source, properties]
+.custom-smart-testing.properties
+----
+smart.testing.strategy.failed=org.arquillian.smart.testing:strategy-failed:1.0.0
```
MatousJobanek: as a first example I would use some non-existing strategy eg:
`smart.testing.strategy.cool=my.company.smart.testing:strategy-cool:1.0.0`
then I would say that this strategy can be activated using property value `cool`
Then, you could also show how it is done for our existing strategies...
This is just my opinion ;-)

core/src/test/resources/configuration/smart-testing.yml
lordofthejars: you mean the name of the property? Actually I am open to better name.

About what you mention, I think that a file is better because then you are able to share it between different projects, and not having to do copy-paste between projects. i see for example that here you'd be able to put a URL so a company can have stored on git and jsut put there the URL where these definitions are stored.

What we can do is allow both ways for now file and embedded.

docs/registering-strategies.adoc
lordofthejars: Yes the thing is that if I pick an existing strategy then I can include directly the code, so if it changes the interface names, then the documentation is automatically updated.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -103,13 +105,22 @@ public void setScm(Scm scm) {
         this.scm = scm;
     }
 
+    public Path getCustomStrategiesDefinition() {
+        return customStrategiesDefinition != null ? Paths.get(customStrategiesDefinition) : null;
```
MatousJobanek: This is probably not the best way of getting the path - don't forget that in case it is called from other than the main directory, you could get a wrong path. eg:

`mvn clean package -f ./smart-testing-test/pom.xml -Dsmart.testing.custom.strategies=./my-project/strategy.properties`

lordofthejars: But then we start the discussion of is `.` referring to project root or `pom.xml` root?

MatousJobanek: Project root is the same as `pom.xml` root, isn't it? 
You can call the project from a different directory using the parameter `-f` (eg. from your home directory)

core/src/test/resources/configuration/smart-testing.yml
MatousJobanek: Nope, I didn't mean the property name. But yeah, if we had it only embedded, then I would use only `customStrategies`.
I would still vote for having it inside of the file and keep the config consistent. 

> I think that a file is better because then you are able to share it between different projects, and not having to do copy-paste between projects.

You could say the same thing about `pom.xml` dependencies or other stuff. I'm not sure if it is really useful. I would keep it simple and if there is a request for such a feature then yeah, we can implement it.

lordofthejars: so just embedded right?

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
lordofthejars: Ok then you mean it is better to use `baseDir` as well?

MatousJobanek: `baseDir` wouldn't help you.
Anyway, I would keep it only embedded https://github.com/arquillian/smart-testing/pull/212#discussion_r143196654 so you don't have to care about this path thingy

lordofthejars: True 😆

core/src/test/resources/configuration/smart-testing.yml
dipak-pawar: +1 for Embedded. 
>
> ```yaml 
> customStrategiesDefinition:
>   - smart.testing.strategy.cool=my.company.smart.testing:strategy-cool:1.0.0
> ```

@MatousJobanek Do we really need to define complete `smart.testing.strategy.cool`? 

How about 
``` yaml
customStrategies:
- cool=my.company.smart.testing:strategy-cool:1.0.0
```

MatousJobanek: Well, the short version looks much better! The only thing I'm afraid of is that we are not consistent with the parameter names in `strategies.properties` file and here in config YAML file. 
@lordofthejars WDYT about it? would that be hard to support both cases?

lordofthejars: For this version I won't touch this yet. I would advocate for talk internally and do change properly in all places if we agree

core/src/test/java/org/arquillian/smart/testing/configuration/ObjectMapperTest.java
```diff
@@ -187,6 +190,7 @@ public void clearMap() {
         private List<String> l;
         private Map<String, String> m;
         private TestEnum e;
+        private String[] multiple;
```
MatousJobanek: there have been added a new variable and imports without using it anywhere

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
```diff
@@ -29,23 +30,44 @@
...
+        this.propertiesPath = null;
+        this.customStrategies = customStrategies;
     }
 
     Map<String, Dependency> resolveDependencies() {
         final Properties properties = new Properties();
         properties.putAll(loadDefaultMapping());
+        properties.putAll(loadCustomStrategies());
         properties.putAll(loadFromFile());
         properties.putAll(System.getProperties());
```
MatousJobanek: Why do we have this line here? I thought that taking care of getting the values from system property should be a responsibility of the `ObjectMapper` class.

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationItem.java
```diff
@@ -1,10 +1,13 @@
 package org.arquillian.smart.testing.configuration;
 
+import java.util.regex.Pattern;
+
 public class ConfigurationItem {
 
     private final String paramName;
     private String systemProperty;
     private Object defaultValue;
+    private Pattern pattern;
```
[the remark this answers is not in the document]
MatousJobanek: btw, instead of using `Pattern` I would just expect that if system property ends with `.*` then it should go through all system properties and gets everything that starts with the string from it and store it as a List

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolverTest.java
```diff
@@ -76,6 +76,26 @@ public void should_overwrite_default_versions_when_property_file_is_used() throw
...
+        // when
+        Map<String, Dependency> dependencies = strategyDependencyResolver.resolveDependencies();
+
+        // then
+        assertThat(dependencies.values()).hasSize(5)
+            .extracting(
+                dependency -> dependency.getGroupId() + ":" + dependency.getArtifactId() + ":" + dependency.getVersion())
+            .contains("org.arquillian.smart.testing:strategy-changed:" + ExtensionVersion.version().toString(),
+                "org.arquillian.smart.testing:strategy-failed:" + ExtensionVersion.version().toString(),
+                "org.arquillian.smart.testing:strategy-changed:" + ExtensionVersion.version().toString(),
```
MatousJobanek: the `strategy-changed` is there twice

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
MatousJobanek: btw. having this opportunity of setting custom strategies in place opens a possibility of creating more structured names of the strategies. eg: `smart.testing.strategy.my.cool=org.company:my-cool:version` then this line won't work: https://github.com/arquillian/smart-testing/pull/212/files#diff-22d18b722b6b02c374c2c2b102ed5b67R103

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -30,6 +31,8 @@
 
     public static final String SMART_TESTING = "smart.testing";
     public static final String SMART_TESTING_MODE = "smart.testing.mode";
+    public static final String SMART_TESTING_CUSTOM_STRATEGIES = "smart.testing.strategy";
+    public static final String SMART_TESTING_CUSTOM_STRATEGIES_PATTERN = SMART_TESTING_CUSTOM_STRATEGIES.replace(".", "\\.") + ".*";
```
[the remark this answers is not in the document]
lordofthejars: It seems so, but in different modules, so it would require some refactor on both sides, I think it is easy

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationItem.java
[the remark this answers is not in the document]
lordofthejars: It is not necessary, it was used before but after code review I delete all code associated but I left erroneously the field

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
lordofthejars: nd they are loaded by ObjectMapper, notice that here I am not reading anything from any system property

lordofthejars: The problem is that then `my.cool` is going to be the name of the strategy. Then for now what it means that when you want to enable it you would need to do something like `mvn .... -Dsmart.testing="affected, my.cool"` which probably it has sense, but just wanted to notice it. I am going to write a test and fix problems.

lordofthejars:
> Why do we have this line here? I thought that taking care of getting the 

Ok now I find what you mean, this code was originally there before my PR, I think that at that time since we didn't have the configuration file yet, it was decided to put it there, but as you mention now maybe it could be moved to `ObjectMapper`, but since @bartoszmajsak added this line I prefer to listen his point of view too, because maybe he had a good reason to put it there.

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationItem.java
```diff
@@ -57,4 +57,5 @@ public String getSystemProperty() {
     public Object getDefaultValue() {
         return defaultValue;
     }
+
```
MatousJobanek: You like white spaces as well :-D https://github.com/arquillian/smart-testing/pull/233#discussion_r144572746

core/src/test/java/org/arquillian/smart/testing/configuration/ObjectMapperTest.java
```diff
@@ -7,11 +7,14 @@
...
 import java.util.Map;
 import java.util.function.Function;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.contrib.java.lang.system.RestoreSystemProperties;
```
MatousJobanek: still unnecessary imports?

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -122,6 +134,7 @@ public void setAutocorrect(boolean autocorrect) {
         configItems.add(new ConfigurationItem("disable", SMART_TESTING_DISABLE, false));
         configItems.add(new ConfigurationItem("debug", SMART_TESTING_DEBUG, false));
         configItems.add(new ConfigurationItem("autocorrect", SMART_TESTING_AUTOCORRECT, false));
+        configItems.add(new ConfigurationItem("customStrategies", Pattern.compile(SMART_TESTING_CUSTOM_STRATEGIES_PATTERN)));
```
MatousJobanek: When you removed the pattern from the `ConfigurationItem` then the second parameter doesn't make sense. There is no such a support for getting system properties based on the pattern or provided prefix in the `ObjectMapper` class - that should be implemented there as part of this PR, otherwise, you don't get the values set by system properties there.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
MatousJobanek: I know, but it should still work. I wouldn't wonder if somebody would use it in this way.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
```diff
@@ -93,4 +116,8 @@ private Properties loadFromFile() {
                     return dependency;
                 }));
     }
+
+    private String filterPrefix(String key) {
+        return key.substring(key.lastIndexOf(SMART_TESTING_STRATEGY_PREFIX) + SMART_TESTING_STRATEGY_PREFIX.length());
```
MatousJobanek: why "last index of" - I think that this is for multiple occurrences which is not this case.
In addition, wouldn't be enough to just call `key.substring(SMART_TESTING_STRATEGY_PREFIX.length())`? I haven't tried it locally, but I believe that something like this should work...

MatousJobanek:
as for `properties.putAll(System.getProperties());`: https://github.com/arquillian/smart-testing/pull/212#issuecomment-336878020
I know that it was before your PR, however, because you implemented the new way then it should be changed.
Looking at the following code, I don't think that it has some other purpose than providing additional strategies using system properties. As this is related to the configuration part, then it should be moved there - this relates to: https://github.com/arquillian/smart-testing/pull/212#discussion_r144964961

MatousJobanek:
you probably forgot to rebase it with the latest changes of the master branch: https://github.com/arquillian/smart-testing/pull/212#pullrequestreview-69519508

docs/configfile.adoc
```diff
@@ -60,6 +61,9 @@ a| To run Smart Testing with `SCM` configuration. You can either define `range`
 
 a| autocorrect
 a| This configures Smart Testing to auto correct misspelled strategies to the closest one. For example in case of user set strategies to `nwe`, if `autocorrect` is enabled then it is automatically changed to `new`.
+
+a| customStrategies
+a| A list of custom strategies in the form of key/value. It is important to notice that the _key_ part must be prefixed with `const:mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java[name="SMART_TESTING_STRATEGY_PREFIX"]`.
```
dipak-pawar: Constant `SMART_TESTING_STRATEGY_PREFIX` is showing `.` as a result in final view. Can you generate documentation locally & check it?

docs/refcard.adoc
```diff
@@ -82,6 +82,11 @@ a| `const:strategies/affected/src/main/java/org/arquillian/smart/testing/strateg
 |Set transitivity enabled
 a|const:strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java[name="DEFAULT_SMART_TESTING_AFFECTED_TRANSITIVITY_VALUE"]
 a|`affected`
+
+a| `const:mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java[name="SMART_TESTING_STRATEGY_PREFIX"]`.<strategyName>
```
dipak-pawar: Constant `SMART_TESTING_STRATEGY_PREFIX` is showing `.` as a result in final view. Can you generate documentation locally & check it?

docs/configfile.adoc
lordofthejars: I did it in this way so user see that he still needs to put the dot, so he don't do (I know that it is stupid) but they don't do `smart.testing.strategy;cool=org.....` notice that I am using `;` as separator

docs/refcard.adoc
lordofthejars: Same as before

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationItem.java
```diff
@@ -5,27 +5,32 @@
...
      * @param systemProperty required to overwrite field value.
+     * @param isRegExp true if system property is a regular expression of form my.proeprty.*
      * @param defaultValue to use for the field.
      */
-    public ConfigurationItem(String paramName, String systemProperty, Object defaultValue) {
+    public ConfigurationItem(String paramName, String systemProperty, boolean isRegExp, Object defaultValue) {
         this.paramName = paramName;
         this.systemProperty = systemProperty;
         this.defaultValue = defaultValue;
+        this.regExp = isRegExp;
```
dipak-pawar: Can't we define method which will check for given system property is a refular expression of required form i.e. `my.property.*`? So we'll get rid out of this `isRegExp` params in constructor which is not necessary.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -30,6 +31,8 @@
 
     public static final String SMART_TESTING = "smart.testing";
     public static final String SMART_TESTING_MODE = "smart.testing.mode";
+    public static final String SMART_TESTING_CUSTOM_STRATEGIES = "smart.testing.strategy";
```
dipak-pawar: I know we haven't used `smart.testing.strategy` variable/property previously. how about `smart.testing.custom.strategy` ?

It won't be confusing for user as well.

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationItem.java
lordofthejars: so you mean if it ends with `.*` means it is a regular expression right?

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
lordofthejars: Yes we used previously in previous versions when you could register strategies using system properties.

docs/configfile.adoc
dipak-pawar: screenshot from 2017-10-17 16-29-49

dipak-pawar: Look how it's displaying in resulting documents

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
dipak-pawar: no we are using `smart.testing` for previous versions. We haven't used `smart.testing.strategies`. 
https://github.com/arquillian/smart-testing/pull/212/files/217a5df9249ef1e56eaba84be0c264ee13a49eed#diff-348861c21efd28f5cc99117b2c67302aR32

lordofthejars: It seems because I moved the bits but look at current master https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java#L29

docs/registering-strategies.adoc
```diff
@@ -0,0 +1,80 @@
+== Registering Custom Strategies
```
dipak-pawar: You haven't included `docs/registering-strategies.adoc` in `README.adoc`

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
dipak-pawar: wdyt about `smart.testing.custom.strategies`?

lordofthejars: @bartoszmajsak any opinion about this? In my opinion original attribute name was ok, since it gives "custom" strategies the same level as internal strategies, but of course I am open to change.

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -69,9 +70,23 @@ private static Object getConvertedObject(Method method, Object configFileValue,
             Object mappedValue = null;
             ConfigurationItem configItem = foundConfigItem.get();
 
-            if (configItem.getSystemProperty() != null) {
+            if (configItem.getSystemProperty() != null && !configItem.getSystemProperty().endsWith("*")) {
                 mappedValue = System.getProperty(configItem.getSystemProperty());
             }
+
+            if (configItem.getSystemProperty() != null && configItem.getSystemProperty().endsWith("*")) {
```
dipak-pawar: Can we check `configItem.getSystemProperty().endsWith(".*")`? to avoid getting wrong propertyName using  `configItem.getSystemProperty().substring(0, configItem.getSystemProperty().lastIndexOf('.'));`

Otherwise something will be wrong for this type of input `smart.testing.strategy.custom*`

core/src/test/java/org/arquillian/smart/testing/configuration/ObjectMapperTest.java
```diff
@@ -226,17 +244,22 @@ public void setDummyObject(DummyObject dummyObject) {
...
-            configItems.add(new ConfigurationItem("m", null, Collections.EMPTY_MAP));
-            configItems.add(new ConfigurationItem("as", null, new String[0]));
-            configItems.add(new ConfigurationItem("e", null, TestEnum.FOO));
+            configItems.add(new ConfigurationItem("d", 0.0));
+            configItems.add(new ConfigurationItem("b", false));
+            configItems.add(new ConfigurationItem("l", Collections.EMPTY_LIST));
+            configItems.add(new ConfigurationItem("m", Collections.EMPTY_MAP));
+            configItems.add(new ConfigurationItem("as", new String[0]));
+            configItems.add(new ConfigurationItem("e", TestEnum.FOO));
+            configItems.add(new ConfigurationItem("multiple", "my.property.*"));
```
dipak-pawar: @lordofthejars : I am not sure Why @MatousJobanek Removed constructor ` ConfigurationItem(String paramName, Object defaultValue)`. You can check it [here](https://github.com/arquillian/smart-testing/commit/2f59136f1e8114df1e373e5deab06296412e652e)

Maybe he has some good reason.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
```diff
@@ -26,26 +28,46 @@
  */
 class StrategyDependencyResolver {
 
-    protected static final String SMART_TESTING_STRATEGY_PREFIX = "smart.testing.strategy.";
+    protected static final String SMART_TESTING_STRATEGY_PREFIX = Configuration.SMART_TESTING_CUSTOM_STRATEGIES + ".";
 
     private final Path propertiesPath; // TODO this could be configurable through system property and with this we need a path
+    private final String[] customStrategies;
```
dipak-pawar: How about initializing it with empty array?
something like `private final String[] customStrategies = new String[0];`

We did the same way for strategies when we were using system property

docs/registering-strategies.adoc
bartoszmajsak: Makes me wonder - no one noticed something is not right with the docs preview we automated? Shouldn't we be getting comments from `alien-ike` about docs preview be ready?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
```diff
@@ -26,26 +28,46 @@
...
+        properties.putAll(loadCustomStrategies());
         properties.putAll(loadFromFile());
-        properties.putAll(System.getProperties());
         return transformToDependencies(properties);
     }
 
+    private Properties loadCustomStrategies() {
+        final Properties properties = new Properties();
+
+        if (customStrategies != null) {
```
bartoszmajsak: As @dipak-pawar said - if you initialize it with an empty array instead of `null` this could be slightly nicer to digest.

docs/registering-strategies.adoc
```diff
@@ -0,0 +1,80 @@
...
+To do it you just need to follow next four steps:
+
+* Implement `org.arquillian.smart.testing.spi.TestExecutionPlannerFactory` Java SPI and registering as such.
+* Implement `org.arquillian.smart.testing.spi.TestExecutionPlanner` interface.
+* Register strategy on Smart Testing configuration file (optional).
+
+=== Implementing `TestExecutionPlannerFactory`
+
+`TestExecutionPlannerFactory` is a Java SPI interface that manages the creation of the strategy.
+You need to implement three simple methods, two related with the name of the strategy and one that generates the strategy class.
```
bartoszmajsak: While reading this - isn't a need for implementing two methods for the name somewhat a smell? @lordofthejars as you were looking at ability to bring those custom strategies - do you see any potential improvements so that we can have lean, but powerful mechanism in place?

docs/registering-strategies.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+You can still use another approach to register the pair _strategy name_:_Maven coordinates without configuring them in Smart Testing configuration file and it is using a System property.
+
+For example <<custom-strategy, previous example>> would be equivalent as running *Smart Testing* with next System property:
+
+`-Dsmart.testing.strategy.failed=org.arquillian.smart.testing:strategy-failed:1.0.0`
+
+Notice again that it must be prefixed with `const:mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java[name="SMART_TESTING_STRATEGY_PREFIX"]`.
+
+
```
bartoszmajsak: What I would like to see in this doc is also some hints about optimizations (calculating upfront before the build starts on the extension side). Something we did for Git resolutions as well as for Failed strategy. This is somewhat a hidden gem, but crucial to have a pleasant experience using our tool (with low overhead).

bartoszmajsak:
> IMPORTANT: Not to be merged until #172 is merged.

Pro tip - if we all use zenhub you can mark that this PR is blocked by other (or issue).

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
lordofthejars: I thought the same as you, but for any reason that now I cannot remember I decided to use `*`. I change it because now seems more reasonable what you mention.

core/src/test/java/org/arquillian/smart/testing/configuration/ObjectMapperTest.java
lordofthejars: this is not the default value, but the second parameter is the system property name

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -114,6 +118,14 @@ public void setAutocorrect(boolean autocorrect) {
...
+        this.customStrategies = customStrategies;
+    }
+
+    public String[] getCustomStrategies() {
+        return customStrategies;
+    }
+
     public List<ConfigurationItem> registerConfigurationItems() {
         List<ConfigurationItem> configItems = new ArrayList<>();
         configItems.add(new ConfigurationItem("strategies", SMART_TESTING, new String[0]));
```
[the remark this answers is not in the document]
lordofthejars: Because @MatousJobanek mentioned in one previous review (and I think he was right) that if system properties are read in `Configuration` class, before this PR the system properties of custom strategies were read in https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java#L45 then some sys props were read from `Configuration` and others from `StrategyDependencyResolver`, to put all in same place you need to configure each property that is going to be read in `Configuration` class in `configItems`

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationItem.java
```diff
@@ -36,6 +36,16 @@ public ConfigurationItem(String paramName) {
         this.paramName = paramName;
     }
 
+    /**
+     * Constructor
```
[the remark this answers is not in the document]
lordofthejars: Just to follow all constructors of this class, which contains this line.

core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationTest.java
```diff
@@ -40,6 +40,7 @@ public void should_load_configuration_with_default_values_if_property_is_not_spe
         expectedConfiguration.setScm(scm);
         expectedConfiguration.setReport(report);
         expectedConfiguration.setAutocorrect(true);
+        expectedConfiguration.setCustomStrategies(new String[]{"smart.testing.strategy.cool=org.arquillian.smart.testing:strategy-cool:1.0.0"});
```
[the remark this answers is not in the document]
lordofthejars: I understand your point, for now and to not make this PR longer, I'd advocate for leaving this file for internal usage.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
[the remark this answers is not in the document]
bartoszmajsak: Wouldn't it be great to have only one simple way though?

docs/registering-strategies.adoc
lordofthejars: I thought exactly the same, but since I didn't have the knowledge on why it was in this way, I decided that for this PR leave it as this and discuss in another PR. Notice that one is the real name and the other is the alias.

core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationTest.java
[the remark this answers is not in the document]
bartoszmajsak: I am not sure if "not keeping PR longer" is the right success criteria - if this change does not mean a lot of work shipping better user experience is more important from my point of view than something which might be quite confusing.

docs/registering-strategies.adoc
bartoszmajsak: All knowledge is in the code ;)

core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationTest.java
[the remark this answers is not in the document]
lordofthejars: I think we tend to add more features in PRs that the ones it describes originally. In my opinion a PR should implement what it mentions, if during the process we see that there is an internal design problem then I'd open a new PR. But it is my opinion.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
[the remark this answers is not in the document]
lordofthejars: there is only one way, and this is by registering the attribute you want to read either from configuration file or system property. @dipak-pawar can give you more details, but in essence this class just read a property from file, then check if there is a system property overriding it and if not, if a default value is provided, then this value is the one set in `Configuration` class.

docs/registering-strategies.adoc
lordofthejars: But anyway, the purpose of this PR is not simplify strategies "architecture" but allowing user to register its custom strategies. I think this should go to another PR.

core/src/test/java/org/arquillian/smart/testing/configuration/ObjectMapperTest.java
MatousJobanek: Yeah, I was thinking about the same - maybe style checker could help

lordofthejars: This import is necessary for now, but yeah, even walkmod could fix this automatically.

core/src/test/java/org/arquillian/smart/testing/configuration/ObjectMapperTest.java
MatousJobanek: Yeah, I had good reason for it @dipak-pawar.
Having the constructor: 
`ConfigurationItem(String paramName, Object defaultValue)`
and at the same time also
`ConfigurationItem(String paramName, String systemProperty)`
may lead to confusions and, in some cases, also to a wrong usage of it. In the original stage of this PR, there was exactly the case of the wrong usage of the constructors. 
In addition, I don't see it as a big benefit to have the constructor there, when it was used only by tests and may lead to confusions...

core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationTest.java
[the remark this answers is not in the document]
bartoszmajsak: I don't think it's adding more features, I think it's striving to have a good feature in the first place.

docs/registering-strategies.adoc
bartoszmajsak: I didn't really ask for changes here - I wanted to know if you gave it some thought.

lordofthejars: yes, I did it, and it seems there are some redundant info.

core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationTest.java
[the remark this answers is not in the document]
lordofthejars: well we can drop this file and add as default initial values to strategies the content of this file. Personally since this file is just used internally I'd leave as is, and for now `strategies.properties` is used for internal/default strategies.

lordofthejars:
@MatousJobanek reviewing your change, I think that it should not be appended strategies blindly. Let me explain why:

Suppose you have next configuration file:

```
customStrategies: 
    - smart.testing.strategy.cool=org.arquillian.smart.testing:strategy-cool:1.0.0
```

And then user as system property add next one:

`-Dsmart.testing.strategy.cool=org.arquillian.smart.testing:strategy-cool:1.0.1`

I think that both are registered when only the system property strategy should be registered. Isn't it?

lordofthejars:
@MatousJobanek I have written a test and I confirm that both are registered.

MatousJobanek:
Yeah, because I'm stupid and I forgot to add the "merge" functionality... 
It was easy to fix: https://github.com/arquillian/smart-testing/pull/212/commits/e246e97145fb82e09ee5081a04cbe6d4cbc8e575
good catch, thx!

MatousJobanek:
Based on offline discussion, I'll move the merge logic to `ObjectMapper`

PR #213, opened 2017-10-06
fix(#173): Adds autocorrect for mistyped strategy

### LINKED ISSUE #173: Enable autocorrect for mistyped strategy
##### Issue Overview

With #172  (and earlier #171) we introduced better information of misconfigured strategies (mistyped).

With this task we introduce autocorrect feature (`strategy.autocorrect`), when if enabled will correct mistyped strategy to the closest matching one.

In case when more than one from existing strategies will be the same match for misspelled one we fail the build with the proper error message pointing which ones are having the same match.

lordofthejars:

Adds autocorrect for mistyped strategy if it is enabled by using `smart.testing.autocorrect` property.

- Since `Configuration` is read in two different places (when dependencies are configured) and when strategies are instantiated (Java SPI), some classes has been moved from Maven extension module to Core module.
- `Configuration` strategies array is modified when strategies are autocorrected, so all parts of code that uses strategies just gets the fixed ones and not the wrong ones.
- In case that an autocorrected strategy becomes a duplicated of configured strategy, an exception is thrown.

Fixes #173


core/src/main/java/org/arquillian/smart/testing/impl/StrategyDependencyResolver.java
```diff
@@ -0,0 +1,68 @@
...
+            throw new IllegalStateException("Unable to load default strategy dependencies mapping.", e);
+        }
+        return properties;
+    }
+
+    protected Properties loadFromFile() {
+        final Properties properties = new Properties();
+        if (propertiesPath != null) {
+            try {
+                properties.load(new FileInputStream(this.propertiesPath.toFile()));
```
MatousJobanek: I remember that we had a discussion about this functionality (autocorrection in general). I wanted to call the `StringSimilarityCalculator` on both sides Maven extension (when the dependencies are resolved) and surefire provider (when the strategies are applied) - you told me that for optimization reasons you want to call it only once. 
But now, you read the file twice - when you apply autocorrection and when you resolve dependencies. I'm afraid that reading a file is bigger load than checking similarities. WDYT?

lordofthejars: Yes the problem is that we need to or instantiate twice `StringSimilarityCalculator ` or we create a singleton for it.

MatousJobanek: Looking at your changes of  `StringSimilarityCalculator` https://github.com/arquillian/smart-testing/pull/213/files#diff-61b559b9fdc0af7def1988ba0482defa there is no (and wasn't) Maven `Dependency` class used. You just take the sample (set by user) and list of targets (available strategies) and you are looking for the similarity. This is exactly what we need on both sides - Maven extension when the dependencies are resolved and surefire provider (`TestStrategyApplier`) when we apply available strategies.

core/src/main/java/org/arquillian/smart/testing/impl/TestExecutionPlannerLoaderImpl.java
```diff
@@ -21,14 +22,23 @@
...
         }
 
         if (availableStrategies.containsKey(strategy)) {
             return availableStrategies.get(strategy).create(projectDir, verifier);
+        } else {
+            if (autocorrect) {
+                final StringSimilarityCalculator stringSimilarityCalculator = new StringSimilarityCalculator();
+                final String closestMatch =
+                    stringSimilarityCalculator.findClosestMatch(strategy, availableStrategies.keySet());
+                if (availableStrategies.containsKey(closestMatch)) {
```
MatousJobanek: is this if necessary?

MatousJobanek: btw, is there any benefit of creating the instance of `StringSimilarityCalculator` using the constructor and not using it as static class/method?

MatousJobanek:
> In case that an autocorrected strategy becomes a duplicated of configured strategy, an exception is thrown.

I haven't found any logic for this check

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java
```diff
@@ -30,21 +32,41 @@ private void addStrategies(Model model) {
...
+                    strategyMismatch.add(String.format("Strategy [%s] was already registered or autocorrected", definedStrategy));
+                } else {
+                    final Dependency dependency = dependencies.get(definedStrategy);
+                    model.addDependency(dependency);
+                    registeredStrategies.add(definedStrategy);
+                }
             }
         }
-        strategyMismatch.forEach((selection, match) -> LOGGER.error("Unable to find strategy [%s]. Did you mean [%s]?", selection, match));
+        strategyMismatch.forEach(msg -> LOGGER.error(msg));
```
dipak-pawar: * `Strategy [%s] was already registered or autocorrected`
 * `Autocorrected [%s] strategy to [%s] but it was already registered` 
this msgs belongs to `info` or `error`?

core/src/test/resources/strategies.properties
```diff
@@ -0,0 +1,3 @@
+smart.testing.strategy.new=org.arquillian.smart.testing:strategy-changed:0.0.11-SNAPSHOT
```
dipak-pawar: I don't see this file used in some tests. Is it redundant?

core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationTest.java
```diff
@@ -76,6 +79,7 @@ public void should_load_default_configuration() {
         assertThat(defaultConfiguration).isEqualToComparingFieldByFieldRecursively(expectedConfiguration);
     }
 
+
```
dipak-pawar: extra line

core/src/main/java/org/arquillian/smart/testing/impl/TestExecutionPlannerLoaderImpl.java
```diff
@@ -21,14 +22,23 @@
         this.projectDir = projectDir;
     }
 
-    public TestExecutionPlanner getPlannerForStrategy(String strategy) {
+    public TestExecutionPlanner getPlannerForStrategy(String strategy, boolean autocorrect) {
```
dipak-pawar: We do have `projectDir` in this. I am not sure do we really need to pass it in constructor. because you can load configuration using `Configuration.loadPreCalculated(projectDir)` method.

Is it because you don't want to reload it agai to avoid two times reading of file?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/MisspelledStrategiesFunctionalTest.java
```diff
@@ -24,7 +25,7 @@
     public TestBed testBed = new TestBed(GIT_CLONE);
 
     @Test
-    public void should_fail_and_log_suggestion_for_misspelled_strategies() throws Exception {
+    public void should_fail_and_log_suggestion_for_misspelled_strategies_when_autocorrect_disabled() throws Exception {
```
dipak-pawar: I don't see anything in this test which is throwing exception

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/MisspelledStrategiesFunctionalTest.java
```diff
@@ -50,4 +51,30 @@ public void should_fail_and_log_suggestion_for_misspelled_strategies() throws Ex
         assertThat(projectMavenLog).contains("Unable to find strategy [" + misspelledNew.getName() + "]. Did you mean [new]?");
     }
 
+    @Test
+    public void should_succeed_for_misspelled_strategies_when_autocorrect_enabled() throws Exception {
```
dipak-pawar: same as above

core/src/main/java/org/arquillian/smart/testing/impl/TestExecutionPlannerLoaderImpl.java
lordofthejars: Well it was initially implemented as this

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java
lordofthejars: It is an error.

core/src/test/resources/strategies.properties
lordofthejars: It was removed and moved to Maven extension,

core/src/main/java/org/arquillian/smart/testing/impl/TestExecutionPlannerLoaderImpl.java
lordofthejars: About condition, well I think it is a safe condition no avoid any NPE

lordofthejars:
@MatousJobanek 
> I haven't found any logic for this check

https://github.com/arquillian/smart-testing/pull/213/files#diff-e5427bc921ef9faf2177273978bd12feR42 and https://github.com/lordofthejars/smart-testing/blob/c4b1bbbc8c84f4e67ef522396cc9ccedde463d96/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java#L65

core/src/main/java/org/arquillian/smart/testing/impl/TestExecutionPlannerLoaderImpl.java
lordofthejars: It is just that it is clean, you don't need to pass the whole object when you only need this boolean.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/MisspelledStrategiesFunctionalTest.java
lordofthejars: There is no exception since it is a ftest, the error is found in the Maven log.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/MisspelledStrategiesFunctionalTest.java
lordofthejars: Same

MatousJobanek:
> https://github.com/arquillian/smart-testing/pull/213/files#diff-e5427bc921ef9faf2177273978bd12feR42 and https://github.com/lordofthejars/smart-testing/blob/c4b1bbbc8c84f4e67ef522396cc9ccedde463d96/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java#L65

but this is not for Surefire side, right?

lordofthejars:
@MatousJobanek This happens when maven extension is registering dependencies.

MatousJobanek:
>  This happens when maven extension is registering dependencies.

sure, that will work for Maven integration, but not for other integration cases such as Che or any other IDE

core/src/test/java/org/arquillian/smart/testing/impl/TestExecutionPlannerLoaderTest.java
```diff
@@ -52,7 +67,7 @@ public void should_throw_exception_when_no_matching_strategy_found() throws Exce
         expectedException.expectMessage("No strategy found for [new]. Available strategies are: [[dummy]]. Please make sure you have corresponding dependency defined.");
 
         // when
-        final TestExecutionPlanner exceptionShouldBeThrown = testExecutionPlannerLoader.getPlannerForStrategy("new");
+        final TestExecutionPlanner exceptionShouldBeThrown = testExecutionPlannerLoader.getPlannerForStrategy("new", false);
```
MatousJobanek: this whole test class is pretty weird - there is no `// then` comment/part

core/src/test/resources/strategies.properties
MatousJobanek: If it was removed, why this is added at the place: ` core/src/test/resources/strategies.properties`

docs/configuration.adoc
```diff
@@ -1,5 +1,11 @@
 == Configuration
 
+This section explains configuration parameters of Smart Testing.
+Notice that they are all summarized at <<Reference Card>> sectin.
```
MatousJobanek: `section` instead of `sectin`

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java
```diff
@@ -30,21 +32,41 @@ private void addStrategies(Model model) {
...
-        final Map<String, String> strategyMismatch = new HashMap<>();
-        for (final String definedStrategy : strategies) {
+        final List<String> strategyMismatch = new ArrayList<>();
+        final List<String> registeredStrategies = new ArrayList<>();
+        for (int i=0; i < strategies.length; i++) {
+            String definedStrategy = strategies[i];
             if (!dependencies.containsKey(definedStrategy)) {
                 final String closestMatch = stringSimilarityCalculator.findClosestMatch(definedStrategy, dependencies.keySet());
-                strategyMismatch.put(definedStrategy, closestMatch);
+                if (configuration.isAutocorrect()) {
```
MatousJobanek: The whole method `addStrategies` is pretty long now and the if hierarchy is quite complicated. Do you think that it could be sliced to multiple methods?

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java
```diff
@@ -56,10 +60,73 @@ public void should_not_add_surefire_dependency_as_it_is_present()
...
+    @Test
+    public void should_autocorrect_strategy()
+        throws Exception {
+        // given
+        Model model = prepareModelWithSurefirePlugin("2.20");
+        final Configuration conf = Configuration.load();
+        conf.setAutocorrect(true);
+        conf.setStrategies("nwe");
+        final DependencyResolver dependencyResolver = new DependencyResolver(conf);
+        model.addDependency(new DependencyResolver.SurefireApiDependency("2.19.1"));
```
MatousJobanek: Extract this whole given part to separated method - the same code is used in other test methods

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java
lordofthejars: I tried but then you get methods with 4 or 5 parameters, so I decided that it was better to maintain together.

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java
bartoszmajsak: Why this didn't happen?

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java
```diff
@@ -56,10 +60,73 @@ public void should_not_add_surefire_dependency_as_it_is_present()
...
+        // when
+        final Throwable exception = catchThrowable(() -> dependencyResolver.addRequiredDependencies(model));
+
+        // then
+        assertThat(exception).isInstanceOf(IllegalStateException.class);
+
+    }
+
+    @Test
+    public void should_fail_when_strategy_is_repeated_by_autocorrect()
```
bartoszmajsak: This test is identical to the one above expect of the order - is it relevant?

bartoszmajsak:
It's interesting to see that PR was merged to master when travis build failed (not to mention without any approval). Should we protect automatically from such situations to occur in the future? This would mean all merges will be performed by Travis if and only if the build is green (also maybe it should be only merged/triggered by the reviewer?). As a matter of fact, failure was caused by the test which should be run as non-thread-safe but being run in parallel. This is fixed now with https://github.com/arquillian/smart-testing/commit/e2186f33f9a6800e3246eb035b017e6ee898ceffbut.

When looking at this test, however, two things caught my attention:

1. [We have two tests which are duplicates](https://github.com/arquillian/smart-testing/blob/507695a1cfdcb27c727178e509f560ee539dd48e/mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java#L92-L124)
2. Feature covered by this test case makes me think it's probably the wrong behaviour to start with. Why should we fail if we have duplicated strategies? Shouldn't we just be smart enough to treat them as the same and register only once? We can obviously hint the user that something in his configuration is not right, but failing fast, in this case, is not really necessary from my point of view. This can be solved as simple as changing the data structure where we keep strategies.

lordofthejars:
It was merged with approval in mattermost. The problem was that it was not updated in PR.

bartoszmajsak:
Let's have a rule of thumb to have approvals where they really belong. So in the PR.

Otherwise we are loosing traceability.

lordofthejars:
> Why should we fail if we have duplicated strategies?

The descrption of the issue mention: "In case when more than one from existing strategies will be the same match for misspelled one we fail the build with the proper error message pointing which ones are having the same match." So I understand this is what is described.

bartoszmajsak:
So it was either poorly formulated or misunderstood. The intention here was that if we have a mistyped strategy for which we find two existing ones with the same matching point we should fail, as we don't know what user meant. So in this case `nwe,new` is not really reflecting this case as `new` is an existing strategy, but `nwe,enw` does.

bartoszmajsak:
So it was either poorly formulated or misunderstood. The intention here was that if we have a mistyped strategy for which we find two existing ones with the same matching point we should fail, as we don't know what user meant. So in this case `nwe,new` is not really reflecting this case as `new` is an existing strategy, so we end up with `new,new` -> `new`.

On the other hand if we would have strategies such as `test` and `set` and user would type `tset` there is only one transformation needed for `tset` to become either `test` or `set` - so both score with `1`. In this case we don't really know what was the user intention with the algorithm we have in place.

PR #220, opened 2017-10-10
Adds package annotation for affected tests

### LINKED ISSUE #199: Smart Testing annotations for Affected
##### Issue Overview

Currently affected strategy just works with black box testing(or white ones if ShrinkWrap or VertX runner is used). The solution would be to use coverage data as a way to detect these cases.

I am not sure when we are going to be able to research on this and also not sure if it will be possible after all to get all this information.

Then a possible solution might be creating a document annotation where developer can set the classes/regexp of classes that this test uses.

In this way we can either detect classes because they are imported or by analyzing the regexp.

lordofthejars:

 Adds package annotation for affected tests

- Adds `@Tests` annotation to set which production classes are affected by annotated tests. Useful for black box testing.
- Adds library to scan classes of given package. It is a really light library (no external dependencies), offers a lot of operations that we might use in the future. Also maybe it can be used as a substitution of Javassist.

Fixes #199


api/pom.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>smart-testing-parent</artifactId>
+    <groupId>org.arquillian.smart.testing</groupId>
+    <version>0.0.3-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>api</artifactId>
```
lordofthejars: Created a new artifact so users of this annotation don't need to import all smart testing affected strategy jar but just an special artifact. Since I think that in the future we might use more annotations, @MatousJobanek and me decided this would be as start a good place to put it.

pom.xml
```diff
@@ -47,6 +47,7 @@
     <version.assertj-core>3.8.0</version.assertj-core>
     <version.system-rules>1.16.1</version.system-rules>
     <version.javassist>3.21.0-GA</version.javassist>
+    <version.fast-classpath-scanner>2.7.4</version.fast-classpath-scanner>
```
lordofthejars: Really light project for scanning classpath. I think we can use more and more features it offers.

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/fakeproject/test/YTest.java
```diff
@@ -0,0 +1,19 @@
+package org.arquillian.smart.testing.strategies.affected.fakeproject.test;
+
+import org.arquillian.smart.testing.strategies.affected.Tests;
+import org.arquillian.smart.testing.strategies.affected.fakeproject.main.superbiz.Alone;
+import org.junit.Ignore;
+import org.junit.Test;
+
+// Test ignored because it is a test that is used to in tests and not to be run as real test by test runner
```
bartoszmajsak: I know what you are trying to say in this sentence, but for a person not familiar with test-bed and how we tests our test selection this sentence would be rather confusing - can you make it more digestable? :)

You can also put it as a part of `@Ignore`.

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraphTest.java
```diff
@@ -141,6 +145,48 @@ public void should_detect_all_changes_transitive() {
     }
 
     @Test
+    public void should_detect_all_changes_adding_package_annotated_transitive() {
```
bartoszmajsak: Can we have an ftest covering this new functionality?

lordofthejars:
Look I know it is not perfect solution, but it is a easy quick win solution for next simple case:

```java
@RunWith(SpringRunner.class)
@SpringBootTest
public class MyTests {

    @Autowired
    private MyInterface interface;
}
```

What's happen if I change is the implementation of `MyInterface`? Then this test is not going to be executed because the implementation is not as import.

Of course you might say that coverage approach is the solution, but it has some disadvantages, and being the most important that the information of coverage might become outdated, so we need to recalculate it often, and this means (or might means) either running the whole suite or have a way to just update the coverage information.

Another problem is what's happen if we have tests that checks if the service is up and running in our OPenShift cluster. Then it means that 1) the deployment code must have an agent or something to get the coverage, and 2) we need to copy this data from a remote site to some local storage, and the problem is that maybe you don't have access to that machine within the cluster.

About outdated information using annotation approach might happen but it is quite more difficult because of two things, the first one is that you can use a regexp so you can always set the base package which is something unlike that is going to happen, but also in case it happens, if we detect that an expression doesn't hit anything, then we just write a warning so user can see it.

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraphTest.java
lordofthejars: Yes, but not done yet until all discussions happens

lordofthejars:
Tasks to be done before this PR can be started to be reviewed_

- [ ] Write Functional test
- [ ] Update documentation with this option
- [ ] Measure the time spent when classpath scanning is done.

bartoszmajsak:
>  Write Functional test (this cannot be done until we release the version) The reason is that we ned to be able to annotate tests in dogfood project with smart testing annotations. Now just testing manually so we need a release to add it as reference. Now I have tested manually and it works. The idea is release a ST with this feature and then provide an automatic test.

Not sure if that is really a problem. If you can test it manually can't you automate? Refer to latest snapshot instead of fixed version.  This will work for the builds, as extension should be "installed" in the local repo prior to executing functional tests. Or am I missing something?

MatousJobanek:
> If you can test it manually can't you automate?

Yeah I agree with @bartoszmajsak 
How I would do it:
1. create a tag in dogfooding repo with the api dependency and the annotations used in the tests
   * the version element of the smart ST api dependency will contain some variable `smart.testing.version` that can be passed as a Maven build parameter / system property
2. in the functional test apply the commit (tag)
3. run the embedded maven build with the system property `smart.testing.version` set to the current snapshot version

lordofthejars:
Great idea, thanks Matous, I'll do it tomorrow.

bartoszmajsak:
> run the embedded maven build with the system property smart.testing.version set to the current snapshot version

Yeah, this would make dogfood repo buildable (in a sense of dependencies), so sure, that can work. It would be great to explain in the test itself why we do it though.

api/src/main/java/org/arquillian/smart/testing/strategies/affected/Tests.java
```diff
@@ -0,0 +1,42 @@
...
+ * Annotation used to set which production classes are tested in current test.
+ * By default it appends all classes defined in all attributes.
+ *
+ * If none of the attributes are set, then all production classes with same package as test and its subpackages are added.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Repeatable(TestsList.class)
+@Documented
+public @interface Tests {
```
bartoszmajsak: By looking at this annotation my first thought was that it will execute list of test classes defined in it, so the name is a bit ambiguous. Maybe `@Covers` or `@ComponentUnderTest` makes it more self-explanatory?

api/src/main/java/org/arquillian/smart/testing/strategies/affected/TestsList.java
```diff
@@ -0,0 +1,14 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Documented
+public @interface TestsList {
```
bartoszmajsak: Then this can become `@ComponentsUnderTest`

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraph.java
```diff
@@ -73,14 +79,86 @@ void buildTestDependencyGraph(Collection<File> testJavaFiles) {
...
     }
 
-    private void addToIndex(JavaElement javaElement, String[] imports) {
+    private List<String> calculateManualAddedDependencies(JavaClass testJavaClass) {
+        final List<String> manualDependencyClasses = new ArrayList<>();
+        final Tests[] allTestsAnnotation = getAllAnnotations(testJavaClass);
+
+        for (Tests tests : allTestsAnnotation) {
+            List<String> packages = getPackages(testJavaClass.packageName(), tests);
+            for (String packag : packages) {
```
bartoszmajsak: Typo `packag` :)

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraphTest.java
```diff
@@ -141,6 +145,48 @@ public void should_detect_all_changes_transitive() {
...
+        // given
+        final ClassDependenciesGraph
+            classDependenciesGraph = new ClassDependenciesGraph(new EndingWithTestTestVerifier());
+
+        final String testLocation = ZTest.class.getResource("ZTest.class").getPath();
+        classDependenciesGraph.buildTestDependencyGraph(Arrays.asList(new File(testLocation)));
+
+        // when
+        Set<File> mainObjectsChanged = new HashSet<>();
+        mainObjectsChanged.add(new File(Unwanted.class.getResource("Unwanted.class").getPath()));
```
bartoszmajsak: How about extracting this to a simple utility method `asFile("Class.class")`? Would make it a bit easier to follow, as we use it in few places.

bartoszmajsak: Or maybe just this part `Unwanted.class.getResource("Unwanted.class").getPath()`. From the test itself it's not really important how we fetch it and it makes it a bit blurry to follow.

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/fakeproject/test/YTest.java
```diff
@@ -0,0 +1,18 @@
+package org.arquillian.smart.testing.strategies.affected.fakeproject.test;
+
+import org.arquillian.smart.testing.strategies.affected.Tests;
+import org.arquillian.smart.testing.strategies.affected.fakeproject.main.superbiz.Alone;
+import org.junit.Ignore;
+import org.junit.Test;
+
+// Test ignored because it is used internally
```
bartoszmajsak: How about making this comment part of `@Ignore`?

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/fakeproject/test/ZTest.java
```diff
@@ -0,0 +1,18 @@
+package org.arquillian.smart.testing.strategies.affected.fakeproject.test;
+
+import org.arquillian.smart.testing.strategies.affected.Tests;
+import org.junit.Ignore;
+import org.junit.Test;
+
+// Test ignored because it is used internally
+@Ignore
```
bartoszmajsak: How about making this comment part of `@Ignore`?

docs/configuration.adoc
```diff
@@ -138,6 +138,36 @@ IMPORTANT: This strategy is currently only applicable for _white box_ testing ap
 
 WARNING: At this moment, this strategy does not work with Java 9.
 
+===== Explicitly Set
+
+By default affected strategy uses _imports_ of tests to build the graph of dependencies.
```
bartoszmajsak: Maybe `graph of collaborators` - dependency is rather broad term and could be understood that we build a graph of libraries we use based on imports

docs/configuration.adoc
```diff
@@ -138,6 +138,36 @@ IMPORTANT: This strategy is currently only applicable for _white box_ testing ap
...
+This approach is fine for unit tests (white box tests) but might not work in all cases of high level tests (black box test).
+
+There are some test technologies that allows you to deploy an application locally and then run tests against it.
+For example Wildfly Swarm has `@DefaultDeployment` annotation or for example Spring (Boot) deploys all application automatically.
```
bartoszmajsak: I would remove second `for example` in this statement. Also `deploys entire application` maybe?

docs/configuration.adoc
```diff
@@ -138,6 +138,36 @@ IMPORTANT: This strategy is currently only applicable for _white box_ testing ap
...
+
+This means that production classes are not directly imported into the test, so there is no way to get a relationship between test and production classes.
+For this reason `affected` provides an annotation to set package(s) of production classes that are deployed by the test.
+
+The first thing you need to do is register next artifact into your build script: `org.arquillian.smart.testing:api:<version>`.
```
bartoszmajsak: next artifact -> following artifact?

docs/configuration.adoc
```diff
@@ -138,6 +138,36 @@ IMPORTANT: This strategy is currently only applicable for _white box_ testing ap
...
+
+Then you can annotate your test with `org.arquillian.smart.testing.strategies.affected.Tests` annotation.
```
bartoszmajsak: Could be awesome to extract `org.arquillian.smart.testing.strategies.affected.Tests` automatically from the code - sounds like small doable extension?

lordofthejars: I am not sure if small but yes, let me open a new issue, since we are using the same approach in several places and several files I think it is better to implement as a separate PR.

bartoszmajsak: Yes, by all means, do so. I didn't mean to have it as part of this PR.

lordofthejars: https://github.com/arquillian/smart-testing/issues/256

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraph.java
lordofthejars: Well it is not a type, `package` is a reserved word, so I decided to put in this way. But any other suggestion is welcome :)

bartoszmajsak: That was late hahahaha. `pkg` maybe? :)

facepalm-real

PR #221, opened 2017-10-11
Initializing git change resolver if dir doesn't contains git folder.

### LINKED ISSUE #216: Should not register ST extension if user is trying to run it with project which doesn't have scm repository setup
##### Issue Overview
If your project doesn't have scm repository. `mvn clean test` is failing due to couldn't initialize `GitChangeResolver`

##### Expected Behavior
ST extension should not be registered if project doesn't have `SCM` repository.

##### Current Behavior
Getting exception as
``` bash
Exception in thread "main" java.util.ServiceConfigurationError: org.arquillian.smart.testing.scm.spi.ChangeResolver: Provider org.arquillian.smart.testing.scm.git.GitChangeResolver could not be instantiated
```
##### Steps To Reproduce
1. Create simple project where you can apply smart testing. 
2. Run smart testing with required configuration.

dipak-pawar:
We are initializing all implementation of `ChangeResolver` using ServiceLoader. Currently if we tries to initialize `GitChangeResolver` it's failing due to `IllegalArgumentException with gitdir is not present` which we haven't catched in either in constrcutor of `GitChangeResolver` or in `isApplicable` method. 

- Catching IllegalArgumentException in constructor of `GitChangeResolver` & `isApplicable` method to avoid unexpected exception while running tests

Fixes #216

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
+++ b/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
@@ -63,6 +63,17 @@ public class GitChangeResolver implements ChangeResolver {
         return diff(projectDir, scm.getRange().getTail(), scm.getRange().getHead());
     }
 
+    @Override
+    public Collection<Change> diff(File projectDir, Configuration configuration, String strategy) {
+        if (isApplicable(projectDir)) {
+            return diff(projectDir, configuration);
+        } else {
+            throw new IllegalStateException(
+                String.format("strategy %s needs scm to be initialized. Git is not initialized. "
+                    + "Please initialize git using `git init`", strategy));
+        }
+    }
+
     Set<Change> diff(File projectDir, String previous, String head) {
         buildGit(projectDir);
 
@@ -78,22 +89,18 @@ public class GitChangeResolver implements ChangeResolver {
 
     @Override
     public boolean isApplicable(File projectDir) {
-        try {
-            final FileRepositoryBuilder builder = new FileRepositoryBuilder();
-            builder.readEnvironment().findGitDir(projectDir).build();
-        } catch (IllegalArgumentException | IOException e) {
-            logger.warn("Working directory is not git directory. Cause: %s", e.getMessage());
-            return false;
-        }
-        return true;
+        return getFileRepositoryBuilder(projectDir).getGitDir() != null;
+    }
+
+    private FileRepositoryBuilder getFileRepositoryBuilder(File currentGitDir) {
+        return new FileRepositoryBuilder().readEnvironment().findGitDir(currentGitDir);
     }
 
     private void buildGit(File projectDir){
         closeGitIfExists();
-        final FileRepositoryBuilder builder = new FileRepositoryBuilder();
         try {
-            git = new Git(builder.readEnvironment().findGitDir(projectDir).build());
-        } catch (IOException e) {
+            git = new Git(getFileRepositoryBuilder(projectDir).build());
+        } catch (IOException | IllegalArgumentException e) {
             throw new IllegalArgumentException("Unable to find git repository for path " + projectDir.getAbsolutePath(), e);
         }
     }
--- a/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
+++ b/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
@@ -81,7 +81,7 @@ public class GitChangeResolver implements ChangeResolver {
         try {
             final FileRepositoryBuilder builder = new FileRepositoryBuilder();
             builder.readEnvironment().findGitDir(projectDir).build();
-        } catch (IOException e) {
+        } catch (IllegalArgumentException | IOException e) {
             logger.warn("Working directory is not git directory. Cause: %s", e.getMessage());
             return false;
         }
--- a/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
+++ b/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
@@ -89,17 +89,17 @@ public class GitChangeResolver implements ChangeResolver {
 
     @Override
     public boolean isApplicable(File projectDir) {
-        return getFileRepositoryBuilder(projectDir).getGitDir() != null;
+        return fileRepositoryBuilder(projectDir).getGitDir() != null;
     }
 
-    private FileRepositoryBuilder getFileRepositoryBuilder(File currentGitDir) {
+    private FileRepositoryBuilder fileRepositoryBuilder(File currentGitDir) {
         return new FileRepositoryBuilder().readEnvironment().findGitDir(currentGitDir);
     }
 
     private void buildGit(File projectDir){
         closeGitIfExists();
         try {
-            git = new Git(getFileRepositoryBuilder(projectDir).build());
+            git = new Git(fileRepositoryBuilder(projectDir).build());
         } catch (IOException | IllegalArgumentException e) {
             throw new IllegalArgumentException("Unable to find git repository for path " + projectDir.getAbsolutePath(), e);
         }
--- a/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
+++ b/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
@@ -123,8 +123,7 @@ public class GitChangeResolver implements ChangeResolver {
             final List<DiffEntry> commitDiffs = git.diff().setNewTree(newTree).setOldTree(oldTree).call();
             return transformToChangeSet(reduceToRenames(commitDiffs), repoRoot);
         } catch (MissingObjectException e) {
-            throw new IllegalArgumentException(format(WRONG_COMMIT_ID_EXCEPTION, e.getObjectId().getName(),
-                repository.getDirectory().getAbsolutePath()));
+            throw new IllegalArgumentException(format(WRONG_COMMIT_ID_EXCEPTION, e.getObjectId().getName(), repository.getDirectory().getAbsolutePath()));
         } catch (IOException | GitAPIException e) {
             throw new IllegalStateException(e);
         }
@@ -132,8 +131,7 @@ public class GitChangeResolver implements ChangeResolver {
 
     private void validateCommitExists(ObjectId retrievedId, String id, Repository repository) {
         if (retrievedId == null) {
-            throw new IllegalArgumentException(
-                format(WRONG_COMMIT_ID_EXCEPTION, id, repository.getDirectory().getAbsolutePath()));
+            throw new IllegalArgumentException(format(WRONG_COMMIT_ID_EXCEPTION, id, repository.getDirectory().getAbsolutePath()));
         }
     }
 
--- a/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
+++ b/core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
@@ -123,15 +123,17 @@ public class GitChangeResolver implements ChangeResolver {
             final List<DiffEntry> commitDiffs = git.diff().setNewTree(newTree).setOldTree(oldTree).call();
             return transformToChangeSet(reduceToRenames(commitDiffs), repoRoot);
         } catch (MissingObjectException e) {
-            throw new IllegalArgumentException(format(WRONG_COMMIT_ID_EXCEPTION, e.getObjectId().getName(), repository.getDirectory().getAbsolutePath()));
-        }catch (IOException | GitAPIException e) {
+            throw new IllegalArgumentException(format(WRONG_COMMIT_ID_EXCEPTION, e.getObjectId().getName(),
+                repository.getDirectory().getAbsolutePath()));
+        } catch (IOException | GitAPIException e) {
             throw new IllegalStateException(e);
         }
     }
 
     private void validateCommitExists(ObjectId retrievedId, String id, Repository repository) {
         if (retrievedId == null) {
-            throw new IllegalArgumentException(format(WRONG_COMMIT_ID_EXCEPTION, id, repository.getDirectory().getAbsolutePath()));
+            throw new IllegalArgumentException(
+                format(WRONG_COMMIT_ID_EXCEPTION, id, repository.getDirectory().getAbsolutePath()));
         }
     }
 
@@ -192,5 +194,4 @@ public class GitChangeResolver implements ChangeResolver {
             })
             .collect(Collectors.toSet());
     }
-
 }
--- a/core/src/main/java/org/arquillian/smart/testing/scm/spi/ChangeResolver.java
+++ b/core/src/main/java/org/arquillian/smart/testing/scm/spi/ChangeResolver.java
@@ -9,6 +9,17 @@ public interface ChangeResolver extends AutoCloseable {
 
     Collection<Change> diff(File projectDir, Configuration configuration);
 
+    /**
+     * If projectDir doesn't have scm initialized, then {@link IllegalStateException} is thrown.
+     *
+     * @param projectDir A directory from which you are calculating git diff changes.
+     * @param configuration A configuration used to configure Smart Testing.
+     * @param strategy name of strategy for which you are calculating diff.
+     *
+     * @return Collection of entire diff in projectDir.
+     */
+    Collection<Change> diff(File projectDir, Configuration configuration, String strategy);
+
     boolean isApplicable(File projectDir);
 
 }
--- a/core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
+++ b/core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
@@ -16,6 +16,7 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.junit.rules.TemporaryFolder;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -25,7 +26,12 @@ import static org.assertj.core.api.Assertions.tuple;
 public class GitChangeResolverTest {
 
     @Rule
-    public final TemporaryFolder gitFolder = new TemporaryFolder();
+    public final TemporaryFolder gitFolder = new TemporaryFolder(new File("/tmp"));
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    public static final String CUSTOM = "custom";
 
     private GitChangeResolver gitChangeResolver;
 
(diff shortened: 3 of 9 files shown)


core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
```diff
@@ -62,10 +62,10 @@ public GitChangeResolver(File dir, String previous, String head) {
         final FileRepositoryBuilder builder = new FileRepositoryBuilder();
         try {
             this.git = new Git(builder.readEnvironment().findGitDir(dir).build());
-        } catch (IOException e) {
-            throw new IllegalArgumentException("Unable to find git repository for path " + dir.getAbsolutePath(), e);
+            this.repoRoot = git.getRepository().getDirectory().getParentFile();
+        } catch (IllegalArgumentException | IOException e) {
```
MatousJobanek: Isn't there a better way than just catching the exception?
eg: `builder.readEnvironment().findGitDir(dir).getGitDir() != null`

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
```diff
@@ -59,23 +59,29 @@ public GitChangeResolver(File projectDir, Scm scm) {
...
+        final FileRepositoryBuilder fileRepositoryBuilder = getFileRepositoryBuilder();
+        if (fileRepositoryBuilder.getGitDir() != null) {
+            try {
+                this.git = new Git(fileRepositoryBuilder.build());
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+            this.repoRoot = git.getRepository().getDirectory().getParentFile();
+        } else {
+            logger.warn("Unable to find git repository for path " + dir.getAbsolutePath());
```
MatousJobanek: How about throwing an exception when any strategy using scm information is used....?

dipak-pawar: We can throw, but in that case user has to use disable flag to run even `mvn clean` because it's failing while registering extension.

So it's unnecessary burdon for user to use disable flag for mvn clean something like `mvn clean -Dsmart.testing.disable` to make clean runnable

MatousJobanek: Yeah, that's why I said "when any strategy using scm information is used"

dipak-pawar: How about "if user is using any strategy using scm information, then don't add any dependency related to that & warn user to use this strategy you need scm in place." this?

MatousJobanek: well, I think that it should fail, because the ST cannot work properly. At least this is what I would expect as a user.
So, here - at this line - I would warn: `"Unable to find a git repository for the path %s - some strategies won't work properly. Make sure it is a git repository."`
Then, when any strategy tries to get the scm information using method [diff](https://github.com/arquillian/smart-testing/pull/221/files#diff-f1255f2030b502bc7deffb7a6797da92R83) then check if the git is set otherwise throw an exception.

MatousJobanek:
btw, is it really WIP as it is in the title of the PR? If yes, then put the label there

dipak-pawar:
> btw, is it really WIP as it is in the title of the PR? If yes, then put the label there

It was. Now it's ready

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
```diff
@@ -59,23 +59,34 @@ public GitChangeResolver(File projectDir, Scm scm) {
...
+        final FileRepositoryBuilder fileRepositoryBuilder = getFileRepositoryBuilder(dir);
+        if (fileRepositoryBuilder.getGitDir() != null) {
+            try {
+                this.git = new Git(fileRepositoryBuilder.build());
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+            this.repoRoot = git.getRepository().getDirectory().getParentFile();
+        } else {
+            logger.warn(String.format("Unable to find a git repository for the path %s - new, changed, affected strategies won't work properly."
```
MatousJobanek: I wouldn't mention the concrete names of the strategies - if we change a name or add a new one, we won't probably remember that should be changed in here...

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
```diff
@@ -59,23 +59,34 @@ public GitChangeResolver(File projectDir, Scm scm) {
...
+        if (git != null) {
+            git.close();
+        }
     }
 
     @Override
     public Set<Change> diff() {
         final Set<Change> allChanges= new HashSet<>();
+        if (git == null) {
+            throw new IllegalStateException("Git is not initialized. new, changed, affected strategies needs scm to be initialized. "
```
MatousJobanek: same here with the concrete names. But I understand what you meant - to say which strategies cannot work if the project is not git repository.
There are two ways - return null and then throw the exception from the particular strategy, or provide strategy name as a parameter and incorporate it into the message.

core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
```diff
@@ -134,25 +170,29 @@ public void should_fetch_all_modified_files() throws IOException {
...
+        this.gitChangeResolver =
+            new GitChangeResolver(gitFolder.getRoot(), "07b181b", "e195e3767591fbd041e041877c541229afaac3c9");
 
         // when
         final Throwable throwable = catchThrowable(() -> gitChangeResolver.diff());
 
         // then
-        assertThat(throwable).isInstanceOf(IllegalArgumentException.class).hasMessageStartingWith("Commit id 'e195e3767591fbd041e041877c541229afaac3c9' is not found in");
+        assertThat(throwable).isInstanceOf(IllegalArgumentException.class)
+            .hasMessageStartingWith("Commit id 'e195e3767591fbd041e041877c541229afaac3c9' is not found in");
```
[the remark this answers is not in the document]
dipak-pawar: you mean file or tests. Because I added some tests in this file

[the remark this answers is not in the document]
MatousJobanek: ok, I meant lines :-)

core/src/main/java/org/arquillian/smart/testing/scm/spi/ChangeResolver.java
```diff
@@ -5,7 +5,7 @@
 
 public interface ChangeResolver extends AutoCloseable {
 
-    Collection<Change> diff();
+    Collection<Change> diff(String strategy);
```
MatousJobanek: Here I would create two methods `diff()` and `diff(String strategy)` with the javadoc explaining the usage and why there are two methods - also cases when the exception is thrown. `diff()` will use `null` as the default strategy value

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
```diff
@@ -59,23 +59,35 @@ public GitChangeResolver(File projectDir, Scm scm) {
...
+            git.close();
+        }
     }
 
     @Override
-    public Set<Change> diff() {
-        final Set<Change> allChanges= new HashSet<>();
+    public Set<Change> diff(String strategy) {
+        final Set<Change> allChanges = new HashSet<>();
+        if (!strategy.isEmpty() && git == null) {
```
MatousJobanek: In case the strategy is `null`, then you get NPE.
In addition, swap the conditions - it makes more sense to check if the git is null as the first condition

core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
```diff
@@ -25,7 +26,12 @@
 public class GitChangeResolverTest {
 
     @Rule
-    public final TemporaryFolder gitFolder = new TemporaryFolder();
+    public final TemporaryFolder gitFolder = new TemporaryFolder(new File("/tmp"));
```
MatousJobanek: `/tmp` is the default location of the temporary directory. Then there are subdirectories created by this rule. There is no need to specify the parent directory. In addition, I believe that the location may change based on the system settings

core/src/main/java/org/arquillian/smart/testing/scm/spi/ChangeResolver.java
dipak-pawar: I would use `""` as default value for `diff()`

core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
dipak-pawar: If you run tests using surefire it's setting `java.io.Tmp` to module's target dir due to which all temp folders are creating in `submodule/target`. But JGit's `findGitDir(tempDir)` method is checking for all it's parent for `.git` due to which it got `.git` directory which is of smart-testing git & tests fail.

This is not in IntelliJ as it won't set any system property like surefire

core/src/main/java/org/arquillian/smart/testing/scm/spi/ChangeResolver.java
MatousJobanek: Why? Is there any specific reason for it?
From my point of view, there shouldn't be anything set. The string containing something (even empty string) can be already considered as a string containing a content. Which means that if you don't want to set anything (just void) then you should pass the `null` as the best representation of "nothing".

dipak-pawar: just to avoid NPE in future

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
dipak-pawar: Here I assumed strategy as `""` not null. that's why I am checking `isEmpty()`

core/src/main/java/org/arquillian/smart/testing/scm/spi/ChangeResolver.java
MatousJobanek: Having the check https://github.com/arquillian/smart-testing/pull/221#discussion_r144232376 if the strategy is null, then you won't get the NPE...
The `GitChangeResolver.diff` is something that can be externally used, so it should properly check if the parameter is null or not, otherwise, you will get unexpected NPE. At least this is what I would expect and how it is usually checked/used. If I check (in case of public classes) whether the string is not provided then as the first thing I check if it is null or not.

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
MatousJobanek: see https://github.com/arquillian/smart-testing/pull/221#discussion_r144244237

core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
MatousJobanek: Ok, I didn't know that. Thx for the explanation. Bloody surefire...
Well, then we should find another way as in some cases this could fail with wrong permissions. Imagine also the case of using Windows - this path doesn't exist there.

core/src/main/java/org/arquillian/smart/testing/scm/spi/ChangeResolver.java
MatousJobanek: Anyway, I don't really insist on having `null` as the default value when the `diff()` method is used, but please, put the check (if it is null or not) there. To avoid NPE.

core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
MatousJobanek: The solution is using:
`gitFolder = Files.createTempDirectory("junit-").toFile();` 
 that will be created in `/tmp` by default

core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
```diff
@@ -59,20 +65,51 @@ public void should_fetch_all_files_from_first_commit_to_given_hash() throws Exce
...
 
         // then
         assertThat(diff).hasSize(18);
     }
 
     @Test
+    public void should_not_applicable_when_git_repository_is_not_initialized() throws IOException {
+        // given
+        gitFolder.delete();
+        gitFolder.create();
```
MatousJobanek: to avoid this workaround, move the tests methods to separated class - eg.: `GitChangeResolverWithoutGitTest` where, in the before method, you will just create the temporary directory and will not unpack the repository.

core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
```diff
@@ -59,20 +65,51 @@ public void should_fetch_all_files_from_first_commit_to_given_hash() throws Exce
...
+
+        this.gitChangeResolver = new GitChangeResolver(gitFolder.getRoot(), "HEAD", "HEAD~0");
+
+        // then
+        thrown.expect(IllegalStateException.class);
+
+        // when
+        gitChangeResolver.diff(CUSTOM);
```
MatousJobanek: Also, verify the case when the strategy name is not specified: 
`gitChangeResolver.diff()` 
In this case, it should return the empty set.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -124,11 +124,13 @@ public void afterSessionEnd(MavenSession session) throws MavenExecutionException
     private void calculateChanges() {
         final Iterable<ChangeResolver> changeResolvers =
             new JavaSPILoader().all(ChangeResolver.class, ChangeResolver::isApplicable);
-        final Collection<Change> changes = stream(changeResolvers.spliterator(), false).map(ChangeResolver::diff)
+        final Collection<Change> changes = stream(changeResolvers.spliterator(), false)
+            .map(changeResolver -> changeResolver.diff(""))
```
MatousJobanek: Yeah, don't forget a separated method in the API: `changeResolver.diff()`

core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
dipak-pawar: @MatousJobanek This won't work. Because internally `createTempDirectory` with look for system property `java.io.tmpdir` if parent folder not specified.

In our case we are using jboss-parent as  a parent for `smart-testing/pom.xml` which has `java.io.tmpdir` set to `${project.build.directory}` due to which this api will create tmpdir in current porject's build dir.

core/src/main/java/org/arquillian/smart/testing/scm/git/GitChangeResolver.java
```diff
@@ -78,22 +89,18 @@ private void closeGitIfExists() {
...
-            builder.readEnvironment().findGitDir(projectDir).build();
-        } catch (IOException e) {
-            logger.warn("Working directory is not git directory. Cause: %s", e.getMessage());
-            return false;
-        }
-        return true;
+        return getFileRepositoryBuilder(projectDir).getGitDir() != null;
+    }
+
+    private FileRepositoryBuilder getFileRepositoryBuilder(File currentGitDir) {
```
bartoszmajsak: I would just name it `fileRepositoryBuilder`

core/src/test/java/org/arquillian/smart/testing/scm/git/GitChangeResolverTest.java
bartoszmajsak: Is the workaround we discussed offline yesterday working for you?

dipak-pawar: no it's not working. Because We won't get `originial.tmp.dir` as in jboss-parent pom.xml it's defined as `${project.build.directory}`. So it's evaluating ``${project.build.directory}` as original tmp dir

dipak-pawar: I am doing it [this](https://github.com/arquillian/smart-testing/pull/221/files#diff-d0f8b495a0c54a61f1eb2e9f2ed992d4R29) way

MatousJobanek: I've just noticed that when I was trying to fix it locally, I did some additional changes and I was setting it manually to `/tmp` in the code...
But exposing another property via surefire config:
```xml
      <plugin>
        <configuration>
          <systemPropertyVariables>
            <original.java.tmp.dir>${java.io.tmpdir}</original.java.tmp.dir>
          </systemPropertyVariables>
        </configuration>
      </plugin>
```
should work. At least, in my case it returns:
```java
"java.io.tmpdir" -> "/home/mjobanek/arquillian/smart-testing/core/target"
"original.java.tmp.dir" -> "/tmp"
```
So technically, it should be also possible to override the settings from the parent and set the same variable to the original one:
```xml
<java.io.tmpdir>${java.io.tmpdir}</java.io.tmpdir>
```

MatousJobanek:
Please, incorporate my comments:
https://github.com/arquillian/smart-testing/pull/221#discussion_r144545602
https://github.com/arquillian/smart-testing/pull/221#discussion_r144546889

dipak-pawar:
@MatousJobanek 
> https://github.com/arquillian/smart-testing/pull/221#discussion_r144545602 (Don't fail when no strategy name is provided)
 
We are using `isApplicable()` method to load all implementations of `ChangeResolver`.  In this PR we changed logic of [isApplicable()](https://github.com/arquillian/smart-testing/pull/221/files#diff-f1255f2030b502bc7deffb7a6797da92R91) such a that it won't fail if you haven't initialized git in your project. Also for this we don't need any strategy.

Now we are calling`Collection<Change> diff(File projectDir, Configuration configuration, String strategy) ` from all implementation of `TestExecutionPlanner`.

Case 1:
If we called it on project which has git initialized, with strategy null, it won't fail, it will give you all required changes as git is initialized.

Case 2:
If we called it on project which has not git initialized, with strategy null, it will fail with `git is not initialized for null strategy`. We are using strategy just for logging msg.



> https://github.com/arquillian/smart-testing/pull/221#discussion_r144546889 (Test for "SmartTestingMavenConfigurer use case")

I can write test for this with strategy `null`, but again it will fail with msg `strategy null needs scm to be initialized` which is same scenario like some random name for which I already added some tests. There is no point to add one more tests for this.

I think we don't need to check this.

MatousJobanek:
OK, I missed that part. Thx for explanation @dipak-pawar

Issue #218, opened 2017-10-10
NoSuchMethodError  `ProviderParameters.getConsoleLogger()` with surefire 2.20.1


dipak-pawar:
##### Issue Overview
If you are trying to run test with surefire version `2.20.1` with current master, it throwing exception without running tests.

It runs fine with surefire `2.19.1`

##### Expected Behavior
Test should run fine with Smart testing enabled.

##### Current Behavior
Getting exception as 
```bash
org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
[ERROR] java.lang.NoSuchMethodError: org.apache.maven.surefire.providerapi.ProviderParameters.getConsoleLogger()Lorg/apache/maven/surefire/report/ConsoleLogger;
```

##### Steps To Reproduce
1. Clone https://github.com/dipak-pawar/configuration-file-demo
2. Install Smart testing snapshot with latest master using `mvn clean install -DskipTests` from smart testing code base 
2. Execute `mvn clean test` 

##### Additional Information
```
 Apache Maven 3.3.9 (NON-CANONICAL_2016-04-07T23:15:34Z_mockbuild; 2016-04-08T04:45:34+05:30)
Maven home: /usr/share/maven
Java version: 1.8.0_101, vendor: Oracle Corporation
Java home: /opt/jdk1.8.0_101/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.7.9-200.fc24.x86_64", arch: "amd64", family: "unix"
</details>
```

dipak-pawar:
I am not sure if this has something to do with #201

hemanik:
Yes @dipak-pawar , It's related to #201.

MatousJobanek:
yeah, it is the same issue, so I'm closing this as a duplication of https://github.com/arquillian/smart-testing/issues/201

PR #223, opened 2017-10-11
Adds support for different instances of surefire logger 

### LINKED ISSUE #201: The Surefire logger instance should be retieved using reflection
##### Issue Overview

In the Surefire 2.20 there was changed the surefire logger class from `ConsoleLogger` to `ConsoleStream` [commit](https://github.com/apache/maven-surefire/commit/6a79127ab63273f0f1f5268c65e4806761871a46#diff-83d6b3fc3aabbf3c0a10152241387225)
The method in `ProviderParameters` class has the same name, just the return type is different

##### Expected Behaviour

Getting and calling the logger provided by surefire should be done by reflection - to be able to use both classes

##### Current Behaviour

We expect that only the class `ConsoleLogger` is returned

hemanik:
Surefire Loggers returns an instance of `ConsoleLogger` in version 2.19 and an instance of `ConsoleStream` in newer versions from 2.20 on wards.

Support for both the versions is now included with use of reflections.

- Use reflections to identify the correct Logger instance.
- Adds option to configure surefire version in test bed.

Fixes #201 #218

### DIFF
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
@@ -1,6 +1,8 @@
 package org.arquillian.smart.testing.ftest.testbed.project;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.net.ServerSocket;
@@ -10,6 +12,10 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.apache.maven.model.Model;
+import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
+import org.apache.maven.model.io.xpp3.MavenXpp3Writer;
+import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
 
 import static java.lang.System.getProperty;
 import static java.util.Arrays.stream;
@@ -183,8 +189,20 @@ public class BuildConfigurator {
         return this;
     }
 
-    public BuildConfigurator useMavenVersion(String mavenVersion){
+    public BuildConfigurator useMavenVersion(String mavenVersion) {
         this.mavenVersion = mavenVersion;
+        return this;
+    }
+
+    public BuildConfigurator useSurefireVersion(String version) {
+        File pomFile = new File(projectBuilder.getRoot().toAbsolutePath().toString()+ File.separator + "pom.xml");
+        try {
+            Model model = new MavenXpp3Reader().read(new FileInputStream(pomFile));
+            model.getProperties().setProperty("version.surefire.plugin", version);
+            new MavenXpp3Writer().write(new FileOutputStream(pomFile), model);
+        } catch (IOException | XmlPullParserException e) {
+            throw new RuntimeException("Failed setting surefire version in pom.", e);
+        }
         return this;
     }
 
@@ -293,4 +311,5 @@ public class BuildConfigurator {
     void setUsingInstallation(Using usingInstallation) {
         this.usingInstallation = usingInstallation;
     }
+
 }
--- a/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
+++ b/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
@@ -2,11 +2,14 @@ package org.arquillian.smart.testing.surefire.provider;
 
 import java.io.File;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+import java.util.Optional;
 import java.util.Set;
 import org.apache.maven.surefire.cli.CommandLineOption;
 import org.apache.maven.surefire.providerapi.ProviderParameters;
 import org.apache.maven.surefire.providerapi.SurefireProvider;
-import org.apache.maven.surefire.report.ConsoleLogger;
 import org.apache.maven.surefire.report.ReporterException;
 import org.apache.maven.surefire.suite.RunResult;
 import org.apache.maven.surefire.testset.TestSetFailedException;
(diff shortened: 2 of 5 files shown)


surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
```diff
@@ -102,4 +107,27 @@ private File findFirstMatchingPom(File source) {
...
+    private Object getConsoleLogger() {
+        try {
+            Optional<Method> method = Arrays.stream(bootParams.getClass().getMethods())
+                .filter(this::isGetConsoleLoggerMethod)
+                .findFirst();
+            if (method.isPresent()) {
+                return method.get().invoke(bootParams);
+            }
+        } catch (IllegalAccessException | InvocationTargetException e) {
+            new DefaultLoggerFactory()
```
lordofthejars: I think you should throw an exception as well, because if not then this method will return null, which means that `consoleLogger` will be null and then when it is going to be called, we will get a `NPE`

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
```diff
@@ -102,4 +107,27 @@ private File findFirstMatchingPom(File source) {
...
+                .getLogger()
+                .warn(NOT_COMPATIBLE_MESSAGE);
+        }
+        return null;
```
lordofthejars: The same as previous comment, if someday Surefire does not have a method called `getConsoleLogger` then a NPE will be thrown.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/logger/SurefireProviderLoggerFactory.java
```diff
@@ -25,40 +32,69 @@ public Logger getLogger() {
...
+        private SurefireProviderLogger(Object consoleLogger) {
             this.consoleLogger = consoleLogger;
+            Optional<Method> method = Arrays.stream(consoleLogger.getClass().getMethods())
+                .filter(this::isPublicConsoleLogMethod)
+                .findFirst();
+
+            if (method.isPresent()) {
+                logMethod = method.get();
+            } else {
+                new DefaultLoggerFactory()
```
lordofthejars: Is it safe to continue having `logMethod` null?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
MatousJobanek: This should have been handled here: https://github.com/arquillian/smart-testing/pull/223/files#diff-72e56dc4b88e2415a66df5915b9e4b6cR40
there should be `if`

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/logger/SurefireProviderLoggerFactory.java
MatousJobanek: This is managed here: https://github.com/arquillian/smart-testing/pull/223/files#diff-72e56dc4b88e2415a66df5915b9e4b6cR76

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
MatousJobanek: The reason why I don't want to throw an exception is that incompatible logging doesn't have to mean that the whole ST tool won't work.

MatousJobanek: > there should be if

fixed now

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
MatousJobanek: the console logger is intended to be used internally only in `SurefireProviderLogger` implementation - it is not exposed anywhere

hemanik:
Added a tag for the surefire version change in the dogfooding repo to verify the different logger implementations.
The tag details are as follows:
```
tag surefire_version
Tagger: hemanik <hemanikatyal@gmail.com>
Date:   Thu Oct 12 15:05:27 2017 +0530

Change surefire version to 2.20.1

commit 91d99ffada7bdb2a73f7b23c628fa1b119110766 (HEAD -> skipping-tests, tag: surefire_version)
Author: hemanik <hemanikatyal@gmail.com>
Date:   Thu Oct 12 15:03:27 2017 +0530

    affected: Change surefire version to 2.20.1

diff --git a/pom.xml b/pom.xml
index ed20c06..8f912fd 100644
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,7 @@
     <plugins>
       <plugin>
+        <version>2.20.1</version>
         <configuration>
           <redirectTestOutputToFile>true</redirectTestOutputToFile>
           <trimStackTrace>false</trimStackTrace>
```

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
```diff
@@ -61,7 +67,8 @@ public BuildConfigurator withRemoteDebugging() {
     }
 
     /**
-     * @param port if set to BuildConfigurator.RANDOM_PORT it will take random, free port
+     * @param port
+     *     if set to BuildConfigurator.RANDOM_PORT it will take random, free port
```
MatousJobanek: please, format only those lines that are changed by you... `Only VCS changed text`

hemanik: Done

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/BuildConfigurator.java
```diff
@@ -183,11 +189,23 @@ public BuildConfigurator excludeProjects(String... projects) {
...
+    public BuildConfigurator useMavenVersion(String mavenVersion) {
         this.mavenVersion = mavenVersion;
         return this;
     }
 
+    public BuildConfigurator useSurefireVersion(String version) {
+        File pomFile = new File(projectBuilder.getRoot().toAbsolutePath().toString()+ File.separator + "pom.xml");
+        try {
+            Model model = new MavenXpp3Reader().read(new FileInputStream(pomFile));
+            model.getProperties().setProperty("version.surefire.plugin", version);
```
dipak-pawar: What if my repository is has defined surefire configuration like 

```xml
<plugin>
        <version>2.19.1</version>
  </plugin>
```
I believe it will take defined in surefire plugin not in properties.

MatousJobanek: Yes, exactly... It should have been set there.

MatousJobanek: In other words, check if the pom contains the plugin definition, if yes then set the version there. If the plugin is not specified, then add the whole plugin definition there.

hemanik: I have checked for this case and it works when set in properties. And this code is only for the test bed, so I don't expect the user to change the structure of the pom.

hemanik: Setting it inside the plugin configuration section of the pom requires extensive parsing of the pom, already implemented as part of this [PR](https://github.com/arquillian/smart-testing/pull/223).

If you review that and it looks fine, I can follow the same here as well.

MatousJobanek: Yeah, you are right - sorry for confusions. I was applying//trying it on a wrong project. It works with the dogfooding repo.

MatousJobanek:
Landed in: https://github.com/arquillian/smart-testing/commit/4d5c7f8b02a7c3183b6919bfaf7edc20459fb1fa Thank you @hemanik 👍

PR #224, opened 2017-10-12
Fixes typo in usage document example.

hemanik:
Command duplicated in usage example.

- Deletes duplicate command in usage.
- Adds copy to clipboard option for smart testing extension.

Fixes #


dipak-pawar:
@hemanik As you are already there Can you add `copyToClipBoard` option for config file in ref card http://arquillian.org/smart-testing/#_configuration_file_2?

PR #225, opened 2017-10-12
Using  `LocalStorage` to write modified pom in `afterProjectsRead`.

### LINKED ISSUE #219: Should delete target directory completely for mvn clean if configuration file has debug true enabled
##### Issue Overview
Currently there is no way to maven extension to check which lifecycle of maven user is trying to execute. Due to which extension is trying to install all required dependency/modification of pom.xml even in `clean, compile` phase.

Now consider this scenario:
You have `smart-testing.yml` is defined as configuration file which has debug enabled in it
```yaml
mode: selecting
strategies: new
debug : true
``` 
Now if user is trying to execute `mvn clean` then as per maven convention it should delete target directory.
But as we are purging all the required reports/effective-poms in `afterSessionEnd` method. It's not deleting target directory because it has purged effective-pom.xml in 
`smart-testing/smart-testing-effective-pom.xml`
This should not be the case for `mvn clean`

##### Expected Behaviour
`mvn clean` should delete target directory completely

##### Current Behaviour
`mvn clean` is keeping purged resources in target directory 

##### Steps To Reproduce
1. Run smart testing using `mvn clean test` on any project with following content for `smart-testing.yml`
```yaml
mode: selecting
strategies: new
debug : true
``` 
2. Again run `mvn clean`
3. Look for `target` directory inside your project & it's content

### LINKED ISSUE #230: [Enhancement] Use LocalStorage api to write modified pom.xml
Currently we are using file based api to write modified pom.xml.

Now we have nice api of `LocalStorageFileAction` which we can use to write updated pom.xml

dipak-pawar:

* Now using `LocalStorage` to write modified pom.xml

Fixes #230

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/hub/storage/local/DuringExecutionLocalStorage.java
+++ b/core/src/main/java/org/arquillian/smart/testing/hub/storage/local/DuringExecutionLocalStorage.java
@@ -53,6 +53,7 @@ public class DuringExecutionLocalStorage extends AfterExecutionLocalStorage {
             Path dirToStore = new LocalStorage(rootDir)
                 .afterExecution(targetDir)
                 .getPathTo(dirNameToStore);
+
             FileSystemOperations.copyDirectory(dirToCopy, dirToStore, true);
         }
         FileSystemOperations.deleteDirectory(Paths.get(rootDir, SMART_TESTING_WORKING_DIRECTORY_NAME), true);
--- a/core/src/main/java/org/arquillian/smart/testing/hub/storage/local/DuringExecutionLocalStorage.java
+++ b/core/src/main/java/org/arquillian/smart/testing/hub/storage/local/DuringExecutionLocalStorage.java
@@ -3,15 +3,11 @@ package org.arquillian.smart.testing.hub.storage.local;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
-import org.arquillian.smart.testing.logger.Log;
-import org.arquillian.smart.testing.logger.Logger;
 
 /**
  * Class that takes care of storing/managing files and directories used and stored during the test execution.
  */
 public class DuringExecutionLocalStorage extends AfterExecutionLocalStorage {
-
-    private static final Logger logger = Log.getLogger();
 
     public static final String SMART_TESTING_WORKING_DIRECTORY_NAME = ".smart-testing";
     public static final String TEMPORARY_SUBDIRECTORY = "temporary";
@@ -57,7 +53,6 @@ public class DuringExecutionLocalStorage extends AfterExecutionLocalStorage {
             Path dirToStore = new LocalStorage(rootDir)
                 .afterExecution(targetDir)
                 .getPathTo(dirNameToStore);
-            logger.debug("copying directories from %s to %s", dirToCopy, dirToStore);
             FileSystemOperations.copyDirectory(dirToCopy, dirToStore, true);
         }
         FileSystemOperations.deleteDirectory(Paths.get(rootDir, SMART_TESTING_WORKING_DIRECTORY_NAME), true);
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/arquillian/smart/testing/hub/storage/local/DuringExecutionLocalStorage.java
```diff
@@ -53,7 +57,7 @@ private void storeDirectory(String dirNameToStore, String targetDir) {
             Path dirToStore = new LocalStorage(rootDir)
                 .afterExecution(targetDir)
                 .getPathTo(dirNameToStore);
-
+            logger.debug("copying directories from %s to %s", dirToCopy, dirToStore);
```
MatousJobanek: Is it necessary? Imagine what it will look like in multi-module project... A lot of same lines that don't contain any additional value... Or is there any use case I don't see?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ModifiedPomExporter.java
```diff
@@ -26,14 +26,15 @@ static void exportModifiedPom(Model model) {
...
-        logger.debug("Modified pom stored at: " + target);
+    private static void writeModifiedPomToTarget(StringWriter modifiedPom, File projectDir) throws IOException {
+        final Path modifiedPomPath = new LocalStorage(projectDir)
+            .duringExecution()
+            .toReporting()
+            .file(SMART_TESTING_POM_FILE)
+            .create();
+
+        logger.debug("Storing modified pom.xml at:" + modifiedPomPath);
+        Files.write(modifiedPomPath, modifiedPom.toString().getBytes());
```
MatousJobanek: Just letting you know that you can do also this:
```java
        final LocalStorageFileAction modifiedPomAction = new LocalStorage(projectDir)
            .duringExecution()
            .toReporting()
            .file(SMART_TESTING_POM_FILE);

        logger.debug("Storing modified pom.xml at:" + modifiedPomAction.getPath());
        modifiedPomAction.create(modifiedPom.toString().getBytes());
```

core/src/main/java/org/arquillian/smart/testing/hub/storage/local/DuringExecutionLocalStorage.java
dipak-pawar: While storing modified pom.xml we are logging that it'll be stored at `.smart-testing/reporting/` directory. But at the end of execution we are moving it `target` folder. But then if there is no logs about this. Then user will try to find modified pom.xml in `.smart-testing` which won't be there at the end of execution. 

So it's quite difficult for user to identify where is it?

So I had two options:
Either we should not log where modified pom.xml is stored initially or log where is it stored & moved

MatousJobanek: Yeah, now I probably understand why you did this. You store the effective pom file before the test execution, then (after test execution) the dir is moved to target, so you probably wanted to show where the effective pom can be found... To be honest, I wouldn't associate these two lines so I would be still trying to find the effective pom inside of the original directory - you know, I see a [line](https://github.com/arquillian/smart-testing/pull/225/files#diff-1848634e0d5e47585ae03dfabcf1e835R36) saying that the effective pom is stored somewhere, then I would expect that it is still there.
That is the reason why I didn't want to store the effective pom files before test execution, but after...

dipak-pawar: maybe we can give info like "effective pom.xml stored in `.smart-testing` for temporary but you can find it in `target/smart-testing` at the end of execution

dipak-pawar: but look we are not traversing again for all modules if we did it duringExecution phase which will save few seconds for big multi-module project

MatousJobanek: wouldn't solve all the problems just storing it once in `afterSessionEnd` phase?

dipak-pawar: but again we need to traverse all projects once in `afterSessionEnd` phase

Issue #226, opened 2017-10-12
SmartTestingSoftAssertions is not working as per expectation


dipak-pawar:
##### Expected Behaviour
Test should fail with all failures as this is SoftAssertions.
Here all failures means we are building `core/impl-base`. So it's writing effective-pom.xml in following directory:
* ```config/api/target/smart-testing```
*  ```config/impl-base/target/smart-testing```

![screenshot from 2017-10-12 17-46-18](https://user-images.githubusercontent.com/7941617/31495444-48c2cd78-af75-11e7-8d62-287010b3b343.png)

So it should assert all assertions at the end. But it's not doing that.

##### Current Behaviour
Test is passing.

##### Steps To Reproduce
1. Go to this [test](https://github.com/arquillian/smart-testing/blob/master/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/DebugModeAndMaven325FunctionalTest.java#L113)
2. Change this line from

```java 
softly.assertThat(projectBuilder.getBuiltProject()).allBuiltSubModulesContainEffectivePom(EFFECTIVE_POM);
``` 
to
``` java
softly.assertThat(projectBuilder.getBuiltProject()).allBuiltSubModulesContainEffectivePom("foo.txt);
```
3. Run the test in Intellij or using `mvn clean test`

Note: Please verify `allBuiltSubModulesWithTestExecutionsContainReport` case as well for assert softly

PR #227, opened 2017-10-12
fix(assertions) Finds effective poms recursively

MatousJobanek:
Finds effective poms recursively otherwise the test is green when the expected file is not present in one of the submodules.

fixes: #226


MatousJobanek:
Yeah, I forgot to add the second file to the commit :-D

Issue #228, opened 2017-10-12
ChangeResolver and ChangeStorage should use correct project directory path 


MatousJobanek:
##### Issue Overview

When the `SmartTesting` API is used, the classes `ChangeResolver` and `ChangeStorage` don't reflect the directory set using method `in`. As a side effect, the strategy `affected` doesn't work correctly in Che - the `SmartTesting` API is called from a different directory than the project root is.

##### Expected Behaviour

When I use `SmartTesting().with(....).in("path/to/project").apply...` the affected strategy should work.

##### Current Behaviour

the affected strategy throws an exception as it cannot find git directory.

##### Steps To Reproduce
1. Call `SmartTesting().with(....).in("path/to/project").apply...` from different directory than the project root is.

PR #229, opened 2017-10-12
fix: ChangeResolver and ChangeStorage are using correct pahts

MatousJobanek:

The classes `ChangeResolver` and `ChangeStorage` are using correct paths set using API method `in`. Similar changes have been applied also to some other parts of the code to have consistent behavior.

Fixes: #228

### DIFF
--- a/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
+++ b/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
@@ -42,7 +42,7 @@ class SmartTestingMavenConfigurer extends AbstractMavenLifecycleParticipant {
     @Requirement
     public org.codehaus.plexus.logging.Logger mavenLogger;
 
-    private final ChangeStorage changeStorage = new LocalChangeStorage(".");
+    private final ChangeStorage changeStorage = new LocalChangeStorage();
 
     private Configuration configuration;
 
@@ -50,8 +50,9 @@ class SmartTestingMavenConfigurer extends AbstractMavenLifecycleParticipant {
 
     @Override
     public void afterProjectsRead(MavenSession session) throws MavenExecutionException {
+        File projectDirectory = session.getCurrentProject().getModel().getProjectDirectory();
         Log.setLoggerFactory(new MavenExtensionLoggerFactory(mavenLogger));
-        configuration = Configuration.load();
+        configuration = Configuration.load(projectDirectory);
 
         Log.setLoggerFactory(new MavenExtensionLoggerFactory(mavenLogger, configuration));
         logger = Log.getLogger();
@@ -65,7 +66,7 @@ class SmartTestingMavenConfigurer extends AbstractMavenLifecycleParticipant {
 
         if (configuration.areStrategiesDefined()) {
             configureExtension(session, configuration);
-            calculateChanges();
+            calculateChanges(projectDirectory);
             Runtime.getRuntime().addShutdownHook(new Thread(() -> purgeLocalStorage(session)));
         } else {
             logStrategiesNotDefined();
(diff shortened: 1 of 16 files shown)


mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -42,16 +42,17 @@
...
-    private final ChangeStorage changeStorage = new LocalChangeStorage(".");
+    private final ChangeStorage changeStorage = new LocalChangeStorage();
 
     private Configuration configuration;
 
     private Boolean skipExtensionInstallation;
 
     @Override
     public void afterProjectsRead(MavenSession session) throws MavenExecutionException {
+        File projectDirectory = session.getCurrentProject().getModel().getProjectDirectory();
```
dipak-pawar: This won't give you correct projectDir if you are running in some module  which again contains module.
e.g.
```xml
parent
 - config
   -- api
     --- pom.xml
   -- impl-base
      --- pom.xml
   -- spi
      --- pom.xml
```
```bash
cd parent && mvn clean test
```
This method will return `config/api` as projectDir. which is not expected. It should return `config` as basedir.

We should use something like
`File projectDirectory = session.getTopLevelProject().getModel().getProjectDirectory()`

MatousJobanek: Good catch! I wouldn't notice this... Thx 👍

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -42,16 +42,17 @@
...
         Log.setLoggerFactory(new MavenExtensionLoggerFactory(mavenLogger));
-        configuration = Configuration.load();
+        configuration = Configuration.load(projectDirectory);
```
dipak-pawar: Suppose this is multi-module project:
```xml
parent
 - config
   -- api
     --- pom.xml
   -- impl-base
      --- pom.xml
   -- spi
      --- pom.xml
 - smart-testing.yml
```
You are trying to execute tests using following:
```bash
cd parent && mvn clean test
```
Assume you have configuration file present in parent dir. Then this method will take wrong configuration as it couldn't find configuration in it's respective module i.e. config.

Here we should look give correct projectdir to `Configuration.load(), otherwise it won't work as per expectation.

MatousJobanek: I probably don't understand this comment.
is the command correct? `cd parent && mvn clean test` ?
if I execute it from parent and the file is in parent, then it should be fine, right?
If you execute it from some submodule, then it won't find the file, which is fine as well I guess. Otherwise, you would need to go through all parent directories looking for some occurrence. I'm not sure if this is correct. If you execute maven build from subdirectory that doesn't contain "pom" file, then it will fail and won't try to find any pom file in the parent directories.

dipak-pawar: but in our case assume that in future we support configuration file per module in future then it will work as per expected if user is executing test goal in sub-module containing config-file.

Assume user is trying to run test from submodule directory. 
sub-module doesn't contain any `smart-testing.yml` but root of project contains `smart-testing.yml` where all smart-testing configuration present.

So Ideally in our case it should take configuration from parent `smart-testing.yml`.

Now it's looking for `smart-testing.yml` in submodule where is it not present. then it'll take all default configuration which is wrong

dipak-pawar:
We decided to create new issue for above scenario

cc @MatousJobanek

PR #233, opened 2017-10-13
fix(goals) Extension checks goals set for the build if any test will be executed

### LINKED ISSUE #219: Should delete target directory completely for mvn clean if configuration file has debug true enabled

MatousJobanek:

In the Maven extension, there is verified if any goal invoking test execution is set. Also, if the target is not present (when the session ends) in any of the modules, then it won't be created by ST extension

- Checking if any of the set goals for the build is present in the list:
`{"test", "prepare-package", "package", "pre-integration-test", "integration-test", "post-integration-test", "verify", "install", "deploy", "pre-site", "site", "post-site", "site-deploy"}`
assuming that only these goals can execute a test.
- I'm also checking if any of the provided goals contains colon `:` - for example in the case: 
`org.apache.maven.plugins:maven-surefire-plugin:2.20.1:test`
the check was copied from [Maven internals](https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecycleTaskSegmentCalculator.java#L150-L153)
- I have refactored the skip installation logic - the system property related code has been moved to one class `SkipInstallationChecker` (that also contains check for the goals) that is executed as the first thing - before the configuration file is loaded
- when the session ends, it is checked if there is any target preset, if not, no modified pom is exported and `reporting` directory from `.smart-testing` is not moved.

Fixes #219

### DIFF
--- a/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SkipInstallationChecker.java
+++ b/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SkipInstallationChecker.java
@@ -2,6 +2,7 @@ package org.arquillian.smart.testing.mvn.ext;
 
 import java.util.Arrays;
 import java.util.List;
+import org.apache.maven.execution.MavenSession;
 
 import static org.arquillian.smart.testing.mvn.ext.MavenPropertyResolver.isSkipTestExecutionSet;
 import static org.arquillian.smart.testing.mvn.ext.MavenPropertyResolver.isSpecificTestClassSet;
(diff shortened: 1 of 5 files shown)


lordofthejars:
One question, is it possible to mask test goal in Maven? Something like in Gradle, so for example you can do something like `mvn clean mysupergoal` which internally calls `test` goal.

MatousJobanek:
Maybe, I don't know... I don't know this feature in Gradle either...

lordofthejars:
heheh with Gradle you can do a lot of customizations which is good in some cases and bad on others.

MatousJobanek:
Ah, dammit - there is some plugin for it: https://www.smartics.eu/alias-maven-plugin/index.html
and also some bloody `defaultGoal` element: https://maven.apache.org/pom.html#BaseBuild_Element

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SkipInstallationChecker.java
```diff
@@ -0,0 +1,52 @@
...
+
+        return reason != null;
+    }
+
+    String getReason(){
+        return reason;
+    }
+
+    private boolean isGoalSpecification(String goal) {
+        return goal.indexOf(':') >= 0;
```
lordofthejars: maybe to be replaced by contains?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SkipInstallationChecker.java
```diff
@@ -0,0 +1,52 @@
...
+class SkipInstallationChecker {
+
+    private static final List<String> EXPECTED_GOALS = Arrays.asList(
+        new String[] {"test", "prepare-package", "package", "pre-integration-test", "integration-test",
+            "post-integration-test", "verify", "install", "deploy", "pre-site", "site", "post-site", "site-deploy"});
+    private final List<String> goals;
+    private String reason;
+
+    SkipInstallationChecker(List<String> goals) {
+
```
lordofthejars: White space?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SkipInstallationChecker.java
MatousJobanek: Yeah, the very same thing came to my mind when I was coding it. As I have written above, this code was stolen from Maven internals: https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/lifecycle/internal/DefaultLifecycleTaskSegmentCalculator.java#L150-L153
No idea why they are doing this, but as I'm not the only one who considers this way of checking it as a weird, then I'll change it to `contains` ;-)

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SkipInstallationChecker.java
MatousJobanek: Yeah, because I love space. And white space in particular :-D
Good catch 👍

bartoszmajsak:
Maybe we should only handle the `clean` case to not have `target` folders and stick to what the goal is expected to perform? Handling "all the possible" test triggers sounds to me like a big effort with not that much of a benefit.

I love the refactorings btw!

MatousJobanek:
As for the "not creating `target`" thingy - this is also managed here, no matter what goal has been set:
https://github.com/arquillian/smart-testing/pull/233/files#diff-4dcbe87c4b69902e40c4da7a56ce9ebaR46
if the `target` doesn't exist, then the files are not moved there, so the directory is not created.
Btw, for example, also the `validate` goal doesn't create `target` directories.

As for the goals. it is not a big effort, it is pretty simple and transparent to the user - I'm checking if any of these [goals](https://github.com/arquillian/smart-testing/pull/233/files#diff-13be4179d9c6a70ac8c694c7ede8af4aR12) has been set, if not, then I log the information to the console:
```
[INFO] Smart Testing Extension - Smart Testing is disabled. Reason: None of the 
goals specified for the build will invoke the tests. Any of the following goals are 
expected: [test, prepare-package, package, pre-integration-test, integration-test, 
post-integration-test, verify, install, deploy, pre-site, site, post-site, site-deploy]
``` 

If you have any objections/proposals/concerns, I'm open for any changes or improvements. ;-)

bartoszmajsak:
I was more thinking about things like alias plugin - do we consider it a challenge?

Issue #15, opened 2017-05-23
Supporting JUnit 5


bartoszmajsak:
Surefire does not support JUnit 5 out of the box. In order to enable it junit5 team created their own provider which has to be added to `surefire` [plugin configuration](http://junit.org/junit5/docs/current/user-guide/#running-tests-build-maven). 

We have to improve our code handling real test providers (as ours just delegate) and have ftest project proving it all works nicely together. Entry point which is wiring all providers together can be found here https://github.com/arquillian/smart-testing/blob/1b30f1e70b77ce24e58aa4e75c6c35c215cfe170/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/ProviderList.java#L13-L18

PR #241, opened 2017-10-24
feat(#15): Add support for JUnit5

lordofthejars:


bartoszmajsak:
I'm reviewing this code, but since it's `Status: Work In Progress` could you make a checklist of what still needs to be done? Thanks!

junit-test-result-parser/pom.xml
```diff
@@ -18,6 +18,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+      <scope>test</scope>
```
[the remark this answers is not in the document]
lordofthejars: No it is not need, but in my way to try to find which artifact is adding junit4 as compile I decided to force everywhere. I am going to delete.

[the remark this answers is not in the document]
bartoszmajsak: That's what I thought :)

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/JUnit5ProviderInfo.java
```diff
@@ -0,0 +1,49 @@
...
+        return getJunitDepVersion() != null && isAnyJunit5() && junit5SurefirePlatformVersion != null;
+    }
+
+    public String getDepCoordinates() {
+        return "org.junit.platform:junit-platform-surefire-provider:" + junit5SurefirePlatformVersion;
+    }
+
+    private String retrieveJunit5SurefirePlatformVersion() {
+
+        final Path junit5PlatformVersionFile = new LocalStorage(Paths.get("").toFile()).duringExecution()
```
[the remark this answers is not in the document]
lordofthejars: Already discussed in bjn

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractorJUnit5Test.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final String junitVersion;
+
+    public LoaderVersionExtractorJUnit5Test(String junitVersion) {
+        this.junitVersion = junitVersion;
+    }
+
+    @Parameterized.Parameters
+    public static Collection<String> data() {
+        return Arrays.asList("5.0.0");
```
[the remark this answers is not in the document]
lordofthejars: Already discussed in bjn

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractorSurefireJUnit5Test.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final String surefirejunitVersion;
+
+    public LoaderVersionExtractorSurefireJUnit5Test(String junitSurefireVersion) {
+        this.surefirejunitVersion = junitSurefireVersion;
+    }
+
+    @Parameterized.Parameters
+    public static Collection<String> data() {
+        return Arrays.asList("1.0.0");
```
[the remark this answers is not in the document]
lordofthejars: Already discussed in bjn

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java
```diff
@@ -108,6 +108,23 @@ public void should_fail_when_autocorrect_repeats_the_strategy()
...
     }
 
+    @Test
+    public void should_remove_junit5_platform_dependency() {
+        // given
+        Model model = prepareModelWithSurefirePlugin("2.19.1");
+        final Configuration conf = configureWithAutocorrect("nwe");
+        final DependencyResolver dependencyResolver = new DependencyResolver(conf);
+        final Plugin plugin = model.getBuild().getPlugins().get(0);
+        plugin.addDependency(new DependencyResolver.JUnit5SurefireProviderDependency());
```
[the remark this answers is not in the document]
lordofthejars: I understand your point but, it is not so easy because in some tests you need both `model` and `dependencyResolver`. One possible solution would be create a method with so many parameters so all possibilities can be fit with this method call.

The other option would be create some classes using builder pattern, but my question is is it worth for just 4 lines when the code is duplicated in just 3 test methods to do this?

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java
```diff
@@ -108,6 +108,23 @@ public void should_fail_when_autocorrect_repeats_the_strategy()
 
     }
 
+    @Test
+    public void should_remove_junit5_platform_dependency() {
+        // given
+        Model model = prepareModelWithSurefirePlugin("2.19.1");
+        final Configuration conf = configureWithAutocorrect("nwe");
```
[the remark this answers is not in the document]
lordofthejars: fixed.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/junit5/LocalChangesTestsSelectionExecutionFunctionalTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import static org.arquillian.smart.testing.ftest.testbed.configuration.Mode.SELECTING;
+import static org.arquillian.smart.testing.ftest.testbed.configuration.Strategy.CHANGED;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class LocalChangesTestsSelectionExecutionFunctionalTest {
+
+    @ClassRule
+    public static final GitClone GIT_CLONE = new GitClone("https://github.com/lordofthejars/junit5-demo.git");
+
```
dipak-pawar: Are we going to use the same repository or we are going to transfer ownership to arquillian org?

lordofthejars: Yes, I am not sure also how we should named in arquillian, because it is not really a dogfooding project. @bartoszmajsak WDYT about this?

My point now is that this is a really small project just to check that the file creation and read is done correctly, which is enough in case of JUnit5. So maybe call it `smart-testing-junit5-demo` in arquillian organization is enough.

bartoszmajsak: Yes, this is rather a demo project which we can transfer to the arq organization. We only do thin smoke test to be sure integration is fine, as the core logic which are tests are focusing on are not the test execution (so junit4 or junit5), but selection logic so our provider which delegates to aforementioned runners.

Let's move it as @lordofthejars suggests.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -39,8 +44,25 @@ void configureTestRunner(Model model) {
 
             dependencyResolver.addRequiredDependencies(model);
 
+
+            final LocalStorage localStorage = new LocalStorage(Paths.get("").toFile());
```
MatousJobanek: location

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -39,8 +44,25 @@ void configureTestRunner(Model model) {
...
-                .forEach(dependencyResolver::addAsPluginDependency);
+                .forEach(plugin -> {
+                    dependencyResolver.addAsPluginDependency(plugin);
+
+                    final Optional<Dependency> dependency = dependencyResolver.findJUnit5PlatformDependency(plugin);
+                    dependency.ifPresent(d -> {
+                        try {
+                            localStorage.duringExecution()
+                                .temporary()
+                                .file("junit5PlatformVersion")
```
MatousJobanek: constant

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -39,8 +44,25 @@ void configureTestRunner(Model model) {
 
             dependencyResolver.addRequiredDependencies(model);
 
+
+            final LocalStorage localStorage = new LocalStorage(Paths.get("").toFile());
             effectiveTestRunnerPluginConfigurations
-                .forEach(dependencyResolver::addAsPluginDependency);
+                .forEach(plugin -> {
+                    dependencyResolver.addAsPluginDependency(plugin);
```
MatousJobanek: what if there are different versions for each plugin

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java
```diff
@@ -75,7 +76,7 @@ public void should_autocorrect_strategy()
         Model model = prepareModelWithSurefirePlugin("2.20");
         final Configuration conf = configureWithAutocorrect("nwe");
         final DependencyResolver dependencyResolver = new DependencyResolver(conf);
-        model.addDependency(new DependencyResolver.SurefireApiDependency("2.19.1"));
+        model.addDependency(new DependencyResolver.SurefireApiDependency("2.20"));
```
MatousJobanek: why this change?

strategies/affected/pom.xml
```diff
@@ -32,7 +32,6 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <scope>test</scope>
```
MatousJobanek: why is the scope removed?

lordofthejars: Because it is inherited

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java
lordofthejars: Well it is a mocked instance, and you can change it without any worries, it works with any number there.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
lordofthejars: what do you mean about different versions of each plugin here, if I understand you correctly you mean what's happening if you have different versions of junit5 in the same project, which I think it is something really strange if this is what I think.

lordofthejars: Also notice that this plugin is added from a foreach which is all plugins `effectiveTestRunnerPlugin`

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java
MatousJobanek: so, again, why you have changed it? is there any reason for it?

strategies/affected/pom.xml
MatousJobanek: that you can say about the rest of the test dependencies... I would prefer to have everything specified there so I don't have to go to the parent pom checking what is specified there...

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/JUnit5ProviderInfo.java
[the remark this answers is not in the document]
MatousJobanek: that should be taken from this method: https://github.com/arquillian/smart-testing/blob/master/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java#L88

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java
lordofthejars: Well because the model is prepared with 2.20, so to make it more real the dependency should be 2.20 as well isn't it?See line 76

strategies/affected/pom.xml
lordofthejars: Originally there were not there the scope. Then I added for testing one thing but I committed by error, and then @bartoszmajsak (And I agree with him) asked me to remove again.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractor.java
```diff
@@ -42,6 +47,14 @@ public static String getJunitVersion() {
         return getVersionFromClassLoader(LIBRARY_JUNIT, currentThread().getContextClassLoader());
     }
 
+    public static String getJUnit5Version() {
+        return getVersionFromClassLoader(LIBRARY_JUNIT_5, currentThread().getContextClassLoader());
+    }
+
+    public static String getSurefireJUnit5Version() {
+        return getVersionFromClassLoader(LIBRARY_SUREFIRE_JUNIT_5, currentThread().getContextClassLoader());
+    }
```
MatousJobanek: Do we really need these methods? To simulate the Surefire's behavior correctly, if the JUnit 5 provider is present, then we should delegate to this provider, no matter what is on the classpath...

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/JUnit5ProviderInfo.java
```diff
@@ -0,0 +1,49 @@
...
+        super(LoaderVersionExtractor.getJUnit5Version());
+        junit5SurefirePlatformVersion = retrieveJunit5SurefirePlatformVersion();
+    }
+
+    public String getProviderClassName() {
+        return "org.junit.platform.surefire.provider.JUnitPlatformProvider";
+    }
+
+    public boolean isApplicable() {
+        return getJunitDepVersion() != null && isAnyJunit5() && junit5SurefirePlatformVersion != null;
```
MatousJobanek: yeah, here we should just check if the `junit5SurefirePlatformVersion` is not null

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractorJUnit5Test.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Collection;
+import org.jboss.shrinkwrap.resolver.api.maven.Maven;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(Parameterized.class)
+public class LoaderVersionExtractorJUnit5Test {
```
MatousJobanek: this class seems to be very similar to the https://github.com/arquillian/smart-testing/blob/master/surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractorJUnitTest.java

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractorSurefireJUnit5Test.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Collection;
+import org.jboss.shrinkwrap.resolver.api.maven.Maven;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(Parameterized.class)
+public class LoaderVersionExtractorSurefireJUnit5Test {
```
MatousJobanek: same as the previous one

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
MatousJobanek: I meant a different version of JUnit5 provider for surefire and failsafe...

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/JUnit5ProviderInfo.java
lordofthejars: This is what I am doing.

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractorJUnit5Test.java
lordofthejars: This test probably needs to be removed.

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/dependencies/SurefireApiDependencyTest.java
MatousJobanek: Yeah, I missed that line :-). Thx
anyway, I'm not sure if adding the surefire dependency is really necessary here... But that can be part of another PR ;-)

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractorSurefireJUnit5Test.java
lordofthejars: If you check the project before this PR you'll notice that each component we detect has its own LoaderVersion test class, one for TestNG, other for JUnit component and so on. So I am just following the same pattern as before. We can change, but in this case this was implemented following the same approach.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractor.java
lordofthejars: In fact needs to be removed.

strategies/affected/pom.xml
MatousJobanek: originally when? looking at the state before your commit, it was there ;-) https://github.com/arquillian/smart-testing/blob/1498d4051942f3b8725712bcd5fbe547cbd2e82b/strategies/affected/pom.xml#L35

lordofthejars: check other pom.xmls files and check that it was not present

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/JUnit5ProviderInfo.java
MatousJobanek: I mean that it should check only if the version is present:  
` return junit5SurefirePlatformVersion != null;`
see this comment: https://github.com/arquillian/smart-testing/pull/241#discussion_r147354761

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
lordofthejars: I think this is a really corner case, I mean I thought about this possibility but adding junit5 surefire platform with one version and in the same project another junit5 platform version for failsafe I would say that it is really really strange.

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractorSurefireJUnit5Test.java
MatousJobanek: Yeah, right... It just looks weird that it is the very same test with one parameter. 
But I agree.

strategies/affected/pom.xml
MatousJobanek: Then, from my point of view, it should be there. Otherwise, you can remove the scope from all other test dependencies

MatousJobanek: We had this kind of conversation already and I do remember that we agreed on having it there, otherwise, it brings a lot of confusions...

lordofthejars: Something to discuss on Cabal I guess to know how to proceed in general.

lordofthejars: Yes I remember this but since I saw that test scope was not in any of the modules, then I decided to remove as well to follow the same convention. In fact if you notice the only one that it had before was `affected`

MatousJobanek: as for the state before your commit: 
`find ./ -name "pom.xml" | xargs grep "artifactId>junit<" -A 1`
```xml
./functional-tests/test-bed/pom.xml:      <artifactId>junit</artifactId>
./functional-tests/test-bed/pom.xml-      <scope>compile</scope>
--
./functional-tests/git-rules/pom.xml:      <artifactId>junit</artifactId>
./functional-tests/git-rules/pom.xml-      <scope>compile</scope>
--
./strategies/changed/pom.xml:      <artifactId>junit</artifactId>
./strategies/changed/pom.xml-      <scope>test</scope>
--
./strategies/failed/pom.xml:      <artifactId>junit</artifactId>
./strategies/failed/pom.xml-      <scope>test</scope>
--
./strategies/affected/pom.xml:      <artifactId>junit</artifactId>
./strategies/affected/pom.xml-      <scope>test</scope>
--
./surefire-provider/pom.xml:      <artifactId>junit</artifactId>
./surefire-provider/pom.xml-      <scope>test</scope>
--
./junit-test-result-parser/pom.xml:      <artifactId>junit</artifactId>
./junit-test-result-parser/pom.xml-    </dependency>
--
./mvn-extension/pom.xml:      <artifactId>junit</artifactId>
./mvn-extension/pom.xml-    </dependency>
--
./core/pom.xml:      <artifactId>junit</artifactId>
./core/pom.xml-    </dependency>
```
so, I see more cases where the scope was specified... still, if you wanted to keep it unified, you should have removed it from all other test dependencies, which means everyting that is in the parent pom file: https://github.com/arquillian/smart-testing/blob/1498d4051942f3b8725712bcd5fbe547cbd2e82b/pom.xml#L85-L114...

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
MatousJobanek: There is another case - consider having surefire with JUnit 5 and failsafe without it...

Issue #231, opened 2017-10-13
Strategies should use Configuration provided by SmartTesting API


MatousJobanek:
##### Issue Overview

All strategies and following implementation should use configuration provided by SmartTesting API and not load it using `Configuration.loadPrecalculated` method because this one is maven-extension specific and cannot be used in other integration cases (such as Che).

PR #243, opened 2017-10-26
Implements strategy configuration support in config file.

### LINKED ISSUE #207: Add support for configuring strategies separately
##### Issue Overview

We have great YAML config file: http://arquillian.org/smart-testing/#_configuration_file where we can configure the general information. Now, we should support also the possibility to configure each strategy separately (eg. [inclusions and exclusions for affected one](http://arquillian.org/smart-testing/#_strategies_2))

example:
```yaml
mode: ordering 
strategies: new, changed, affected 
debug: true 
report:
    enable: true 
affected:
    inclusion:
        - org.package.*
        - org.arquillian.package.*
    exclusions:
        - org.package.exclude.*
        - org.arquillian.package.exclude.*
    
```

dipak-pawar:
Now we are providing `strategy` configuration support in configuration file.

- `core` has interface `StrategyConfiguration` which needs to implement for a strategy  with it's required configuration options.
- Setting strategy configuration to configuration in `SmartTestingMavenConfigurer` after setting autocorrected strategies & adding it's dependency on classpath. 
- Added builder in test bed to configure this in `test-bed`

This is how strategy configuration will look like in `smart-testing.yml`
```yaml
strategiesConfiguration:
    affected:
      config: affected-configuration.properties
      transitivity: true 
      exclusions: 
         - org.package.*
         - org.arquillian.package.*
      inclusions:
         - org.package.exclude.*
         - org.arquillian.package.exclude.*
```

Fixes #207


bartoszmajsak:
I thought we use labels for issues only. We can discuss priority though if there is an urgent bug to fix and release, but on the other hand, this is already defined in the issue itself.

dipak-pawar:
Thanks for information @bartoszmajsak. I will remove labels.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -137,6 +152,24 @@ public void setCustomStrategies(String[] customStrategies) {
...
+        new JavaSPILoader().all(StrategyConfiguration.class,
+            strategyConfiguration -> Arrays.asList(strategies).contains(strategyConfiguration.name()))
+            .forEach(strategyConfiguration -> {
+                final Class<StrategyConfiguration> aClass1 =
+                    (Class<StrategyConfiguration>) strategyConfiguration.getClass();
+                convertedList.add(mapToObject(aClass1, new HashMap<>(0)));
+            });
+
+        if (convertedList.isEmpty()) {
+            return null;
```
bartoszmajsak: Why not return an empty list instead? What extra information `null` gives us?

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -137,6 +152,24 @@ public void setCustomStrategies(String[] customStrategies) {
         return configItems;
     }
 
+    public List<StrategyConfiguration> strategyConfigurations(String[] strategies) {
+        List<StrategyConfiguration> convertedList = new ArrayList<>();
+
+        new JavaSPILoader().all(StrategyConfiguration.class,
```
lordofthejars: I am rising a question, instead of creating an ew SPI interface, couldn't we create a new method at `TestExecutionPlannerFactory` which returns this object? So for the implementators needs only to create one SPI

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java
```diff
@@ -20,9 +25,23 @@
...
-        readFile(System.getProperty(SMART_TESTING_AFFECTED_CONFIG));
+    AffectedRunnerProperties(File rootDirectory) {
+        final Configuration configuration = Configuration.loadPrecalculated(rootDirectory);
+        final List<StrategyConfiguration> strategyConfigurations = configuration.getStrategiesConfiguration();
+        if (strategyConfigurations == null) {
+            readFile(null);
+            return;
+        }
+
+        final StrategyConfiguration affectedConfig = strategyConfigurations.stream()
```
lordofthejars: Maybe this logic could be moved inside `Configuration` object so you can do `configuration.getStrategyConfiguration("affected)"`

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
dipak-pawar: We can do that but in that case we don't have configuration for each implementation of `TestExecutionPlannerFactory`. I mean we have configuration only for `Affected`. In other cases it will be only null.

but in this way even for custom strategy with configuration option user can implement it with one spi.

lordofthejars: I agree with you, but what I see is that from the point of view of developer we can always return `null` or we can create an `Empty` implementation. What I see is that in this way for a developer it is just implementing a new method returning empty class or his own implementation. In the other case you need to start dealing with SPI thing. In my opinion the SPI is the strategy, and the strategy itself is the one which provides you all the information required, let's call it configuration parameters, method call to select tests to execute, ....

bartoszmajsak: This is very valid point - I think we should make it happen as part of this PR

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -137,6 +152,24 @@ public void setCustomStrategies(String[] customStrategies) {
...
+            strategyConfiguration -> Arrays.asList(strategies).contains(strategyConfiguration.name()))
+            .forEach(strategyConfiguration -> {
+                final Class<StrategyConfiguration> aClass1 =
```
MatousJobanek: do we have a better name for this variable?

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -137,6 +152,24 @@ public void setCustomStrategies(String[] customStrategies) {
         return configItems;
     }
 
+    public List<StrategyConfiguration> strategyConfigurations(String[] strategies) {
```
MatousJobanek: Is it possible to incorporate this method to the `load()` method - as another option? To be possible to call the Configuration load method only once in other integration cases (such as Che)

MatousJobanek: Also, I would propose changing the method name to `loadStrategyConfigurations` to express what it really does

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/DependencyResolver.java
```diff
@@ -68,6 +68,7 @@ private void addStrategies(Model model) {
                 + "for the list of available strategies.");
         }
         configuration.setStrategies(strategies);
+        configuration.setStrategiesConfiguration(configuration.strategyConfigurations(strategies));
```
MatousJobanek: This seems to be a bit hacky - calling on an instance of the configuration object a method and assign the result to the very same instance... it can be assigned directly inside of the method call...

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -158,6 +160,8 @@ private static Object convert(Method method, Object mappedValue) {
             return mappedValue;
         } else if (ConfigurationSection.class.isAssignableFrom(parameterType)) {
             return mapToObject((Class<ConfigurationSection>) parameterType, (Map<String, Object>) mappedValue);
+        } else if (parameterType.isAssignableFrom(String.class) && List.class.isAssignableFrom(mappedValue.getClass())) {
+            return ((List) mappedValue).stream().collect(Collectors.joining(", "));
```
MatousJobanek: could you shed more light on this case? which use-case are you covering by this?

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -223,6 +227,23 @@ private static Object handleList(Method method, Object mappedValue) {
                 convertedList.add((T) convertToType(parameterType, v));
             }
             return convertedList;
+        } else if (Map.class.isAssignableFrom(aClass) && StrategyConfiguration.class.isAssignableFrom(parameterType)) {
+            final Map<String, Object> objectMap = (Map<String, Object>) mappedValue;
+            List<T> convertedList = new ArrayList<>(objectMap.size());
+            objectMap.forEach((k, v) -> {
+                final JavaSPILoader javaSPILoader = new JavaSPILoader();
+                final Optional<StrategyConfiguration> strategyConfiguration =
+                    javaSPILoader.onlyOne(StrategyConfiguration.class, config -> config.name().equals(k));
```
MatousJobanek: I'm a bit confused here - this seems to be a duplication of the logic in `Configuration.strategyConfigurations`.
Maybe appropriate naming of the variables & refactoring would help...

core/src/test/resources/configuration/smart-testing.yml
```diff
@@ -13,3 +13,13 @@ autocorrect: true #<!--9-->
 customStrategies:  #<!--10-->
     - smart.testing.strategy.cool=org.arquillian.smart.testing:strategy-cool:1.0.0
     - smart.testing.strategy.experimental=org.arquillian.smart.testing:strategy-experimental:1.0.0
+strategiesConfiguration: #<!--11-->
+    affected:
+      config: affected-configuration.properties #<!--12-->
```
MatousJobanek: I think that the way of using affected related config file should be replaced by this new implementation - in other words, there should be only one config file for everything.

docs/configfile.adoc
```diff
@@ -21,6 +21,11 @@ copyToClipboard:config-file[]
 <8> This sets last commit sha or `HEAD` notation for inspecting changes.
 <9> This defines if smart testing should auto correct misspelled strategies.
 <10> This defines the pair key/value of custom strategies as list.
+<11> This defines list of strategies with it's configuration that you want to apply while applying strategy.
```
MatousJobanek: > list of strategies

list of strategies is in `strategies` field. Here is a list of strategy related configurations

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java
```diff
@@ -20,9 +25,23 @@
     static final String EXCLUSIONS = "exclusions";
 
     private final Properties properties = new Properties();
+    private AffectedConfiguration affectedConfiguration;
 
-    AffectedRunnerProperties(){
-        readFile(System.getProperty(SMART_TESTING_AFFECTED_CONFIG));
+    AffectedRunnerProperties(File rootDirectory) {
+        final Configuration configuration = Configuration.loadPrecalculated(rootDirectory);
```
MatousJobanek: you shouldn't load the precalculated configuration here - you should use that one that was set by ST API in the surefire provider.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java
```diff
@@ -20,9 +25,23 @@
...
+        final List<StrategyConfiguration> strategyConfigurations = configuration.getStrategiesConfiguration();
+        if (strategyConfigurations == null) {
+            readFile(null);
+            return;
+        }
+
+        final StrategyConfiguration affectedConfig = strategyConfigurations.stream()
+            .filter(strategyConfiguration -> "affected".equals(strategyConfiguration.name()))
```
MatousJobanek: > "affected"

I would use a constant or method that returns the name of the strategy

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
dipak-pawar: no specific reason.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java
dipak-pawar: We are using it only one. So I haven't used any constant.

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
dipak-pawar: We need to set list of string to string for `inclusion` & `exclusion` from `affected` configuration.

you can find respective test it [here](https://github.com/arquillian/smart-testing/pull/243/files#diff-4b3896eea6384f8873440fb7836cbfa0R86).

```yaml
strategiesConfiguration:
    affected:
      config: affected-configuration.properties 
      transitivity: true 
      exclusions:
         - org.package.*
         - org.arquillian.package.*
      inclusions:
         - org.package.exclude.*
         - org.arquillian.package.exclude.*
```

OR

```yaml
strategiesConfiguration:
    affected:
      config: affected-configuration.properties 
      transitivity: true 
      exclusions: org.package.*,  org.arquillian.package.*
      inclusions: org.package.*, org.arquillian.package.exclude.*
```
Here in yaml configuration file we have user can provide `inclusion` & `exclusion` in comma separated string or list.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java
dipak-pawar: any good reason for it?

core/src/test/resources/configuration/smart-testing.yml
dipak-pawar: no currently it's not implemented this way. We are reading only `exclusions` & `inclusions` from config file. We are not reading `transitivity` in config file. may be something to improve later.

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
dipak-pawar: > Is it possible to incorporate this method to the load() method - as another option? To be possible to call the Configuration load method only once in other integration cases (such as Che)

if we move this method in `load()`, then we won't have strategy dependency added on classpath in case of mvn extension. So we won't get any any implementation of `strategyConfiguration`

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
MatousJobanek: wouldn't make more sense to keep it just as a list? The string has to be split anyway - to process it (I guess).

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
MatousJobanek: I meant to have a possibility to call `load(String[] strategies)` to load everything at once... both the current `load()` and `strategyConfigurations(String[] strategies)` would stay

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java
MatousJobanek: The reason is this issue I fixed some time ago: https://github.com/arquillian/smart-testing/issues/231

MatousJobanek: > Because we are doing the same in it 

no, we are not - we provide the instance of the configuration as a parameter: https://github.com/arquillian/smart-testing/blob/1498d4051942f3b8725712bcd5fbe547cbd2e82b/surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java#L81
and now imagine that I need to call the ST API from a different directory than the project root directory is.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java
MatousJobanek: second usage: https://github.com/arquillian/smart-testing/blob/master/strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedChangesDetectorFactory.java#L13

core/src/test/resources/configuration/smart-testing.yml
MatousJobanek: what do you mean by the fact that the `transitivity` is not read in a config file? Which config file? I can see that everything is loaded in this new implementation (including `transitivity` parameter): https://github.com/arquillian/smart-testing/pull/243/files#diff-b222e33537ad5f62b9ddccd10f1d89d9R19

The purpose of this PR is to have one configuration file for both the general configuration and all strategy related configurations. There shouldn't be any additional configuration file for the strategies...

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
MatousJobanek: Now I understand why you did this - you were trying to keep the same way as it was done previously - in the `affected-configuration.properties` but that was properties file, not YAML file.

core/src/test/resources/configuration/smart-testing.yml
dipak-pawar: you can define eveything in configuration file using current approach. But along with this you can define `inclusions` & `exclusions` using properties file.

Also if we remove this `properties` file support & ask user to give all `inclusions` & `exclusions` in yaml file, then user has to define all includes & excludes in config file, then while overwriting inclusions & exclusions using system property then user has to write whole list of inclusions & exclusions. That's why it's good to provide support for both.

MatousJobanek: > you can define eveything in configuration file using current approach. But along with this you can define inclusions & exclusions using properties file.

I would really welcome one consistent way of defining the properties. This means a usage of one config file or the system properties. Not several configuration files. We had this discussion before and IIRC we agreed on having it in one config file. Am I right @bartoszmajsak @lordofthejars ?

> Also if we remove this properties file support & ask user to give all inclusions & exclusions in yaml file, then user has to define all includes & excludes in config file, 

Yes, this is what we want - to have everything in one place. It is not expected that the includes and excludes will be somehow long...

> then while overwriting inclusions & exclusions using system property then user has to write whole list of inclusions & exclusions.

But this will be same also for another external property file - you need to either change the properties file or define everything using system properties...
BTW. introducing another properties file brings another complication - what takes the precedence? Values specified
* in a properties file
* in YAML config file
* in a properties file that is set by a system property
* by a system property. 

I would vote for keeping it consistent and simple - having one config file.

dipak-pawar: Ideally we'll be taking everything specified in properties file I think we are not overwriting anything in properties file.

then whatever we are defining in yaml configuration file we are combining it with properties file.

However we are overwriting `inclusions` & `exclusion` defined in yaml configuration file using system properties.

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraphUsingPropertyTest.java
```diff
@@ -0,0 +1,121 @@
...
+        // given
+        final AffectedConfiguration affectedConfiguration = new AffectedConfiguration();
+        affectedConfiguration.setTransitivity(false);
+
+        final Configuration configuration = Configuration.load();
+        configuration.setStrategiesConfiguration(Collections.singletonList(affectedConfiguration));
+        configuration.dump(temporaryFolder.getRoot());
+
+        final ClassDependenciesGraph
+            classDependenciesGraph = new ClassDependenciesGraph(new EndingWithTestTestVerifier(), Configuration.load());
```
MatousJobanek: I'm slightly confused here - on line 69, you dump the config to the temporary folder, but here you again read the default values and not from the temp folder...? can be applied for all the following test methods. 
In my opinion, the configuration shouldn't be dumped but passed as a parameter to the `ClassDependenciesGraph`
This opens an interesting question - do you really test anything here if all tests are green with wrong configuration?

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraphUsingPropertyTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+        final String testLocation = ATest.class.getResource("ATest.class").getPath();
+        final String testLocation2 = BTest.class.getResource("BTest.class").getPath();
+        final String testLocation3 = CTest.class.getResource("CTest.class").getPath();
+        classDependenciesGraph.buildTestDependencyGraph(Arrays.asList(new File(testLocation), new File(testLocation2),
+            new File(testLocation3)));
```
MatousJobanek: this part is same for all test methods - could be moved to a separated method

strategies/affected/src/test/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraphUsingPropertyTest.java
```diff
@@ -0,0 +1,121 @@
...
+        // then
+        assertThat(testsDependingOn)
+            .isEmpty();
+    }
+
+    @Test
+    public void should_include_only_imports_if_property_set() {
+        // given
+        final AffectedConfiguration affectedConfiguration = new AffectedConfiguration();
+        affectedConfiguration.setInclusions("org.arquillian.smart.testing.strategies.affected.fakeproject.main.A");
```
MatousJobanek: I've been trying to break this test anyhow - putting non-existing FQNs here, and the test is always green

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/FailedTestsDetectorFactory.java
```diff
@@ -23,5 +24,9 @@ public TestExecutionPlanner create(File projectDir, TestVerifier verifier, Confi
         return new FailedTestsDetector();
     } // <2>
 
+    @Override
+    public StrategyConfiguration strategyConfiguration() {
+        return null;
```
MatousJobanek: I'm not sure if this is the best thing for including into the documentation

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/AffectedRunnerProperties.java
```diff
@@ -1,80 +1,45 @@
...
+import org.arquillian.smart.testing.configuration.Configuration;
+import org.arquillian.smart.testing.spi.StrategyConfiguration;
 
 class AffectedRunnerProperties {
 
-    private static final String SMART_TESTING_AFFECTED_CONFIG = "smart.testing.affected.config";
-
     static final String SMART_TESTING_AFFECTED_TRANSITIVITY = "smart.testing.affected.transitivity";
+    @SuppressWarnings("unused")
     static final String DEFAULT_SMART_TESTING_AFFECTED_TRANSITIVITY_VALUE = "true";
```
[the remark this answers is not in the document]
dipak-pawar: Used in documentation

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/FailedTestsDetectorFactory.java
bartoszmajsak: In general returning `null` is a bit odd, isn't it?

dipak-pawar: then maybe I will create Empty Configuration class

core/src/test/java/org/arquillian/smart/testing/configuration/ObjectMapperTest.java
```diff
@@ -83,6 +83,18 @@ public void should_set_string_to_object() {
     }
 
     @Test
+    public void should_set_list_of_string_as_string_to_object() {
```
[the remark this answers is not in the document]
dipak-pawar: @MatousJobanek We are supporting both formats with this. 

```yaml
strategies: affected
strategiesConfiguration:
  affected:
    transitivity: true
    exclusions:
       - a
       - b
       - c
    inclusions:
       - d
       - e
```

OR 

```
strategies: affected
strategiesConfiguration:
  affected:
    transitivity: true
    exclusions: a, b, c
    inclusions: d, e
```

Both are valid `yaml` formats

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/FailedTestsDetectorFactory.java
bartoszmajsak: What would that solve? Why are we having this interface with only `name` method in it?

dipak-pawar: > What would that solve?

Currently we don't have any configuration for `new`, `failed`, `affected` strategy for which we are returning null. May be instead of that Either create empty configuration so we can filter it out in `loadStrategyConfiguration()`

> Why are we having this interface with only name method in it?

If any strategy has configuration like affected (transitivity, inclusions, exclusions) then it can be defined by implementing `StrategyConfiguration`. In case of `affected` strategy we are defining [AffectedConfiguration](https://github.com/arquillian/smart-testing/pull/243/files#diff-94774a660c8fa42f85c66691e6db1fb8) which we are [loading](https://github.com/arquillian/smart-testing/pull/243/files#diff-348861c21efd28f5cc99117b2c67302aR161) after runtime dependency of strategy is added to classpath using `name()` method of `StrategyConfiguration`.

dipak-pawar: Or maybe create configuration for each strategy with it's name having empty list of `ConfigurationItem` something like

```java
return new StrategyConfiguration() {
            @Override
            public String name() {
                return "failed";
            }

            @Override
            public List<ConfigurationItem> registerConfigurationItems() {
                return new ArrayList<>();
            }
        };
```

PR #246, opened 2017-10-27
fix(#245): Allows different versions of JUnit in same project

### LINKED ISSUE #245: JUnit5 and JUnit4 cannot be mixed
Currently a project can only work in JUnit 4 or JUnit 5 but cannot work in cases where one module of the project uses JUnit 4 and other in JUnit 5 in forked surefire configuration.

lordofthejars:

Adds creation of different version files depending if junit5 is defined in surefire or failsafe, also taking into consideration defining different versions of JUnit through different modules.

- Create one file storing JUnit5 version in each module directory and named _failsafe_ or _surefire_.

Fixes #245


core/src/main/java/org/arquillian/smart/testing/hub/storage/local/LocalStorage.java
```diff
@@ -7,6 +7,8 @@
  */
 public class LocalStorage {
 
+    public static String JUNIT_5_PLATFORM_VERSION = "junit5PlatformVersion";
```
bartoszmajsak: Minor thing - I'm not sure this should belong here - it's not related to storage at all. Maybe `LoaderVersionExtractor` would be a better place?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/JUnit5ProviderInfo.java
```diff
@@ -11,28 +12,31 @@
 
     private String junit5SurefirePlatformVersion;
 
-    public JUnit5ProviderInfo() {
-        super(LoaderVersionExtractor.getJUnit5Version());
-        junit5SurefirePlatformVersion = retrieveJunit5SurefirePlatformVersion();
+    public JUnit5ProviderInfo(File projectDir) {
+        super(null);
```
bartoszmajsak: Any particular reason why extend from `JUnitProviderInfo` instead of implementing `ProviderInfo`? This `null` looks weird.

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractorSurefireJUnit5Test.java
```diff
@@ -1,50 +0,0 @@
...
-import java.util.Collection;
-import org.jboss.shrinkwrap.resolver.api.maven.Maven;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
-
-import static org.assertj.core.api.Assertions.assertThat;
-
-@RunWith(Parameterized.class)
-public class LoaderVersionExtractorSurefireJUnit5Test {
```
bartoszmajsak: Those two tests were removed in favor of `JUnit5ProviderInfoTest`?

core/src/main/java/org/arquillian/smart/testing/hub/storage/local/LocalStorage.java
MatousJobanek: The reason why we put it here is that we need to have it available on both sides - maven extension and surefire
We had the same discussion - where we should put it and we decided for this place

bartoszmajsak: Don't we have `LoaderVersionExtractor` available for both side though?

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/info/JUnit5ProviderInfo.java
MatousJobanek: good point - this is kind of zombie

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/LoaderVersionExtractorSurefireJUnit5Test.java
MatousJobanek: nope, there have been removed methods that were getting the version so the tests are not needed https://github.com/arquillian/smart-testing/pull/246/files#diff-f265edf8b316db4ccfc2829915f4af85L46

bartoszmajsak:
Thanks guys!

PR #247, opened 2017-10-31
Looking for configuration file recursively.

### LINKED ISSUE #235: Look for configuration file recursively in the parent directories.
##### Issue Overview

In the case when a configuration file is not in the directory we are running the build from, we should browse all parent directories looking for any configuration file we could load. 

consider this structure:
```
parent
 - config
   -- api
     --- pom.xml
   -- impl-base
      --- pom.xml
   -- spi
      --- pom.xml
 - smart-testing.yml
```
and running the build from `config/impl-base`, then the `smart-testing.yml` located in `parent` directory should be loaded.

If it is possible, take into consideration also the structure of the Maven project - in other words, having the `parent` directory as the "highest one" then  stop looking there.

dipak-pawar:
Now looking for `smart-testing.yml` file recursively in parent directory till multi-module project's root directory.

Fixes #235

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
+++ b/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
@@ -107,20 +107,20 @@ public class ConfigurationLoader {
     }
 
     private static Path getConfigurationFilePath(File[] files) {
-        if (files.length == 1) {
-            final File configFile = files[0];
-            logger.info("Using configuration from " + configFile.getName());
-            return configFile.toPath();
+        if (files.length == 2) {
+            logger.warn(
+                "Found multiple config files with supported names: " + SMART_TESTING_YAML + ", " + SMART_TESTING_YML);
         }
 
-        logger.warn("Found multiple config files with supported names: " + SMART_TESTING_YAML + ", " + SMART_TESTING_YML);
-        logger.warn("Using configuration from " + SMART_TESTING_YML);
-
-        return Arrays.stream(files)
-            .filter(file -> file.getName().equals(SMART_TESTING_YML))
+        final Path configFilePath = Arrays.stream(files)
+            .filter(file -> file.getName().equals(SMART_TESTING_YML) || file.getName().equals(SMART_TESTING_YAML))
             .map(File::toPath)
-            .findFirst()
+            .findAny()
             .get();
+
+        logger.info("Using configuration from " + configFilePath);
+
+        return configFilePath;
     }
 
     private static Configuration parseConfiguration(Map<String, Object> yamlConfiguration) {
--- a/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
+++ b/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
@@ -113,9 +113,14 @@ public class ConfigurationLoader {
         }
 
         final Path configFilePath = Arrays.stream(files)
-            .filter(file -> file.getName().equals(SMART_TESTING_YML) || file.getName().equals(SMART_TESTING_YAML))
+            .filter(file -> {
+                if (files.length == 2) {
+                    return file.getName().equals(SMART_TESTING_YML);
+                }
+                return file.getName().equals(SMART_TESTING_YAML) || file.getName().equals(SMART_TESTING_YML);
+            })
             .map(File::toPath)
-            .findAny()
+            .findFirst()
             .get();
 
         logger.info("Using configuration from " + configFilePath);
--- a/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
+++ b/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
@@ -25,7 +25,7 @@ public class LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest 
     public final TestBed testBed = new TestBed(GIT_CLONE);
 
     @Test
-    public void should_load_configuration_from_parent_dir_if_not_present_in_current_execution_dir() throws Exception {
+    public void should_only_execute_new_tests_related_to_single_local_change() throws Exception {
         // given
         final Project project = testBed.getProject();
 
@@ -38,7 +38,8 @@ public class LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest 
         final Collection<TestResult> expectedTestResults = project
             .applyAsLocalChanges("Adds new unit test");
 
-        final TestResults actualTestResults = project.build(Paths.get("config","impl-base")).run();
+        // when
+        final TestResults actualTestResults = project.build().run();
 
         // then
         assertThat(actualTestResults.accumulatedPerTestClass()).containsAll(expectedTestResults).hasSameSizeAs(expectedTestResults);
@@ -68,5 +69,26 @@ public class LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest 
         // then
         assertThat(actualTestResults.accumulatedPerTestClass()).containsAll(expectedTestResults).hasSameSizeAs(expectedTestResults);
 
+    }
+
+    @Test
+    public void should_load_configuration_from_parent_dir_if_not_present_in_current_execution_dir() throws Exception {
+        // given
+        final Project project = testBed.getProject();
+
+        project.configureSmartTesting()
+                .executionOrder(NEW)
+                .inMode(SELECTING)
+                .createConfigFile()
+            .enable();
+
+        final Collection<TestResult> expectedTestResults = project
+            .applyAsLocalChanges("Enable both surefire and failsafe plugin with sample integration test");
+
+        // when
+        final TestResults actualTestResults = project.build(Paths.get("core")).run("clean", "verify");
+
+        // then
+        assertThat(actualTestResults.accumulatedPerTestClass()).containsAll(expectedTestResults).hasSameSizeAs(expectedTestResults);
     }
 }
--- a/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
+++ b/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
@@ -24,7 +24,7 @@ public class LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest 
     public final TestBed testBed = new TestBed(GIT_CLONE);
 
     @Test
-    public void should_only_execute_new_tests_related_to_single_local_change() throws Exception {
+    public void should_load_configuration_from_parent_dir_if_not_present_in_current_execution_dir() throws Exception {
         // given
         final Project project = testBed.getProject();
 
@@ -38,7 +38,7 @@ public class LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest 
             .applyAsLocalChanges("Adds new unit test");
 
         // when
-        final TestResults actualTestResults = project.build().run();
+        final TestResults actualTestResults = project.build("config/impl-base").run();
 
         // then
         assertThat(actualTestResults.accumulatedPerTestClass()).containsAll(expectedTestResults).hasSameSizeAs(expectedTestResults);
--- a/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
+++ b/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
@@ -46,32 +46,6 @@ public class LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest 
     }
 
     @Test
-    public void should_only_execute_new_tests_related_to_single_local_change_using_failsafe() {
-
-        // given
-        final Project project = testBed.getProject();
-
-        project.configureSmartTesting()
-                    .executionOrder(NEW)
-                    .inMode(SELECTING)
-                .createConfigFile()
-            .enable();
-
-        project
-            .applyAsCommits("Disable surefire and enable just failsafe plugin");
-
-        final Collection<TestResult> expectedTestResults = project
-            .applyAsLocalChanges("Adds new unit test");
-
-        // when
-        final TestResults actualTestResults = project.build().run("clean", "verify");
-
-        // then
-        assertThat(actualTestResults.accumulatedPerTestClass()).containsAll(expectedTestResults).hasSameSizeAs(expectedTestResults);
-
-    }
-
-    @Test
     public void should_load_configuration_from_parent_dir_if_not_present_in_current_execution_dir() throws Exception {
         // given
         final Project project = testBed.getProject();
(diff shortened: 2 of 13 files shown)


mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -76,7 +79,10 @@ private void loadConfigAndCheckIfInstallationShouldBeSkipped(MavenSession sessio
             return;
         }
 
-        File projectDirectory = session.getTopLevelProject().getModel().getProjectDirectory();
+        final File multiModuleProjectDirectory = session.getRequest().getMultiModuleProjectDirectory();
+        final File executionDir = session.getTopLevelProject().getModel().getProjectDirectory();
+        final File projectDirectory = findFirstDirectoryWithConfigFile(executionDir, multiModuleProjectDirectory);
```
MatousJobanek: When the configuration file is loaded this message is printed on the stdout:
`Smart Testing Extension - Using configuration from smart-testing.yaml`
Now, when the config file can be loaded from a different directory it would be good to have the full path to the file visible in the log message.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -173,4 +179,25 @@ private void purgeLocalStorageAndExportPom(MavenSession session) {
             }
         });
     }
+
+    private File findFirstDirectoryWithConfigFile(File projectDir, File multiModuleProjectDirectory) {
```
MatousJobanek: actually, it doesn't find the first directory with the config file in all cases - not in the case when the file is not present anywhere - then it returns the root directory. how about: `getFirstDirWithConfigOrRootProjectDir`
or just to explain it in the javadoc

MatousJobanek:
It would be great to have a test for it.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
bartoszmajsak: Full path or relative to the project folder (whichever is easier to implement for the time being), but I agree we should have it.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
bartoszmajsak: If we can have self-explanatory method name it's better than javadoc I believe.

bartoszmajsak:
We have few things to address here, so I will capture them in one place

- [ ] resolve merge conflict
- [ ] fix the build
- [ ] find better name for [`findFirstDirectoryWithConfigFile`](https://github.com/arquillian/smart-testing/pull/247#discussion_r148393117) method
- [ ] Provide better information about which config files are used https://github.com/arquillian/smart-testing/pull/247#discussion_r148392952
- [ ] Provide tests for this change

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
bartoszmajsak: But this made me thinking - shouldn't we have hierarchical lookup here? e.g. I might have a config file defining e.g. strategies, but the mode we have globally enabled on the root level. Do we handle such a case now @dipak-pawar?

dipak-pawar: No we don't handle such case.

dipak-pawar: > shouldn't we have hierarchical lookup here?

then we'll miss scenario where we have config file defined in root directory of project & user is trying to run test from sub-module directory. Then as in sub-module directory we don't have config file it'll take default values which is not correct.

bartoszmajsak: I thought that was the idea overall, so user can have only the relevant bits configured in the sub-module. But I understand current solution is all-or-nothing? 

> then we'll miss scenario where we have config file defined in root directory of project & user is trying to run test from sub-module directory. Then as in sub-module directory we don't have config file it'll take default values which is not correct.

I am not sure I understand your point clearly. The idea might be the same as with hierarchy of poms. But if that's not how it is implemented we have to make sure we document it properly and if we agree this is the right thing to do - create a task.

bartoszmajsak:
@dipak-pawar can you update the task list https://github.com/arquillian/smart-testing/pull/247/#issuecomment-341253657 ?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
bartoszmajsak: Will be discussed offline with @dipak-pawar.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
```diff
@@ -38,7 +38,7 @@ public void should_only_execute_new_tests_related_to_single_local_change() throw
             .applyAsLocalChanges("Adds new unit test");
 
         // when
-        final TestResults actualTestResults = project.build().run();
+        final TestResults actualTestResults = project.build("config/impl-base").run();
```
MatousJobanek: Are you sure that this will really verify the recursive lookup?
IIRC, this will still execute the build from the parent directory - which means that the maven extension will be invoked from the same directory where the configuration is placed as well. By specifying the submodules you are saying that you want to execute only this module of the whole project (setting `-pl` parameter).

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ConfigLookUp.java
```diff
@@ -0,0 +1,48 @@
...
+    private final File executionDir;
+
+    ConfigLookUp(MavenSession session) {
+        this.mavenExecutionRequest = session.getRequest();
+        this.executionDir = new File(session.getExecutionRootDirectory());
+    }
+
+    File getFirstDirWithConfigOrProjectRootDir() {
+        return getFirstDirWithConfigOrProjectRootDir(executionDir,
+            new File(mavenExecutionRequest.getSystemProperties().getProperty("env.MAVEN_PROJECTBASEDIR")));
```
MatousJobanek: What is the difference between your version and just getting the system enviroment: `System.getenv("MAVEN_PROJECTBASEDIR")` ?

MatousJobanek: btw, what was wrong with the original way of getting the parent directory: `session.getRequest().getMultiModuleProjectDirectory()`?

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/ConfigLookUpTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Before
+    public void configureMavenExecutionRequest() {
+        mavenExecutionRequest = new DefaultMavenExecutionRequest();
+        Properties properties = new Properties();
+        properties.put("env.MAVEN_PROJECTBASEDIR", temporaryFolder.getRoot().toString());
+        mavenExecutionRequest.setSystemProperties(properties);
+    }
+
+    @Test
+    public void should_return_first_dir_with_config_file() throws IOException {
```
MatousJobanek: you forgot to add the comments:
```java
// given
// when 
// then
```

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/ConfigLookUpTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
```
MatousJobanek: just a detail: better is not using the annotations (runners) and just use the `Mockito.mock*` methods

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/ConfigLookUpTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+        Assertions.assertThat(firstDirWithConfigOrProjectRootDir)
+            .isEqualTo(new File(rootPath + File.separator + "parent"));
+    }
+
+    @Test
+    public void should_return_parent_dir_if_config_file_is_not_present() throws IOException {
+        temporaryFolder.newFolder("parent", "config");
+        temporaryFolder.newFile(String.join(File.separator, "parent", "pom.xml"));
+        temporaryFolder.newFile(String.join(File.separator, "parent", "config", "pom.xml"));
```
MatousJobanek: Just a detail: I would probably use the `nio` API: `Paths.get("parent", "config", "pom.xml").toString()`
For me, it is just more natural to use files/paths related API when I want to create a path

mvn-extension/src/test/java/org/arquillian/smart/testing/mvn/ext/ConfigLookUpTest.java
lordofthejars: why no annotations? what is the problem?

MatousJobanek: I didn't say that there is any problem. 
There are just two benefits when the annotation is not used:
 * it is more portable (e.g. migration to JUnit5)
 * when it is needed, we can easily change/extend the test using an additional runner (e.g. ParametrizedRunner)

This is just my approach/opinion ;-)

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
dipak-pawar: You are right. To set `-f` or working directory option from maven we need to release new version of `shrinkwrap-resolver-impl-maven-embedded` as @hemanik already fixed issue with `setWorkingDir` & added functionality for `setAlternatePomFile` in https://github.com/shrinkwrap/resolver/pull/125

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ConfigLookUp.java
dipak-pawar: We can use `System.getenv("MAVEN_PROJECTBASEDIR")`. There is no difference.

We don't have `multiModuleProjectDirectory` field in maven version `3.2.5`. So `session.getRequest().getMultiModuleProjectDirectory()` will fail for maven version 3.2.5

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
MatousJobanek: the `-f` functionality is not really necessary here. you can change the path here https://github.com/arquillian/smart-testing/blob/master/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java#L63
But yeah, I can release it. I wanted to do it anyway this week...

MatousJobanek: Looking at the code - it won't change really anything - both calls: that one I referenced and the `setAlternatePomFile` one would be placed in the same method. So you don't need to wait for the release as it doesn't bring any benefit to your code.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/LocalChangesNewTestsSelectionExecutionWithConfigFileFunctionalTest.java
```diff
@@ -69,4 +70,25 @@ public void should_only_execute_new_tests_related_to_single_local_change_using_f
         assertThat(actualTestResults.accumulatedPerTestClass()).containsAll(expectedTestResults).hasSameSizeAs(expectedTestResults);
 
     }
+
+    @Test
+    public void should_load_configuration_from_parent_dir_if_not_present_in_current_execution_dir() throws Exception {
```
MatousJobanek: Looking at the [tag commit](https://github.com/arquillian/smart-testing-dogfood-repo/commit/7b519516fd5e14ecb76a0cd67783293169715f26) - this is can replace the previous [test method](https://github.com/arquillian/smart-testing/pull/247/files#diff-34d1e596686e34c6dfc22210f43e6ffdR49) as it tests almost the same thing - there is no need to add a new test-method here and increase the test execution time
So just remove the test method `should_only_execute_new_tests_related_to_single_local_change_using_failsafe`

MatousJobanek:
LGTM
Good job @dipak-pawar

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
```diff
@@ -107,20 +107,20 @@ static Configuration load(Path path) {
     }
 
     private static Path getConfigurationFilePath(File[] files) {
-        if (files.length == 1) {
-            final File configFile = files[0];
-            logger.info("Using configuration from " + configFile.getName());
-            return configFile.toPath();
+        if (files.length == 2) {
+            logger.warn(
+                "Found multiple config files with supported names: " + SMART_TESTING_YAML + ", " + SMART_TESTING_YML);
```
bartoszmajsak: Should we warn or fail here? Or maybe `warn`, but decide which one we use? Wondering if `findAny` always returns the same result.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ConfigLookUp.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static org.arquillian.smart.testing.configuration.ConfigurationLoader.SMART_TESTING_YAML;
+import static org.arquillian.smart.testing.configuration.ConfigurationLoader.SMART_TESTING_YML;
+
+class ConfigLookUp {
```
bartoszmajsak: Minor thing - I believe it's just `Lookup` without `U` uppercase.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/ConfigLookUp.java
```diff
@@ -0,0 +1,43 @@
...
+    File getFirstDirWithConfigOrProjectRootDir() {
+        return getFirstDirWithConfigOrProjectRootDir(executionRootDir, new File(System.getenv("MAVEN_PROJECTBASEDIR")));
+    }
+
+    private File getFirstDirWithConfigOrProjectRootDir(File projectDir, File multiModuleProjectDirectory) {
+        if (isSameFile(projectDir.toPath(), multiModuleProjectDirectory.toPath())) {
+            return projectDir;
+        }
+
+        if (projectDir.isDirectory() && new File(projectDir, "pom.xml").exists() &&
```
bartoszmajsak: General thought here:

- should this really belong to `mvn` part? I think configuration as such is a general concept.
- this implementation couples us with Maven - maybe we should split it up then. Not for now, but can result in the follow-up story

WDYT @MatousJobanek?

MatousJobanek: I was thinking about the same during our discussion today. 
IIRC, the reason why it was done it in this way is that ST is Maven extension, so it kind of reflects Maven behavior/project structure.
But, you are right about it that we probably couple with Maven too much.

PR #249, opened 2017-10-31
Fixes NPE for empty config file.

### LINKED ISSUE #248: Getting NPE if providing empty smart-testing.yml
*Steps To Reproduce*
* Create empty config file `smart-testing.yml` in project's root dir.
* Run test with `mvn test`

*Actual Result*
Getting NPE 

*Expected Result*
We should get appropriate msg if config file is empty

dipak-pawar:
Fixes #248

### DIFF
--- a/core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationTest.java
+++ b/core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationTest.java
@@ -1,6 +1,12 @@
 package org.arquillian.smart.testing.configuration;
 
+import java.io.IOException;
+import net.jcip.annotations.NotThreadSafe;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.contrib.java.lang.system.SystemErrRule;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.TemporaryFolder;
 
 import static org.arquillian.smart.testing.RunMode.ORDERING;
 import static org.arquillian.smart.testing.RunMode.SELECTING;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
```diff
@@ -155,9 +155,13 @@ public static Configuration load(File projectDir) {
             logger.info("Config file `" + SMART_TESTING_YAML + "` OR `" + SMART_TESTING_YML + "` is not found. "
                 + "Using system properties to load configuration for smart testing.");
         } else {
-            try (InputStream io = Files.newInputStream(getConfigurationFilePath(files))) {
+            final Path filePath = getConfigurationFilePath(files);
+            try (InputStream io = Files.newInputStream(filePath)) {
                 final Yaml yaml = new Yaml();
                 yamlConfiguration = yaml.load(io);
+                if (yamlConfiguration == null) {
```
lordofthejars: What's happening if file is empty but there are system properties set? Should we fail? Maybe the file is empty because someone else is the responsible of configuring.

core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationTest.java
```diff
@@ -15,6 +20,12 @@
 
 public class ConfigurationTest {
 
+    @Rule
+    public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+    @Rule
```
lordofthejars: Use assertJ assertions instead of this rule http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion

core/src/main/java/org/arquillian/smart/testing/configuration/Configuration.java
lordofthejars: I am just thinking loudly

MatousJobanek: I would expect that if the file is empty then we should not fail, just load the default values or those taken from system properties.
The only thing what we can do in this case is just logging the information/warn that the file is empty.

bartoszmajsak: I would log warning too to be honest. I guess that's what we do with empty `arquillian.xml` for example? Only fail when no sysprops are defined as well.

lordofthejars: Not sure about `arquillian.xml` but I guess that just warning is ok in this case. I agree with @MatousJobanek approach.

MatousJobanek:
I introduced a few refactoring changes and a test case + rebased with the master branch & resolved the conflicts

PR #254, opened 2017-11-02
Enables setting custom config file.

### LINKED ISSUE #242: Introduce parameter for pointing to a custom configuration file
##### Issue Overview

Currently, both config name and location of the configuration file is fixed. The name can be either `smart-testing.yml` or `smart-testing.yaml` and it should be located in the root of the directory the build is executed from. 
There should be introduced a parameter for using some custom YAML configuration file. The parameter can be eg.: `smart.testing.config` and should contain a path and point to the file, eg:
`-Dsmart.testing.config=../diferent/project/my-custom-config.yaml`
This means that the parameter should point to the file itself not only a directory. The parameter should take both the absolute path and relative path.

hemanik:
Enables use of custom configuration file (eg. file from a different project in multi-module project) using system property `smart.testing.config` for configuring ST.

- Adds system property to specify custom property for configuration file.
- Updates test bed for related changes and tests to verify this feature.

Fixes #242

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
+++ b/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
@@ -32,26 +32,24 @@ public class ConfigurationLoader {
     }
 
     public static Configuration load(File projectDir) {
-        Map<String, Object> yamlConfiguration;
+        File configFile = projectDir;
 
         final String customConfigFile = System.getProperty(SMART_TESTING_CONFIG);
         if (customConfigFile != null) {
-            yamlConfiguration = getConfigurationFromCustomConfigFile(customConfigFile);
-        } else {
-            yamlConfiguration = getConfigurationFromDefaultConfigFile(projectDir);
+            configFile = Paths.get(customConfigFile).toAbsolutePath().toFile();
+            if (!isCustomConfigFileValid(configFile)) {
+                configFile = projectDir;
+            }
         }
-
-        final Object strategiesConfiguration = yamlConfiguration.get("strategiesConfiguration");
-
-        final Configuration configuration = parseConfiguration(yamlConfiguration);
-        if (strategiesConfiguration != null) {
-            configuration.setStrategiesConfig((Map<String, Object>) strategiesConfiguration);
-        }
-
-        return configuration;
+        Map<String, Object> yamlConfiguration = readConfiguration(configFile);
+        return parseConfiguration(yamlConfiguration);
     }
 
-    private static Map<String, Object> getConfigurationFromDefaultConfigFile(File projectDir) {
+    private static Map<String, Object> readConfiguration(File projectDir) {
+        if (!projectDir.isDirectory()) {
+            return getConfigParametersFromFile(getConfigurationFilePath(projectDir));
+        }
+
         final File[] files =
             projectDir.listFiles((dir, name) -> name.equals(SMART_TESTING_YML) || name.equals(SMART_TESTING_YAML));
 
@@ -64,17 +62,6 @@ public class ConfigurationLoader {
                 + "Using system properties to load configuration for smart testing.");
         } else {
             return getConfigParametersFromFile(getConfigurationFilePath(files));
-        }
-        return Collections.EMPTY_MAP;
-    }
-
-    private static Map<String, Object> getConfigurationFromCustomConfigFile(String customConfigFile) {
-        final File file = Paths.get(customConfigFile).toAbsolutePath().toFile();
-        if (!file.exists()) {
-            logger.warn("Config file `" + file.getName() + "` is not found. "
-                + "Checking for config file `" + SMART_TESTING_YAML + "` OR `" + SMART_TESTING_YML);
-        } else {
-            return getConfigParametersFromFile(getConfigurationFilePath(file));
         }
         return Collections.EMPTY_MAP;
     }
@@ -160,6 +147,25 @@ public class ConfigurationLoader {
     }
 
     private static Configuration parseConfiguration(Map<String, Object> yamlConfiguration) {
-        return mapToObject(Configuration.class, yamlConfiguration);
+        final Object strategiesConfiguration = yamlConfiguration.get("strategiesConfiguration");
+        final Configuration configuration = mapToObject(Configuration.class, yamlConfiguration);
+        if (strategiesConfiguration != null) {
+            configuration.setStrategiesConfig((Map<String, Object>) strategiesConfiguration);
+        }
+        return configuration;
+    }
+
+    private static Boolean isCustomConfigFileValid(File customConfigFile) {
+        if (!customConfigFile.exists()) {
+            logger.warn("Config file `" + customConfigFile.getName() + "` is not found. "
+                + "Checking for config file `" + SMART_TESTING_YAML + "` OR `" + SMART_TESTING_YML);
+            return false;
+        }
+        if (customConfigFile.isDirectory()) {
+            logger.warn(customConfigFile.getName()
+                + " is a directory. Using the default configuration or please specify a `yaml` configuration file.");
+            return false;
+        }
+        return true;
     }
 }
--- a/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
+++ b/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
@@ -22,6 +22,8 @@ public class ConfigurationLoader {
 
     public static final String SMART_TESTING_YML = "smart-testing.yml";
     public static final String SMART_TESTING_YAML = "smart-testing.yaml";
+    public static final String SMART_TESTING_CONFIG = "smart.testing.config";
+
     private static final Logger logger = Log.getLogger();
 
     public static Configuration load() {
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
```diff
@@ -22,26 +22,35 @@
...
-        }
-
+        String customConfigFile = System.getProperty(SMART_TESTING_CONFIG);
         Map<String, Object> yamlConfiguration = new HashMap<>(0);
-        if (files.length == 0) {
-            logger.info("Config file `" + SMART_TESTING_YAML + "` OR `" + SMART_TESTING_YML + "` is not found. "
-                + "Using system properties to load configuration for smart testing.");
+
+        if (customConfigFile != null) {
+            File file = Paths.get(customConfigFile).toAbsolutePath().toFile();
```
lordofthejars: I'd add some protection of kind if file does not exists then log as warning explaining that the specified file cannot be found. Like it is done in else block `files.length()==0`

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectConfigurator.java
```diff
@@ -58,9 +61,12 @@ public ProjectConfigurator withConfiguration(Configuration configuration) {
         return this;
     }
 
-    public ProjectConfigurator createConfigFile() {
+    public ProjectConfigurator createConfigFile(String... customConfigFile) {
```
lordofthejars: if we are always taking the first one, why not creating two methods `createConfigFile` one with no args and the other with just one string

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectConfigurator.java
```diff
@@ -114,8 +126,20 @@ public String strategies() {
        return Arrays.stream(getStrategies()).map(Strategy::getName).collect(Collectors.joining(","));
     }
 
-    private void createConfigurationFile() {
-        final Path configFilePath = Paths.get(root.toString(), "smart-testing.yml");
+    private void createConfigurationFile(String... customConfigFile) {
```
lordofthejars: Same as `createConfigFile`

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
```diff
@@ -22,36 +22,60 @@
...
     public static Configuration load() {
         return load(Paths.get("").toAbsolutePath().toFile());
     }
 
     public static Configuration load(File projectDir) {
+        Map<String, Object> yamlConfiguration = new HashMap<>(0);
+
+        final String customConfigFile = System.getProperty(SMART_TESTING_CONFIG);
+        if (customConfigFile != null) {
+            yamlConfiguration = getConfigurationFromCustomConfigFile(customConfigFile, yamlConfiguration);
```
MatousJobanek: Providing the `yamlConfiguration` instance as a method parameter and then getting the very same instance as a return value of the method looks weird. Use one of the solutions, not both at once. The method name should be changed to reflect what it really does-in other words use:
either:
`loadConfigParametersFromCustomConfigFile(customConfigFile, yamlConfiguration);`
or:
`yamlConfiguration = getConfigParametersFromCustomConfigFile(customConfigFile);`

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
```diff
@@ -22,36 +22,60 @@
...
 
-        final Object strategiesConfiguration = yamlConfiguration.get("strategiesConfiguration");
-
-        final Configuration configuration = parseConfiguration(yamlConfiguration);
-        if (strategiesConfiguration != null) {
-            configuration.setStrategiesConfig((Map<String, Object>) strategiesConfiguration);
+    private static Map<String, Object> getConfigurationFromCustomConfigFile(String customConfigFile, Map<String, Object> yamlConfiguration) {
+        final File file = Paths.get(customConfigFile).toAbsolutePath().toFile();
+        if (!file.exists()) {
+            logger.info("Config file `" + file.getName() + "` is not found. "
```
MatousJobanek: if the file doesn't exist, then it should log `WARNING` and should load the configuration using the default way: `getConfigurationFromDefaultConfigFile`

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectConfigurator.java
```diff
@@ -119,6 +135,18 @@ private void createConfigurationFile() {
...
 
+    private void createConfigurationFile(String customConfigFile) {
+        Path configFilePath = Paths.get(root.toString(), customConfigFile);
+        if (!Files.exists(configFilePath)) {
+            try {
+                Files.createFile(configFilePath);
+            } catch (IOException e) {
+                throw new UncheckedIOException("Failed creating custom configuration file.", e);
+            }
+        }
```
MatousJobanek: I'm asking myself - what is the difference between this custom way using the `createConfigurationFile(String customConfigFile)` method and the default way using the `createConfigurationFile()` method -  there is no such a check of the presence of the custom file

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectConfigurator.java
```diff
@@ -103,9 +113,15 @@ public Project enable(Using usingInstallation) {
...
-                .createConfigFile();
+            if (customConfigFile != null) {
+                this.project.configureSmartTesting()
+                    .withConfiguration(configuration)
+                    .createConfigFile(customConfigFile);
+            } else {
+                this.project.configureSmartTesting()
+                    .withConfiguration(configuration)
+                    .createConfigFile();
+            }
```
MatousJobanek: Btw, if you moved the lines
[Paths.get(root.toString(), "smart-testing.yml");](https://github.com/arquillian/smart-testing/pull/254/files#diff-7aaa07c0a9bc90f3bf91fc1b16ce0fa3R134) 
[Paths.get(root.toString(), customConfigFile)](https://github.com/arquillian/smart-testing/pull/254/files#diff-7aaa07c0a9bc90f3bf91fc1b16ce0fa3R139)
here to this if, then you could use the same method `createConfigFile(customConfigFile)` for both cases

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
hemanik: I am have a query regarding the Default Configuration File. Currently, there is nothing like a default configuration file that we support. In the event the of missing custom file, we check if the default config file i.e. `smart-testing.yml` is present in the else part. If not, then we use the configuration from the system properties. 
Do you suggest a change to this flow by having some default config file ?

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
```diff
@@ -22,36 +23,60 @@
...
     }
 
     public static Configuration load(File projectDir) {
+        Map<String, Object> yamlConfiguration;
+
+        final String customConfigFile = System.getProperty(SMART_TESTING_CONFIG);
+        if (customConfigFile != null) {
+            yamlConfiguration = getConfigurationFromCustomConfigFile(customConfigFile);
+        } else {
+            yamlConfiguration = getConfigurationFromDefaultConfigFile(projectDir);
```
bartoszmajsak: This as well might be a one liner:
`yamlConfiguration = readConfiguration(System.getProperty(SMART_TESTING_CONFIG, projectDir));`

Assuming we can have just one method taking care of loading a file from a given location.

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
```diff
@@ -22,36 +23,60 @@
...
+        }
+
+        final Object strategiesConfiguration = yamlConfiguration.get("strategiesConfiguration");
+
+        final Configuration configuration = parseConfiguration(yamlConfiguration);
+        if (strategiesConfiguration != null) {
+            configuration.setStrategiesConfig((Map<String, Object>) strategiesConfiguration);
+        }
```
bartoszmajsak: This flow is a bit confusing for me:

- we first read the `strategiesConfiguration` object from bare Map
- we convert this map to `Configuration` object
- we only then set the property `strategiesConfiguration` back to `Configuration` object

Can't we just have it as a part of `parseConfiguration` instead?

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
MatousJobanek: > Currently, there is nothing like a default configuration file that we support. 

Yeah, there is no default file, we just have default values for some of the parameters which means that as a result, we can construct a configuration with these default values. 

Correct me if I'm wrong - if the custom file does not exist, you just warn the message and return an empty Map:
```java
        if (!file.exists()) {
            logger.warn("Config file `" + file.getName() + "` is not found. "
                + "Checking for config file `" + SMART_TESTING_YAML + "` OR `" + SMART_TESTING_YML);
        } else {
            return getConfigParametersFromFile(getConfigurationFilePath(file));
        }
        return Collections.EMPTY_MAP;
```
What I meant is that when the custom file does not exist, then we should try to load the configuration from the default locations - as it is done in `getConfigurationFromDefaultConfigFile`

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
MatousJobanek: This would just move the problem to the other method.
The reason why it is done in this way is that the `strategiesConfig` parameter is kind of temporary storage for this Map of strategy parameters. We need to store it there because when the `Configuration` is loaded for the first time, we still don't have all dependencies on classpath - we don't know which strategies should be loaded. When the strategies are defined and loaded, we can call `Configuration.loadStrategyConfigurations` which will use this variable and loads/maps the strategy related configurations.
But you are right - these lines can be moved to the `parseConfiguration` method, but has to stay in the same way - something like this:
```java
    private static Configuration parseConfiguration(Map<String, Object> yamlConfiguration) {
        final Object strategiesConfiguration = yamlConfiguration.get("strategiesConfiguration");
        Configuration configuration = mapToObject(Configuration.class, yamlConfiguration);
        if (strategiesConfiguration != null) {
            configuration.setStrategiesConfig((Map<String, Object>) strategiesConfiguration);
        }
        return configuration;
    }
```
this will make the code cleaner

bartoszmajsak:
Upstream https://github.com/arquillian/smart-testing/commit/002355d6fd110b17b5a000d398e2565f783826d0, thanks @hemanik!

PR #251, opened 2017-10-31
Fixes plugin selection done via 'apply-to' property.

hemanik:
#### Short description of what this resolves:
Plugin selection done via `apply-to` property throws Illegal State Exception.

#### Changes proposed in this pull request:

- Check for the plugin name `surefire/failsafe` in artifactID.

Fixes #253

PR #255, opened 2017-11-02
Refactors effective plugin filters and error messages.

### LINKED ISSUE #253: Plugin selection using `ApplyTo` property not working
##### Issue Overview

Selecting a plugin using the `applyTo` property either from system properties or config file throws Illegal State Exception.

##### Expected Behaviour
Smart Testing should be be enabled only for the selected plugin set by the `apply.to` property.

##### Current Behaviour
Execution fails due to Illegal State Exception.

##### Steps To Reproduce
1. Execute any test from the test-bed with `apply.to` property set either as system property or via config file.
2. The test should fail.

hemanik:
Fixes ambiguous error message conveying version mismatch despite the correct version used for effective plugin selection.  

- refactors plugin selection filters and error messages. 

Fixes #253

### DIFF
--- a/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
+++ b/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
@@ -95,29 +95,23 @@ class MavenProjectConfigurator {
         Version version = Version.from(plugin.getVersion().trim());
         if (!version.isGreaterOrEqualThan(MINIMUM_VERSION)) {
             failBecauseOfPluginVersionMismatch(model);
-            return false;
         }
         return true;
     }
 
     private void failBecauseOfMissingApplicablePlugin(Model model) {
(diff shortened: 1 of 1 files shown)


hemanik:
@bartoszmajsak, with reference to your comment https://github.com/arquillian/smart-testing/pull/251#issuecomment-340849584

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java
```diff
@@ -92,6 +83,34 @@ void configureTestRunner(Model model) {
...
+        }
+        // If not set the plugin is usable
+        return true;
+    }
+
+    private boolean hasMinimumVersionRequired(Plugin plugin, Model model) {
+        Version version = Version.from(plugin.getVersion().trim());
+        if (!version.isGreaterOrEqualThan(MINIMUM_VERSION)) {
+            failBecauseOfPluginVersionMismatch(model);
+            return false;
```
bartoszmajsak: If we are failing with exception this line will never be reached.

PR #262, opened 2017-11-15
feat(#114): implements configuration file per module.

### LINKED ISSUE #114: Introduce per-module configuration file
Add ability to overwrite configuration defined by root `smart-testing.yaml` file within each project module.

**Current assumption**: we require to have a file in the root folder as an enabler for our tool.

- [x] implement precedence logic
- [ ] update documentation (In Progess)
- [x] add ability for `Test Bed` to define per-module configuration (and at adjust at least one existing test)

dipak-pawar:
- Supporting configuration file per module
- Introduced `inherit` parameter in configuration file. 
- Updated documentation
- Adds  ftest in test-bed for multi-module project.

```xml
parent
 - config
   - smart-testing.yml 
   - api
   - impl-base
   - spi
     - smart-testing.yml
 - container
    - api
    - impl-base
 - smart-testing.yml 
```

Configuration file selection for the above example will be as follows:

* config/api - parent/config/smart-testing.yml
* config/impl-base - parent/config/smart-testing.yml
* config/spi - parent/config/spi/smart-testing.yml
* container/api - parent/smart-testing.yml
* container/impl-base - parent/smart-testing.yml

Fixes #114


functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectConfigurator.java
```diff
@@ -63,16 +62,19 @@ public ProjectConfigurator withConfiguration(Configuration configuration) {
         return this;
     }
 
-    public ProjectConfigurator createConfigFile() {
+    public ProjectConfigurator withConfigFile() {
```
bartoszmajsak: What is the reason behind this rename?

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/ConfigurationFilePerModuleTest.java
```diff
@@ -0,0 +1,71 @@
...
+                .mode(SELECTING)
+            .build();
+
+        project.configureSmartTesting()
+                .withConfiguration(parentConfiguration)
+                    .withConfigFile()
+                .withConfiguration(newConfiguration)
+                    .withConfigFileIn("config/impl-base")
+                .withConfiguration(changedConfiguration)
+                    .withConfigFileIn("junit/core")
```
bartoszmajsak: This chain is a bit hard for me to follow. Can you explain what is the desired flow of action here? I think `with...` makes it a bit blurry to follow as not everything should be prefixed like that.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -124,16 +127,47 @@ private void calculateChanges(File projectDirectory, Configuration configuration
...
+    private void configureMavenProject(MavenProjectConfigurator mavenProjectConfigurator, MavenProject mavenProject,
+        Configuration configuration) {
+        boolean wasConfigured = mavenProjectConfigurator.configureTestRunner(mavenProject.getModel());
+        if (wasConfigured) {
+            configuration.dump(mavenProject.getBasedir());
+            if (isFailedStrategyUsed()) {
+                SurefireReportStorage.copySurefireReports(mavenProject.getModel());
             }
-        });
+        }
```
bartoszmajsak: My proposal for refactoring:

```java
private void configureExtension(MavenSession session, Configuration configuration) {
    final Consumer<MavenProject> configureSmartTestingExtensionAction;
    if (hasModuleSpecificConfigurations(session)) {
        configureSmartTestingExtensionAction = applyModuleSpecificConfiguration(configuration);
    } else {
        configureSmartTestingExtensionAction = mavenProject -> configureMavenProject(mavenProjectConfigurator, mavenProject, configuration);
    }
    session.getAllProjects().forEach(configureSmartTestingExtensionAction);
}

private boolean hasModuleSpecificConfigurations(MavenSession session) {
    final ConfigLookup lookUp = new ConfigLookup(session.getExecutionRootDirectory());
    final boolean moreThanOneConfigFile = lookUp.hasMoreThanOneConfigFile(SMART_TESTING_YML, SMART_TESTING_YAML);
    return moreThanOneConfigFile && System.getProperty(SMART_TESTING_CONFIG) == null;
}

private Consumer<MavenProject> applyModuleSpecificConfiguration(Configuration configuration) {
    return mavenProject -> {
        final ConfigLookup configLookup = new ConfigLookup(mavenProject.getBasedir().toString());
        final File dirWithConfig = configLookup.getFirstDirWithConfigOrProjectRootDir();
        final Configuration mavenProjectConfiguration = configLookup.isConfigFromProjectRootDir() ? configuration :
            ConfigurationLoader.load(dirWithConfig);
        if (mavenProjectConfiguration.isDisable()) {
            logExtensionDisableReason(logger,
                SMART_TESTING_DISABLE + " is set for module " + mavenProject.getArtifactId());
            return;
        }
        if (mavenProjectConfiguration.areStrategiesDefined()) {
            final MavenProjectConfigurator mavenProjectConfigurator =
                new MavenProjectConfigurator(mavenProjectConfiguration);
            configureMavenProject(mavenProjectConfigurator, mavenProject, mavenProjectConfiguration);
        } else {
            logger.warn(
                "Smart Testing Extension is installed but no strategies are provided for %s module. It won't influence the way how your tests are executed. "
                    + "For details on how to configure it head over to http://bit.ly/st-config",
                mavenProject.getArtifactId());
        }
    };
}
```

docs/configfile.adoc
```diff
@@ -13,27 +13,70 @@ include::../core/src/test/resources/configuration/smart-testing.yml[]
...
+=== How Configuration parameters are inherited?
+If config file has `inherit` defined with absolute or relative path of config file, then Smart Testing will
+lookup for all undefined properties in config file defined using inherit. However it won't lookup for `disable` parameter.
+
+If config file doesn't contain `inherit` parameter. Then Smart Testing won't inherit any configuration properties.
+
+=== Configuration File per Maven Module:
+You can define configuration file per maven module. However it's not mandatory to define configuration per each module.
+
+NOTE: Whenever you are using configuration file per maven module, make sure to define strategies & scm properties
```
bartoszmajsak: Are scm properties really required? Why should they be?

docs/configfile.adoc
```diff
@@ -13,27 +13,70 @@ include::../core/src/test/resources/configuration/smart-testing.yml[]
...
+ - smart-testing.yml #<!--2-->
+```
+
+Configuration file selection for the above example will be as follows:
+
+* config/api - parent/config/smart-testing.yml
+* config/impl-base - parent/config/smart-testing.yml
+* config/spi - parent/config/spi/smart-testing.yml
+* container/api - parent/smart-testing.yml
+* container/impl-base - parent/smart-testing.yml
```
bartoszmajsak: What do you think about having concrete file content listed and then used as an example to show effective resolution instead?

docs/configfile.adoc
dipak-pawar: yes it needs to be define in parent config file OR define it using system properties. Because we are loading scm changes by looking into parent configuration file.

otherwise ST couldn't find required changes between commit ranges. as it will take default range which is `HEAD-HEAD~0`

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -26,13 +69,60 @@
...
+            }
+            if (field.getType().isArray() && Arrays.equals(((Object[]) actualValue), ((Object[]) defaultValue))) {
+                defaultValues.add(field);
+                return;
+            }
+            if (actualValue != null && actualValue.equals(defaultValue)) {
+                defaultValues.add(field);
+            }
+        } catch (IllegalAccessException e) {
+            throw new IllegalStateException("Failed to access fieldName: " + field, e);
```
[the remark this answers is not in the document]
dipak-pawar: Good catch

docs/configfile.adoc
bartoszmajsak: So it is not required but needed if you use it. Maybe we could clarify this sentence then.

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -26,13 +69,60 @@
...
+    private String fieldName(Method method) {
+        final String field = method.getName().substring(3);
+        return Character.toLowerCase(field.charAt(0)) + field.substring(1);
+    }
+
+    private <T extends ConfigurationSection> T mapToDefaultObject(Class<T> aClass, Map<String, Object> map) {
+        this.userSetProperty = false;
+        final T defaultObject = readValue(aClass, map);
+        this.userSetProperty = true;
+        return defaultObject;
```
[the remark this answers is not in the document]
dipak-pawar: mapToObject(readValue) logic is looking for system property -> config file -> default value.
Not if user is running his test with any system property, then to get instance with default values I am setting this property to `false` first & after getting defaultInstance setting it back to `true`.

With this property changing the look up logic like in readValue
```
if !userSetProperty
  look for only default values
else 
 system property -> config file -> default
```
Have a look at  [this](https://github.com/arquillian/smart-testing/pull/262/files/eea8c8c051eb5f31a0cab508824d3bcbe1fc7154#diff-ffcb72af5109482890b254359d4a68f4R163)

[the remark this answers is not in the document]
bartoszmajsak: I understand what it does, but it's a weird piece of code for my taste if we rely on the exception to either set it to `true` or `false`.

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -26,13 +69,60 @@
...
+        List<Field> defaultValues) {
+        if (field.getName().equals(DISABLE) || field.getName().equals(INHERIT)) {
+            return;
+        }
+        final ConfigurationSection defaultConfiguration = mapToDefaultObject(Configuration.class, new HashMap<>(0));
+        field.setAccessible(true);
+        try {
+            final Object actualValue = field.get(configuration);
+            final Object defaultValue = field.get(defaultConfiguration);
+            if (actualValue == null && defaultValue == null) {
```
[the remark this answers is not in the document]
dipak-pawar: "default" means creating Configuration object without any system property & config file.

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -56,21 +146,23 @@
         }
     }
 
-    private static Object getConvertedObject(Method method, Object configFileValue, Optional<ConfigurationItem> foundConfigItem) {
+    private Object getConvertedObject(Method method, Object configFileValue,
+        Optional<ConfigurationItem> foundConfigItem) {
```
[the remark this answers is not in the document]
dipak-pawar: It was already there. But yes we can improve this.

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
```diff
@@ -39,8 +39,17 @@ public static Configuration load(File projectDir) {
...
+
+       return readParseOverWriteConfiguration(configFile);
+    }
+
+    private static Configuration readParseOverWriteConfiguration(File configFile) {
+        Path configFileDir = configFile.isFile()? configFile.getParentFile().toPath(): configFile.toPath();
         Map<String, Object> yamlConfiguration = readConfiguration(configFile);
-        return parseConfiguration(yamlConfiguration);
+        final Configuration configuration = parseConfiguration(yamlConfiguration);
+        final ObjectMapper objectMapper = new ObjectMapper();
```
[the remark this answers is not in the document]
dipak-pawar: We are using it only once in ConfigurationLoader. I am not sure why do we really need to declare it as constant?

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
```diff
@@ -64,13 +73,17 @@ public static Configuration load(File projectDir) {
         return Collections.emptyMap();
     }
 
-    private static Map<String, Object> getConfigParametersFromFile(Path filePath) {
+    static Map<String, Object> getConfigParametersFromFile(Path filePath) {
+        if (!filePath.toFile().exists()) {
+            logger.warn(String.format("The configuration file %s is not exists.", filePath));
+            return new HashMap<>(0);
```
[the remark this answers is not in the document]
dipak-pawar: no we are not mutating this map afterwords

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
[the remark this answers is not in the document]
bartoszmajsak: I misinterpreted it with Jackson object mapper object at first, so probably not. OTOH I don't feel like having `mapToObject` method with now a constructor and a call to `readValue` is really needed.

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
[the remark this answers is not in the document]
bartoszmajsak: I still don't feel like this `if` statement is easy to reason about. If it doesn't have actual value and no default value treat is as default value?

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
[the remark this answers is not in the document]
bartoszmajsak: Sure, it only means we missed it in the previous review.

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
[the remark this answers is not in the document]
bartoszmajsak: Then maybe `.emptyMap()` would be better so we actually prevent from mutating it in the caller anyway (at least the objects in the map, not the content itself)

core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationFileBuilder.java
```diff
@@ -7,44 +7,44 @@
...
     private HashMap<String, Object> properties;
 
-    private ConfigurationFile() {
+    private ConfigurationFileBuilder() {
         properties = new HashMap<>();
     }
 
-    static ConfigurationFile SmartTestingConfigurationFile() {
-        return new ConfigurationFile();
+    static ConfigurationFileBuilder SmartTestingConfigurationFile() {
```
bartoszmajsak: Uppercase in method name is a bit unusual for Java coders unless it's a constructor. Can we rename it to simple `configurationFile`? Also `create` could be `writeTo`, so it could be written as

```java
configurationFile()
    .mode("ordering")
    .applyTo("surefire")
    .inherit(Paths.get(root, SMART_TESTING_YAML).toString())
    .writeTo(Paths.get(root, SMART_TESTING_YML));
```

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationReader.java
```diff
@@ -88,4 +91,27 @@ private Path getDefaultConfigFile(File... files) {
...
+        Map<String, Object> config = getConfigParametersFromFile(filePath);
+
+        List<Map<String, Object>> configs = new ArrayList<>();
+        configs.add(config);
+        while (config.get(INHERIT) != null) {
+            String inherit = String.valueOf(config.get(INHERIT));
+            filePath = filePath.getParent().resolve(inherit);
+            config = getConfigParametersFromFile(filePath);
+            if (!config.isEmpty()) {
+                configs.add(0, config);
```
bartoszmajsak: Alternatively, you can use [`Deque`](https://docs.oracle.com/javase/7/docs/api/java/util/Deque.html) as a Stack - might make it a bit easier to understand what we really do with inheritance.

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -8,15 +8,15 @@
...
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
 class ObjectMapper {
 
-    static <T extends ConfigurationSection> T mapToObject(Class<T> aClass, Map<String, Object> map) {
+    <T extends ConfigurationSection> T readValue(Class<T> aClass, Map<String, Object> map) {
```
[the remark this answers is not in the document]
MatousJobanek: And also this usage: https://github.com/arquillian/smart-testing/pull/262/files#diff-4b3896eea6384f8873440fb7836cbfa0R31

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/ConfigurationFilePerModuleTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+import static org.arquillian.smart.testing.RunMode.SELECTING;
+import static org.arquillian.smart.testing.ftest.testbed.TestRepository.testRepository;
+import static org.arquillian.smart.testing.ftest.testbed.configuration.Strategy.CHANGED;
+import static org.arquillian.smart.testing.ftest.testbed.configuration.Strategy.NEW;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ConfigurationFilePerModuleTest {
```
MatousJobanek: Most of the functional tests end with `FunctionalTest` string

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/ConfigurationFilePerModuleTest.java
```diff
@@ -0,0 +1,71 @@
...
+                .withConfiguration(newConfiguration)
+                    .createConfigFileIn("config/impl-base")
+                .withConfiguration(changedConfiguration)
+                    .createConfigFileIn("junit/core")
+            .enable();
+
+        final Collection<TestResult> expectedTestResults = project
+            .applyAsLocalChanges("Adds new unit test");
+
+        expectedTestResults.addAll(project.applyAsCommits("Deletes one test", "Renames unit test"));
```
MatousJobanek: Having this set of commits with the given configuration won't verify that the correct configuration is used.
configuration in parent dir has `selecting` (default) mode and `new, changed` strategies
for `config/impl-base` with the strategy `new`, mode `selecting` and you use `Adds new unit test` commit, which means that even if the parent configuration is read, then the test will pass - always run only the new test.
Similarly, for the `junit/core` module.
you can verify it removing the lines 53-56

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -133,16 +139,56 @@ private void calculateChanges(File projectDirectory, Configuration configuration
...
+        final ConfigLookup lookUp = new ConfigLookup(projectDir, stopCondition);
+        final boolean moreThanOneConfigFile = lookUp.hasMoreThanOneConfigFile(SMART_TESTING_YML, SMART_TESTING_YAML);
+        return moreThanOneConfigFile && System.getProperty(SMART_TESTING_CONFIG) == null;
+    }
+
+    private Consumer<MavenProject> applyModuleSpecificConfiguration(Configuration configuration) {
+        return mavenProject -> {
+            final ConfigLookup configLookup = new ConfigLookup(mavenProject.getBasedir(), this::isProjectRootDirectory);
+            final File dirWithConfig = configLookup.getFirstDirWithConfigOrWithStopCondition();
+            final Configuration mavenProjectConfiguration = configLookup.isConfigFromProjectRootDir() ? configuration :
```
MatousJobanek: On the previous line, you call `configLookup.getFirstDirWithConfigOrWithStopCondition()` and here using the method `configLookup.isConfigFromProjectRootDir()` you [call](https://github.com/arquillian/smart-testing/pull/262/files#diff-0d186905ac15e0569433af7b3e556ccbR51) the same method again: `getFirstDirWithConfigOrWithStopCondition()`
To be honest, this is very cryptic to my taste.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -133,16 +139,56 @@ private void calculateChanges(File projectDirectory, Configuration configuration
...
+                ConfigurationLoader.load(dirWithConfig);
+            if (mavenProjectConfiguration.isDisable()) {
+                logExtensionDisableReason(logger,
+                    SMART_TESTING_DISABLE + " is set for module " + mavenProject.getArtifactId());
```
MatousJobanek: Please, use the similar message as it is used here: https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/MavenProjectConfigurator.java#L67

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -133,16 +139,56 @@ private void calculateChanges(File projectDirectory, Configuration configuration
...
+            if (mavenProjectConfiguration.areStrategiesDefined()) {
+                final MavenProjectConfigurator mavenProjectConfigurator =
+                    new MavenProjectConfigurator(mavenProjectConfiguration);
+                configureMavenProject(mavenProjectConfigurator, mavenProject, mavenProjectConfiguration);
+            } else {
+                logger.warn(
+                    "Smart Testing Extension is installed but no strategies are provided for %s module. It won't influence the way how your tests are executed. "
+                        + "For details on how to configure it head over to http://bit.ly/st-config",
+                    mavenProject.getArtifactId());
+            }
```
MatousJobanek: maybe, it would be good to have the checks (if the ST should be installed for the module) in separated class. 
As it is done for [SkipInstallationChecker](https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SkipInstallationChecker.java) as well as for [SkipModuleChecker](https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SkipModuleChecker.java)

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
```diff
@@ -133,16 +139,56 @@ private void calculateChanges(File projectDirectory, Configuration configuration
...
+                ConfigurationLoader.load(dirWithConfig);
+            if (mavenProjectConfiguration.isDisable()) {
+                logExtensionDisableReason(logger,
+                    SMART_TESTING_DISABLE + " is set for module " + mavenProject.getArtifactId());
+                return;
             }
-        });
+            if (mavenProjectConfiguration.areStrategiesDefined()) {
+                final MavenProjectConfigurator mavenProjectConfigurator =
+                    new MavenProjectConfigurator(mavenProjectConfiguration);
```
MatousJobanek: the instantiation of this class can be inside of the `configureMavenProject` method - you pass the configuration instance there anyway

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
MatousJobanek: I'm asking myself, is it really worth having this complicated check there? When looking at the code, it doesn't seem to be an optimization at all - this solution is maybe more time and resource consuming.

core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationOverWriteUsingInheritTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+        SmartTestingConfigurationFile()
+            .inherit("../smart-testing.yml")
+            .create(Paths.get(root, CONFIG, SMART_TESTING_YML));
+
+        SmartTestingConfigurationFile()
+            .inherit("../smart-testing.yml")
+            .mode("selecting")
+            .debug(true)
+            .create(Paths.get(root, CONFIG, IMPL_BASE, SMART_TESTING_YML));
```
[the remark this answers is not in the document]
MatousJobanek: It would be also good to change the names of the files to express the inheritance

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
MatousJobanek: To be honest, it took me some time to understand the code. as for assigning the consumer function to the local variable - I would use it only when the method returns some value. In this case I would use this:
```java 
    private void configureExtension(MavenSession session, Configuration configuration) {
        final File executionRootDirectory = new File(session.getExecutionRootDirectory());
        List<MavenProject> allProjects = session.getAllProjects();
        if (hasModuleSpecificConfigurations(executionRootDirectory, this::isProjectRootDirectory)) {
            allProjects.forEach(mavenProject -> applyModuleSpecificConfiguration(configuration, mavenProject));
        } else {
            allProjects.forEach(mavenProject -> configureMavenProject(mavenProjectConfigurator, mavenProject, configuration));
        }
    }

    private void applyModuleSpecificConfiguration(Configuration configuration, MavenProject mavenProject) {
        final ConfigLookup configLookup = new ConfigLookup(mavenProject.getBasedir(), this::isProjectRootDirectory);
        final File dirWithConfig = configLookup.getFirstDirWithConfigOrWithStopCondition();
        final Configuration mavenProjectConfiguration = configLookup.isConfigFromProjectRootDir() ? configuration :
            ConfigurationLoader.load(dirWithConfig);
        if (mavenProjectConfiguration.isDisable()) {
            logExtensionDisableReason(logger,
                SMART_TESTING_DISABLE + " is set for module " + mavenProject.getArtifactId());
            return;
        }
        if (mavenProjectConfiguration.areStrategiesDefined()) {
            final MavenProjectConfigurator mavenProjectConfigurator =
                new MavenProjectConfigurator(mavenProjectConfiguration);
            configureMavenProject(mavenProjectConfigurator, mavenProject, mavenProjectConfiguration);
        } else {
            logger.warn(
                "Smart Testing Extension is installed but no strategies are provided for %s module. It won't influence the way how your tests are executed. "
                    + "For details on how to configure it head over to http://bit.ly/st-config",
                mavenProject.getArtifactId());
        }
    }
```
Just my opinion ;-)

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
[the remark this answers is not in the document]
dipak-pawar: I am fine with it. But as @bartoszmajsak suggested me https://github.com/arquillian/smart-testing/pull/262#discussion_r152486529 & also we have a discussion about the same in mattermost.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/ConfigurationFilePerModuleTest.java
dipak-pawar: yes this way we can make sure that configuration file applied correctly.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
dipak-pawar: ok let me see how can we make it more readable & less complicated

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
dipak-pawar: I think we are just checking `configuration.isDisable()` to check to ST should be installed for module. So I don't think we should move it to separated class.

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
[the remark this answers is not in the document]
bartoszmajsak: My comment was about object mapper variable in that method, not really about the method itself being static.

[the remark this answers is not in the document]
dipak-pawar: ok then I will make it static.

core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationReader.java
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+        Map<String, Object> effectiveConfig = configs.get(0);
+        for (int i = 1; i < configs.size(); i++){
+            effectiveConfig.putAll(configs.get(i));
```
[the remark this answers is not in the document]
dipak-pawar: something to discuss over BJN

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java
MatousJobanek: You check two things here: 
`mavenProjectConfiguration.isDisable()`
and
`mavenProjectConfiguration.areStrategiesDefined()`

docs/configfile.adoc
```diff
@@ -13,27 +13,92 @@ include::../core/src/test/resources/configuration/smart-testing.yml[]
...
+the parent dir containing pom.xml.
+
+=== How Configuration parameters are inherited?
+If config file has `inherit` defined with absolute or relative path of config file, then Smart Testing will
+lookup for all undefined properties in config file defined using inherit. However it won't lookup for `disable` parameter.
+
+If config file doesn't contain `inherit` parameter. Then Smart Testing won't inherit any configuration properties.
+
+=== Configuration File per Maven Module:
+You can define configuration file per maven module. However it's not mandatory to define configuration per each module.
```
MatousJobanek: this is not necessary I think -  it describes the same as on the line 39

docs/configfile.adoc
```diff
@@ -13,27 +13,92 @@ include::../core/src/test/resources/configuration/smart-testing.yml[]
...
+
+NOTE: Whenever you are using configuration file per maven module, make sure to define strategies in project's
+parent configuration file or using system property. If you are running tests against range of commits, you need to define scm properties in project's parent config file or using system property.
```
MatousJobanek: > Whenever you are using configuration file per maven module, make sure to define strategies in project's parent configuration file or using system property

This is very confusing - it can be specified also inside of the module config file. I would remove it.

docs/configfile.adoc
```diff
@@ -13,27 +13,92 @@ include::../core/src/test/resources/configuration/smart-testing.yml[]
...
+   - impl-base
+   - spi
+     - smart-testing.yml #<!--2-->
+ - container
+    - api
+    - impl-base
+ - smart-testing.yml #<!--3-->
+```
+
+<1>
```
MatousJobanek: these numbers are not correctly formatted

docs/configfile.adoc
```diff
@@ -13,27 +13,92 @@ include::../core/src/test/resources/configuration/smart-testing.yml[]
...
+.parent/config/smart-testing.yml
+----
+inherit: ../smart-testing.yml
+strategies: affected
+----
+
+<2>
+parent/config/spi/smart-testing.yml
```
MatousJobanek: in the other cases, there is a dot at the beginning. not here. is it intentional?

docs/configfile.adoc
```diff
@@ -43,6 +108,9 @@ NOTE: You can use either a `.yml` or `.yaml` extension for this file.
 |===
 |Field | Description
 
+a| inherit
+a| This is used to define absolute or relative path for configuration file from where you want to overwrite paremeters not defined in current config file.
```
MatousJobanek: you don't overwrite parameters that are not defined - there is nothing to overwrite ;-)
I think that enough is saying:
This is used to define absolute or relative path to parent configuration file

docs/configfile.adoc
```diff
@@ -13,27 +13,92 @@ include::../core/src/test/resources/configuration/smart-testing.yml[]
...
+lookup for all undefined properties in config file defined using inherit. However it won't lookup for `disable` parameter.
+
+If config file doesn't contain `inherit` parameter. Then Smart Testing won't inherit any configuration properties.
+
+=== Configuration File per Maven Module:
+You can define configuration file per maven module. However it's not mandatory to define configuration per each module.
+
+NOTE: Whenever you are using configuration file per maven module, make sure to define strategies in project's
+parent configuration file or using system property. If you are running tests against range of commits, you need to define scm properties in project's parent config file or using system property.
+It won't work if you define scm properties in any module's config file and not in parent config file.
```
MatousJobanek: Please, create an issue for it.

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
```diff
@@ -17,6 +17,11 @@
 class ObjectMapper {
 
     static <T extends ConfigurationSection> T mapToObject(Class<T> aClass, Map<String, Object> map) {
+        final ObjectMapper objectMapper = new ObjectMapper();
+        return objectMapper.readValue(aClass, map);
+    }
+
+    <T extends ConfigurationSection> T readValue(Class<T> aClass, Map<String, Object> map) {
```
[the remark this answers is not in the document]
dipak-pawar: it's not necessary. We can remove this. but I created it, to keep it clean. & with `mapToObject` we have static factory method due to which we don't have to create `ObjectMapper` each time.

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/ConfigurationFilePerModuleFunctionalTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+public class ConfigurationFilePerModuleFunctionalTest {
+    @ClassRule
+    public static final GitClone GIT_CLONE = new GitClone(testRepository());
+
+    @Rule
+    public final TestBed testBed = new TestBed(GIT_CLONE);
+
+    @Test
+    public void should_load_configuration_from_modules_config_file_for_local_changes_where_parent_config_file_has_only_strategies_defined() {
```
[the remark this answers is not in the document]
dipak-pawar: yes, there is difference, in one test -  parent config having only `strategy` defined. & in another parent config having `scm` & `strategy` defined which is already mentioned in method name

docs/configfile.adoc
```diff
@@ -13,27 +13,92 @@ include::../core/src/test/resources/configuration/smart-testing.yml[]
...
 
 All parameters in configuration file are optional. If you haven't used any parameter in configuration file, Smart testing will use default value for that parameter.
 You can look at <<_reference_card, references>> for default value of parameter.
 
 However you can overwrite all configuration options using system properties supported by Smart Testing.
 You can look <<_reference_card, references>> for all supported system properties.
 
+=== How configuration is applying to each module?
+Smart Testing is looking in each module's root dir for configuration file. If it didn't find config file there, it'll look recursively for the first config file in
+the parent dir containing pom.xml.
```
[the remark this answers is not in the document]
dipak-pawar: yes after all your changes this is not true

docs/configfile.adoc
dipak-pawar: I think it's good to have it.

docs/configfile.adoc
dipak-pawar: no you can't specify it inside module config file. You have to define it in project's parent config file or using system property. Otherwise Smart testing will load parent configuration having no strategy defined which will disable ST having reason no strategy defined at least in case of mvn extension

docs/configfile.adoc
dipak-pawar: no I missed it here. thanks for catching it up.

docs/configfile.adoc
```diff
@@ -13,27 +13,92 @@ include::../core/src/test/resources/configuration/smart-testing.yml[]
...
+Configuration file selection for the above example will be as follows:
+
+* config/api - parent/config/smart-testing.yml
+* config/impl-base - parent/config/smart-testing.yml
+* config/spi - parent/config/spi/smart-testing.yml
+* container/api - parent/smart-testing.yml
+* container/impl-base - parent/smart-testing.yml
+
+<1> inside this file we have inherit: ../smart-testing.yml`, so it will take all undefined properties in this file from the one in `parent/smart-testing.yml`.
+<3> inside this file we have `strategies: new` & `scm: lastChanges: 1` defined.(You can set it using respective system property also).
```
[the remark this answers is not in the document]
dipak-pawar: I think it's not needed. but let's add it to make it consistent

docs/configfile.adoc
dipak-pawar: you mean numbers mentioned in xml config file

docs/configfile.adoc
dipak-pawar: created #274

core/src/main/java/org/arquillian/smart/testing/configuration/ObjectMapper.java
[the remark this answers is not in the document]
MatousJobanek: > With mapToObject we have static factory method due to which you don't have to create ObjectMapper instance each time.

if it is a static method, then you don't create any instance of the class.
This approach is wrongly used in several cases in the ST code - instead of using classes such as `ConfigurationChecker` and `ConfigurationReader` where you need to create an instance of them and then call one method (without any need to keep the state) you can use them as static "util" classes.

docs/configfile.adoc
MatousJobanek: > Smart Testing is looking in each module's root dir for configuration file. If it didn't find config file there, it'll look recursively for the first config file in the parent dir containing pom.xml.

has the very same meaning as 

> You can define configuration file per maven module. However, it's not mandatory to define configuration per each module.

both of them says that you can define the config file in maven the module (or in parent dirs) and if it's not there (which means that it is not mandatory) it will look recursively for the first config file

docs/configfile.adoc
MatousJobanek: Then describe it in this way as you just did in your comment.

This also brings a good point - that you can use at these [places](https://github.com/arquillian/smart-testing/blob/master/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/SmartTestingMavenConfigurer.java#L61) your `ModuleSTInstallationChecker` logic

docs/configfile.adoc
MatousJobanek: I mean the numbers `<1>` in the resulting documentation page

docs/configfile.adoc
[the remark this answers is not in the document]
MatousJobanek: it is needed, otherwise the documentations fails

functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configurationfile/ConfigurationFilePerModuleFunctionalTest.java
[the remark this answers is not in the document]
MatousJobanek: is it possible to merge it into one test?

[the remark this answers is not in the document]
dipak-pawar: ok I will merge it into one.

dipak-pawar:
@bartoszmajsak , @MatousJobanek Can you have a look into it?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/checker/SkipSTInstallationChecker.java
```diff
@@ -0,0 +1,39 @@
+package org.arquillian.smart.testing.mvn.ext.checker;
+
+import org.apache.maven.project.MavenProject;
+import org.arquillian.smart.testing.configuration.Configuration;
+import org.arquillian.smart.testing.mvn.ext.dependencies.ExtensionVersion;
+
+import static org.arquillian.smart.testing.configuration.Configuration.SMART_TESTING_DISABLE;
+
+public class SkipSTInstallationChecker {
```
MatousJobanek: This class name is very similar to: `SkipInstallationChecker` it's very confusing.
Please, merge these two classes. The constructor can stay same, you just add another method `shouldSkipForConfig(Configuration config)` which you will call at the places where you call your `SkipSTInstallationChecker`

bartoszmajsak:
@dipak-pawar thanks for your work! Please take care of the merge conflict and merge.

PR #265, opened 2017-11-21
chore: Refactored & moved autocorrect functionality to Configuration class

### LINKED ISSUE #269: Move autocorrect functionality located in DependencyResolver class to the Configuration class
##### Issue Overview

In the DependencyResolver class, there is a quite complex logic taking care of misspelled strategy names that would be useful to have it available also of other integration cases (eg. with Che). Thus, this logic should be moved to the Configuration class located in core.

MatousJobanek:

The autocorrect functionality has to be available in core to have it accessible (applicable) also in other integration implementations such as in Che

- Moved autocorrect functionality from Maven extension (`DependencyResolver.java`) to Core (`Configuration.java`). The `autocorrectStrategies` method is then called from Smart Testing API

Fixes #269

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/impl/TestStrategyApplierImpl.java
+++ b/core/src/main/java/org/arquillian/smart/testing/impl/TestStrategyApplierImpl.java
@@ -57,18 +57,21 @@ class TestStrategyApplierImpl implements TestStrategyApplier {
     }
 
     private Set<TestSelection> selectTests(Configuration configuration) {
-
-        final List<String> strategies = Arrays.asList(configuration.getStrategies());
-        if (strategies.isEmpty()) {
+        if (configuration.getStrategies().length == 0) {
             logger.warn(
                 "Smart Testing Extension is installed but no strategies are provided. It won't influence the way how your tests are executed. "
                     + "For details on how to configure it head over to http://bit.ly/st-config");
             return Collections.emptySet();
         }
 
+        List<String> errorMessages = new ArrayList<>();
+        configuration.autocorrectStrategies(testExecutionPlannerLoader.getAvailableStrategyNames(), errorMessages);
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/arquillian/smart/testing/impl/TestStrategyApplierImpl.java
```diff
@@ -57,18 +57,21 @@
...
             logger.warn(
                 "Smart Testing Extension is installed but no strategies are provided. It won't influence the way how your tests are executed. "
                     + "For details on how to configure it head over to http://bit.ly/st-config");
             return Collections.emptySet();
         }
 
+        List<String> errorMessages = new ArrayList<>();
+        configuration.autocorrectStrategies(testExecutionPlannerLoader.getAvailableStrategyNames(), errorMessages);
+        errorMessages.forEach(msg -> logger.error(msg));
+
```
dipak-pawar: We are throwing exception for [DependencyResolver](https://github.com/arquillian/smart-testing/pull/265/files#diff-e5427bc921ef9faf2177273978bd12feR41) if we have any errorMessages.

Is there any reason why are we silently ignoring it in ST api?

MatousJobanek: Well, the usage of the autocorrection in the ST is kind of safety.
There are two usages:
* Maven integration where the first check is performed on the Maven extension side
* Che and other integrations, where the Configuration is loaded before the ST API is called, so the integration can autocorrect the mistyped strategies and throw an exception if needed. 

This line in `TestStrategyApplierImpl.java` class is just for sure that no understandable exception is thrown and just prints the warning. 
This is my understanding and expectations. If there are cases/integrations when the exception should be thrown, then it should do on its own.

PR #271, opened 2017-11-27
fix(#270): when loading test reports it reflects location set by API 

### LINKED ISSUE #270: When the test reports are loaded it should use the directory set by API
##### Issue Overview

Currently, the `InProjectTestReportLoader` uses the current directory the build is running in `Paths.get(".")` for loading test report files. It should use that one set by API, otherwise, it is not possible to use the API from a different directory.

MatousJobanek:

When the test reports are being loaded it takes into account the location set by API
I have also added some additional constructors to `TestResult` class for easier usage.

Fixes #270

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/spi/TestResult.java
+++ b/core/src/main/java/org/arquillian/smart/testing/spi/TestResult.java
@@ -9,11 +9,19 @@ public class TestResult {
     private final Float testDuration;
     private Result result;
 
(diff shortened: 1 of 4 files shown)


bartoszmajsak:
Please rename the PR header, so it starts with `fix(#issue_nr)`. Also maybe we should agree on starting the description after the label with lower or uppercase - minor thing, but having it unified will make a log a bit cleaner.

PR #273, opened 2017-11-27
fix(#272): ownership of test report directory storage is moved to TemporaryInternalFiles class

### LINKED ISSUE #272: Ownership of test report directory storage should be moved to TemporaryInternalFiles class
##### Issue Overview

To improve the usage in case of other integration implementations (eg. Che) the ownership/management of the test report directory storage should be moved to `TemporaryInternalFiles` class (in similar way as it is done for the `scm-changes` file)
Currently, it is quite hard to guess the correct path to the directory where the report files are read.

MatousJobanek:

The ownership of the test report directory storage has been moved to `TemporaryInternalFiles` class, in a similar way as it is done for the `scm-changes` file.
Having this in one place, then it will be easier to use it from other integration implementations (eg. Che) when the test result report needs to be stored. Otherwise, it is quite had to guess the correct path to the directory where the files are read.

Fixes #272


bartoszmajsak:
Please rename the PR header, so it starts with `fix(#issue_nr)`. Also maybe we should agree on starting the description after the label with lower or uppercase - minor thing, but having it unified will make a log a bit cleaner.

bartoszmajsak:
As for failing build - something is buggy with doc generation - `const-inline-extension` is failing. To start with I don't think the doc generation should be started at all as we don't have any `.adoc` files changed. But on the other hand, I'm wondering what is the problem - any idea @lordofthejars?

core/src/main/java/org/arquillian/smart/testing/hub/storage/local/TemporaryInternalFiles.java
```diff
@@ -12,4 +15,19 @@ public static String getScmChangesFileName(){
...
+
+    public String getTestReportDirectoryName(){
+        return TEMP_REPORT_DIR;
+    }
+
+    public LocalStorageDirectoryAction getTestReportDirectoryAction(String rootDir){
+        return getTestReportDirectoryAction(new File(rootDir));
+    }
+
+    public LocalStorageDirectoryAction getTestReportDirectoryAction(File rootDir){
```
bartoszmajsak: Minor thing really, but I'm wondering if we should stick to `get` from Java Bean convention or rather have a hint that action is created, so `createTestReportDirectoryAction` instead.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
```diff
@@ -164,4 +162,10 @@ private TestResults accumulatedTestResults() {
             throw new RuntimeException("Failed extracting test results", e);
         }
     }
+
+    private boolean isSurefireReportFile(Path path) {
```
bartoszmajsak: :+1:

core/src/main/java/org/arquillian/smart/testing/hub/storage/local/TemporaryInternalFiles.java
MatousJobanek: +1, I was thinking about the same - `createTestReportDirectoryAction` sounds more natural for me. I just kept the method naming in the same way as it is used in other cases in this class. But looking at it this morning, the other methods just return the constants, but this one creates the class. So I'll change it

bartoszmajsak:
Added poor's man debug tool commit 4e5f38b - maybe it will hint something. We should revert it before merging with upstream.

.asciidoctor/lib/const-inline-macro/extension.rb
```diff
@@ -49,8 +49,8 @@ def process parent, target, attrs
       end
     end
     if const_value[0][0].nil?
```
dipak-pawar: This will again fail. Because `const_value` will be null as it won't find line with require const_name
Look at [1](https://github.com/arquillian/smart-testing/blob/master/.asciidoctor/lib/const-inline-macro/extension.rb#L29) & [2](https://github.com/arquillian/smart-testing/blob/master/.asciidoctor/lib/const-inline-macro/extension.rb#L42)

bartoszmajsak: Ah right :) I just assumed it's initialized as `[]`

.asciidoctor/lib/const-inline-macro/extension.rb
```diff
@@ -49,8 +49,8 @@ def process parent, target, attrs
...
-        missing_attr = if attrs.has_key? 'tag' then attrs['tag'] else attrs['name'] end
-        raise "Failed reading '#{data_path}'. Check if constant '#{missing_attr}' exists"
+      missing_attr = if attrs.has_key? 'tag' then attrs['tag'] else attrs['name'] end
```
dipak-pawar: This can be simplified as `missing_attr =  attrs.has_key? 'tag' ? attrs['tag'] : attrs['name'] `

bartoszmajsak: I'm not sure, as IRB gives me this:

```ruby
2.4.1 :034 > attrs = {}
 => {} 
2.4.1 :035 > attrs['tag'] = "test"
 => "test" 
2.4.1 :036 > missing_attr = attrs.has_key? 'tag' ? attrs['tag'] : attrs['name']
(irb):36: warning: string literal in condition
 => false 
2.4.1 :037 > puts missing_attr
false
 => nil 
2.4.1 :038 > 
```

.asciidoctor/lib/const-inline-macro/extension.rb
bartoszmajsak: I think we should check for both, really `const_value.nil? or const_value[0][0].nil?`

.asciidoctor/lib/const-inline-macro/extension.rb
bartoszmajsak: `missing_attr = (attrs.has_key? 'tag') ? attrs['tag'] : attrs['name']` would do

PR #276, opened 2017-12-05
Added -Z flag to work with SELinux

dipak-pawar:
While generating docs, I have observed that on my host, asciidoc container could not write to shared/mounted private volume.

Maybe you guys are not facing this issue as you are using docker-ce. But in my case I am using fedora distribution of docker installed using `dnf install docker`

Ref - http://www.projectatomic.io/blog/2016/03/dwalsh_selinux_containers/

### DIFF
--- a/.asciidoctor/generate.sh
+++ b/.asciidoctor/generate.sh
@@ -37,7 +37,7 @@ fi
 
 echo "Generating doc in $TARGET_FOLDER"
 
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
That's correct. I'm using `docker-ce`. I tried that locally and it seems to be working fine in my setup too. Great job. Go ahead with the merge if the build is green.

PR #286, opened 2017-12-08
fix(#279): strategy deps are shaded and excluded their transitive deps

### LINKED ISSUE #279: Be as less intrusive as possible with the dependencies
##### Issue Overview

Most of the strategy implementations, core and also surefire provider has several transitive dependencies. These transitive deps may cause conflicts, classloading issues or other problems such as failures when transitive dependencies are forbidden (using force plugin).
All dependencies that are added to the pom should be as less intrusive as possible.
Proposed solution:
* surefire provider & core should be handled by this issue: #277 
* all strategies should be shaded as well
* when the dependency is added to the effective `pom.xml` file, an exclusion for all dependencies should be set
* jar should be minimized

##### Steps To Reproduce
1. clone https://github.com/wildfly/wildfly
2. install ST
3. run build with some strategy specified

MatousJobanek:

All strategies are distributed as shaded jars and when they are added to the model, all their transitive dependencies are excluded.

- added shaded plugin configuration to all strategies
- added complete exclusion of transitive deps for all default (shaded) strategy dependencies
- introduced whole coordinate format: `groupId:artifactId:packaging:classifier:version:scope`

- [x]  not added - relocations: https://github.com/arquillian/smart-testing/issues/279#issuecomment-350393260

Fixes #279

### DIFF
--- a/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
+++ b/mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
@@ -7,8 +7,9 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.stream.Collectors;
 import org.apache.maven.model.Dependency;
-import org.apache.maven.model.Exclusion;
 import org.arquillian.smart.testing.configuration.Configuration;
+
+import static org.arquillian.smart.testing.mvn.ext.dependencies.MavenCoordinatesResolver.createDependencyFromCoordinates;
 
 /**
  * Resolves dependencies for strategies defined by keywords (e.g. new, changed, affected)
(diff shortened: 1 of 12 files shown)


mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
```diff
@@ -67,26 +67,46 @@ private Properties loadDefaultMapping() {
...
+            dependency.setClassifier(coordinates[3]);
+            dependency.setVersion(coordinates[4]);
+        }
+        if (number == 6) {
+            dependency.setScope(coordinates[5]);
+        }
+        if (dependency.getVersion() == null || dependency.getVersion().isEmpty()) {
+            dependency.setVersion(ExtensionVersion.version().toString());
+        }
+        if (excludeTrainsitive) {
```
bartoszmajsak: Typo - should be `transitive`, right? Not only in this scope, but everywhere in this class :)

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
```diff
@@ -67,26 +67,46 @@ private Properties loadDefaultMapping() {
...
-                        version = gav[2];
-                    }
-                    dependency.setVersion(version);
-                    dependency.setScope("runtime");
-                    return dependency;
-                }));
+                key -> createDependency((String) properties.get(key), excludeTrainsitive)));
+    }
+
+    private Dependency createDependency(String coordinatesString, boolean excludeTrainsitive) {
```
bartoszmajsak: Maybe we can move it to external class called `MavenCoordinatesResolver` and also provide unit tests for just this feature?

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
```diff
@@ -67,26 +67,46 @@ private Properties loadDefaultMapping() {
...
+        final String[] coordinates = coordinatesString.split(":");
+        int number = coordinates.length;
```
bartoszmajsak: `number` is quite generic, maybe `amountOfCoordinates` or even just inline.

mvn-extension/src/main/java/org/arquillian/smart/testing/mvn/ext/dependencies/StrategyDependencyResolver.java
MatousJobanek: I had the same feeling when I was writing it. But I was too lazy :-D

MatousJobanek: Ok, I moved it to separated class and also created a test with custom assertion for it

PR #288, opened 2017-12-11
Add @WatchFile annotation in affected tests so none classes change are detected too

### LINKED ISSUE #263: Add @Resource-like annotation in affected tests.
##### Issue Overview

Currently affected tests annotations only allows you to set a hard dependency with classes. For example:

```
@ComponentUnderTest(packages = "org.arquillian.smart.testing.strategies.affected.fakeproject.main.superbiz.*")
public class ZTest {
```

But would be great to add an annotation to allow specify "configuration" files as well. For example if a test must be run if `persistence.xml` file is modified or `kubernetes.json`.

For this reason I purpose to have some annotation for example called `@Resource` (with also its `@Repeatable` annotation) where you can set this can of relationship.

For example:

```
@Resource("src/main/resources/persistence.xml")
public class CarDaoTest {}
```

So if `src/main/resources/persistence.xml"` is changed then `CarDaoTest` is run too.

lordofthejars:

Add `@WatchFile` annotation in affected tests so none classes change are detected too

```
@WatchFile("src/main/resources/META-INF/p.xml")
public class WatchedFilesAnnotationTestCase {}
```

- Affected graph supports class files and none class files such as XML, JSON, ...
- `@WatchFile` annotation is created to set which modified files should make a test run.

Fixes #263


strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraph.java
```diff
@@ -86,13 +92,28 @@ void buildTestDependencyGraph(Collection<File> testJavaFiles) {
...
-                addToIndex(new JavaElement(testJavaClass), manualProductionClasses);
+
+                final List<Path> files = calculateWatchedFiles(testJavaClass);
+
+                addToIndex(new JavaElement(testJavaClass), manualProductionClasses, files);
             }
         }
     }
 
+    private List<Path> calculateWatchedFiles(JavaClass testJavaClass) {
```
bartoszmajsak: Isn't it rather `resolve` or even `find` than `calculate`?

bartoszmajsak: Isn't it rather `resolve` or even `find` than `calculate`? Same for method about `calculateManuallyAddedDependencies` - we don't really do any calculation by the pure definition of this word I believe.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraph.java
```diff
@@ -162,9 +197,24 @@ void buildTestDependencyGraph(Collection<File> testJavaFiles) {
...
     }
 
-    private void addToIndex(JavaElement javaElement, List<String> imports) {
+    private void addToIndex(JavaElement javaElement, List<String> imports, List<Path> files) {
         addToGraph(javaElement);
         updateJavaElementWithImportReferences(javaElement, imports);
+        updateJavaElementWithFiles(javaElement, files);
+    }
+
+    private void updateJavaElementWithFiles(JavaElement javaElement, List<Path> files) {
```
bartoszmajsak: How about giving it a name more related to what we do with the graph itself?

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraph.java
```diff
@@ -217,24 +269,41 @@ private boolean addImport(JavaElement javaElementParentClass, String importz) {
...
+            .map(Change::getLocation)
+            .map(this::transformToElement)
+            .filter(Optional::isPresent)
+            .map(Optional::get)
             .filter(graph::containsVertex)
             .map(this::getParents)
             .flatMap(Collection::stream)
+            .map(e -> (JavaElement) e)
             .map(JavaElement::getClassName)
             .collect(Collectors.toSet());
```
bartoszmajsak: Can't we somehow combine some operations to one? I am not sure what is the impact to performance here (I would assume consecutive `map`s are combined for example), but I'm more thinking about the readability here. We flip between domain objects (where domain is a `Change` and `JavaElement`) and `Optional`s for example.

strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ClassDependenciesGraph.java
```diff
@@ -56,26 +59,29 @@
...
     private final JavaClassBuilder builder;
-    private final DirectedGraph<JavaElement, DefaultEdge> graph;
+    private final DirectedGraph<Element, DefaultEdge> graph;
     private final Filter filter;
     private final TestVerifier testVerifier;
     private final boolean enableTransitivity;
+    private final Path projectDir;
 
-    ClassDependenciesGraph(TestVerifier testVerifier, Configuration configuration) {
+    ClassDependenciesGraph(TestVerifier testVerifier, Configuration configuration, File projectDir) {
```
bartoszmajsak: This class is growing to do too many things:

- take care of graph construction, update and travesal
- finding resources / classes to be added to the graph
- transformation between types

Would it make sense to split it to several, more focused components?

bartoszmajsak:
@lordofthejars Can we have some unit tests for these new components which could cover boundary cases?

lordofthejars:
I thought the same, but now they are acting somehow as sociable tests, in the sense that we have not added new behavior so we have just moved bits from one place to another, so all the logic is still tested. But it is true that there are some cases that we could cover there as well.

I'll add some tests.

bartoszmajsak:
That's why I mentioned boundary cases, as I know we have it covered through feature tests one level up. Thanks for your great work!

bartoszmajsak:
Both on Travis, as well as locally a test is failing with https://gist.github.com/b2f8188845cdf1d45cb12a8a1bdf526d

Notice `javassist.NotFoundException: org.jboss.arquillian.test.test.AbstractTestTestBase` (double `test.test`) - something wrong with transformation probably. Worth a reproducer test and a fix ;)

bartoszmajsak:
Please fix tests (and probably a bug) before merging :)

lordofthejars:
In my case it occurs with this exception `java.lang.IllegalArgumentException: javassist.NotFoundException: org.jboss.arquillian.core.test.AbstractManagerTestBase` so it is something new, this is happening for a change I did on recursion for finding annotations, I'll try to reproduce in unit test.

PR #291, opened 2017-12-11
feat(#287): the set of tests-to-run is provided to TestExecutionPlanner implementations

### LINKED ISSUE #287: Provide set of tests-to-run to TestExecutionPlanner implementations
##### Issue Overview

Provide set of tests-to-run to TestExecutionPlanner implementations.
There are two benefits of it
* we can optimize the execution of the planners by minimizing of the scope (eg. calculating the graph for the test classes)
* make feasible implementation of new strategies that need to retrieve some concrete information from the test classes (eg. annotations - see: #283)

##### Expected Behaviour

Replace the method 
```java
Collection<TestSelection> getTests();
```
by two methods:
```java
Collection<TestSelection> selectTestsFromNames(Iterable<String> testsToRun);

Collection<TestSelection> selectTestsFromClasses(Iterable<Class<?>> testsToRun);
```

MatousJobanek:

The implementations of the strategies didn't have the information about the set of tests that were selected to be executed. In this PR I'm introducing the changes in `TestExecutionPlanner` API that will provide this kind of information.

- in `TestExecutionPlanner`, instead of one method `getTests()` there are two methods:
```java
Collection<TestSelection> selectTestsFromNames(Iterable<String> testsToRun);
Collection<TestSelection> selectTestsFromClasses(Iterable<Class<?>> testsToRun);
```

Fixes #287

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/impl/TestStrategyApplierImpl.java
+++ b/core/src/main/java/org/arquillian/smart/testing/impl/TestStrategyApplierImpl.java
@@ -1,29 +1,13 @@
 package org.arquillian.smart.testing.impl;
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
 import java.util.Set;
-import java.util.function.BiFunction;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 import org.arquillian.smart.testing.TestSelection;
 import org.arquillian.smart.testing.api.TestStrategyApplier;
 import org.arquillian.smart.testing.configuration.Configuration;
-import org.arquillian.smart.testing.logger.Log;
-import org.arquillian.smart.testing.logger.Logger;
-import org.arquillian.smart.testing.report.SmartTestingReportGenerator;
-import org.arquillian.smart.testing.spi.TestExecutionPlanner;
 
 class TestStrategyApplierImpl implements TestStrategyApplier {
 
-    private static final Logger logger = Log.getLogger();
     private final TestExecutionPlannerLoader testExecutionPlannerLoader;
     private final File projectDir;
     private final Configuration configuration;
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/arquillian/smart/testing/impl/TestStrategyApplierImpl.java
```diff
@@ -35,16 +36,28 @@
...
+    private Collection<TestSelection> selectTestsFromClasses(TestExecutionPlanner testExecutionPlanner,
+        Iterable<Class<?>> testsToRun) {
+        return testExecutionPlanner.selectTestsFromClasses(testsToRun);
     }
 
     private <TESTCLASS> Set<TestSelection> apply(Iterable<TESTCLASS> testsToRun,
-        Function<TESTCLASS, String> mapperToName) {
-        final Set<TestSelection> selectedTests = selectTests(configuration);
+        Function<TESTCLASS, String> mapperToName,
+        BiFunction<TestExecutionPlanner, Iterable<TESTCLASS>, Collection<TestSelection>> selectionFunction) {
```
bartoszmajsak: I am not sure if using `BiFunction` here is not making it hard to follow (plus `<TESTCLASS>` "generic"). I have a feeling that this is a smell that the code here became overcomplicated (if it wasn't already).

bartoszmajsak: We have a mix of high-level, general purpose concepts such as `apply` and then within it `selectTests`. Not sure how we can make it better, but maybe a good start would be to separate `selectTests` logic from applying a strategy (so to call them one after the other instead of one from within another)?

docs/registering-strategies.adoc
```diff
@@ -33,9 +33,17 @@ include::../strategies/failed/src/main/resources/META-INF/services/org.arquillia
...
-The important method there is `getTests()` which returns a collection of the tests that must be executed.
-You also need to set there the strategy name.
+`TestExecutionPlanner` is the implementation of the strategy.
+Apart from the method `getName()` that returns a name of the strategy there are two important methods:
+
+[source, java]
+----
+selectTestsFromNames(Iterable<String> testsToRun)
+selectTestsFromClasses(Iterable<Class<?>> testsToRun)
+----
```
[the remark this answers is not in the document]
MatousJobanek: Sure, great idea

core/src/main/java/org/arquillian/smart/testing/impl/TestStrategyApplierImpl.java
MatousJobanek: Yeah, I do agree that it is overcomplicated - I wanted to avoid copy-pasting and code duplication.
I tried to refactor it, but I'm not sure if I made it better - I introduced a new abstraction - `TestSelector` that has two implementations (based on the class info type - classes/names). It does the whole magic and from the applier you call only `selectTests` or `orderTests`

bartoszmajsak: Way nicer :) Let's see if we can further improve it going forward, but this is good enough! Thank you!

bartoszmajsak:
borat-very-nice

PR #292, opened 2017-12-12
feat(#283): added categorized strategy

### LINKED ISSUE #283: Categorized strategy
##### Issue Overview

Implement `categorized` strategy that would select/order tests based on JUnit 4 `@Category` annotation.
Create appropriate logic and configuration so the strategy is able to check the classes to be run and select those that:
* has any category of the specified categories 
* has exactly the same list of categories as specified in config
* doesn't have any of the specified categories

MatousJobanek:

Introducing `categorized` strategy.  Selects tests based on JUnit 4 annotation `@Category`. This annotation has a list of classes that represent categories - eg:
`@Category(FirstCategory.class, SecondCategory.class)`

* it doesn't expect only fqn of the classes specified in the category, but a simple name can be used as well
* if no category is specified, but the strategy is used, then it selects all classes that contain the annotation, no matter which classes are specified inside of the annotations.
* it is possible to set
  * list of categories (fully qualified names or just simple class names)
  * case sensitivity (default is non-case sensitive)
  * if it should match all specified categories or when one category class is matched then it is enough (default is the second one - when one category class is matched then the test is selected)
  * reversed function - selects classes that don't match (default value is false)

Fixes #283 

- [x] write documentation

### DIFF
--- a/strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/CategoriesParser.java
+++ b/strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/CategoriesParser.java
@@ -1,0 +1,88 @@
+package org.arquillian.smart.testing.strategies.categorized;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.arquillian.smart.testing.logger.Log;
+import org.arquillian.smart.testing.logger.Logger;
+
+class CategoriesParser {
+
+    private final Logger logger = Log.getLogger();
+    private final CategorizedConfiguration strategyConfig;
+    private final List<String> specifiedCategories;
+
+    CategoriesParser(CategorizedConfiguration strategyConfig) {
+        this.strategyConfig = strategyConfig;
+        specifiedCategories = Arrays.stream(strategyConfig.getCategories())
+            .map(this::changeIfNonCaseSensitive)
+            .collect(Collectors.toList());
+    }
+
+    boolean hasCorrectCategories(Class<?> clazz) {
+        List<String> presentCategories = getPresentCategories(clazz);
+
+        if (!presentCategories.isEmpty()) {
+            if (specifiedCategories.size() > 0) {
+
+                List<String> intersection = presentCategories.stream()
+                    .filter(category -> isSpecified(category, specifiedCategories))
+                    .collect(Collectors.toList());
+
+                if (strategyConfig.isMatchAll()) {
+                    return intersection.size() == specifiedCategories.size();
+                } else {
+                    return !intersection.isEmpty();
+                }
+            }
+            return true;
+        }
+        return false;
+    }
+
+    private List<String> getPresentCategories(Class<?> clazz) {
+        return Arrays.stream(clazz.getAnnotations())
+            .filter(this::isJUnit4Category)
+            .flatMap(this::retrieveCategoriesFromAnnotation)
+            .map(this::changeIfNonCaseSensitive)
+            .collect(Collectors.toList());
+    }
+
+    private boolean isSpecified(String category, List<String> specifiedCategories) {
+        if (!specifiedCategories.contains(category)) {
+            if (category.contains(".")) {
+                String categoryClassName = category.substring(category.lastIndexOf(".") + 1);
+                return specifiedCategories.contains(categoryClassName);
+            }
+            return false;
+        }
+        return true;
+    }
+
+    private String changeIfNonCaseSensitive(String category) {
+        return strategyConfig.isCaseSensitive() ? category : category.toLowerCase();
+    }
+
+    private Stream<String> retrieveCategoriesFromAnnotation(Annotation categoryAnnotation) {
+        try {
+            Method valueMethod = categoryAnnotation.getClass().getMethod("value");
+            return Arrays.asList((Class[]) valueMethod.invoke(categoryAnnotation))
+                .stream()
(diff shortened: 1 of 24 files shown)


MatousJobanek:
> How is that different from built-in category support we use for our tests as well? Can you highlight the key reason for this feature to exist? Is it to provide support for this when our provider is used instead of JUnit one?

The benefits:
 * no need for additional executors or profiles
 * if you are fine with having profile in your pom, then with our one Category it is fine, but imagine that you have many of them and several combinations - then the pom would grow
 * in some projects it is hard to introduce some significant changes in the pom structures (such as profiles), but this is an external tool used only during test execution with no changes in the poms
 * can reuse already existing code where the categories are already set
 * having all the functionality above, you can easily combine it with other strategies

so summing this up - the main benefit is that you don't need profile :-D

MatousJobanek:
> Everything in terms of code is perfect, but I miss documentation in configuration.adoc file and refcard.adoc

Sure, that's why there is the checkbox task "write documentation" at the end of the description. I'll try to finish it tomorrow, I just wanted to know your feedback on the code.

bartoszmajsak:
Thanks - we might use this for the docs or blog announcement.

> so summing this up - the main benefit is that you don't need profile

Can't I have just one profile with an env variable I can set per build (with some default)? This way it would be similar to what we are proposing with properties too. I am not saying this is not needed, because we could mix and match strategies in the future (e.g. saying categories + affected), but I have a hard time seeing "no need for a profile" as a killer argument :)

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/CategoriesParser.java
```diff
@@ -0,0 +1,88 @@
...
+                if (strategyConfig.isMatchAll()) {
+                    return intersection.size() == specifiedCategories.size();
+                } else {
+                    return !intersection.isEmpty();
+                }
+            }
+            return true;
+        }
+        return false;
+    }
```
bartoszmajsak: This could be as well written as 

```java
final List<String> presentCategories = getPresentCategories(clazz);
        if (presentCategories.isEmpty()) {
            return false;
        }

        final List<String> intersection = presentCategories.stream()
            .filter(category -> isSpecified(category, specifiedCategories))
            .collect(Collectors.toList());

        if (strategyConfig.isMatchAll() || specifiedCategories.isEmpty()) {
            return intersection.size() == specifiedCategories.size();
        } else {
            return !intersection.isEmpty();
        }
```

Which might be a bit simpler to read (at least for me). But this makes me wonder if we shouldn't write some unit test for parser covering boundary cases if we don't cover all of them yet in the feature tests for detector.

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/CategoriesParser.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private String changeIfNonCaseSensitive(String category) {
+        return strategyConfig.isCaseSensitive() ? category : category.toLowerCase();
+    }
+
+    private Stream<String> retrieveCategoriesFromAnnotation(Annotation categoryAnnotation) {
+        try {
+            Method valueMethod = categoryAnnotation.getClass().getMethod("value");
+            return Arrays.asList((Class[]) valueMethod.invoke(categoryAnnotation))
```
bartoszmajsak: This can simply be replaced with `Arrays.stream()`

strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/CategorizedTestsDetectorTest.java
```diff
@@ -0,0 +1,137 @@
...
+        categorizedConfig.setReversed(true);
+
+        // when
+        Collection<TestSelection> testSelection =
+            new CategorizedTestsDetector(config).selectTestsFromClasses(classesToProcess);
+
+        // then
+        assertThat(testSelection).hasSize(2);
+        assertThat(SmartTesting.getClasses(new HashSet<>(testSelection)))
+            .contains(ThirdCategorizedClass.class, NonCategorizedClass.class);
```
bartoszmajsak: Instead of calling size maybe we can change those assertions to one using `containsExactly(...)`?

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/CategoriesParser.java
MatousJobanek: I see your point with the refactoring - I'll change it with your version. 

As for the test - all boundary cases should be covered by the `CategorizedTestsDetectorTest`. But I do agree that it can be confusing that there is missing the unit test for this class. The reason is that the whole logic was previously inside of the `CategorizedTestsDetector` (when I was writing the test). I moved it to separated class afterward. 
I'll create an additional unit test for it.

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/CategoriesParser.java
MatousJobanek: Yeah, right... zombie :-)

strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/CategorizedTestsDetectorTest.java
MatousJobanek: Yeah, and in case of multiple classes: `containsExactlyInAnyOrder` so I don't have to care about the order :-)

PR #297, opened 2018-01-02
fix(#285): adds missing tests scope for all tests dependencies.

### LINKED ISSUE #285: Add missing test scope to all test dependencies
##### Issue Overview

For better readability and understanding add missing `test` scope to all test dependencies in all pom files. No matter what is set in the parent pom or in dependency management - the `test` scope should be specified everywhere, otherwise, it leads to confusions.

dipak-pawar:
Adds missing tests scope for all tests dependencies.

Fixes #285

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -14,12 +14,10 @@
     <dependency>
       <groupId>org.eclipse.jgit</groupId>
       <artifactId>org.eclipse.jgit</artifactId>
-      <version>${version.jgit}</version>
     </dependency>
     <dependency>
       <groupId>org.yaml</groupId>
       <artifactId>snakeyaml</artifactId>
-      <version>${version.snakeyaml}</version>
     </dependency>
     <dependency>
       <groupId>junit</groupId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -10,6 +10,17 @@
   <artifactId>core</artifactId>
 
   <dependencies>
+    <!-- To be removed in the future when SCM is a Java SPI -->
+    <dependency>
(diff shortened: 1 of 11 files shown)


mvn-extension/pom.xml
```diff
@@ -52,19 +52,21 @@
...
       <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.github.stefanbirkner</groupId>
       <artifactId>system-rules</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
```
MatousJobanek: the version in this dependency is not necessary here - it is specified in the parent pom.

strategies/changed/pom.xml
```diff
@@ -27,6 +27,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.assertj</groupId>
```
MatousJobanek: when we have all the test dependencies that are used in several places defined in the parent pom, could you please to put the `arquillian-spacelift` dependency there as well?

core/pom.xml
```diff
@@ -13,14 +13,17 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+      <scope>test</scope>
```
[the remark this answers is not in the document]
MatousJobanek: the same thing also for the parent pom

PR #298, opened 2018-01-02
fix(#267): removes deprecated Configuration.load()

### LINKED ISSUE #267: Remove ConfigurationLoader.load()  method
##### Issue Overview

In the `ConfigurationLoader` class, there is a method [load()](https://github.com/arquillian/smart-testing/blob/master/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java#L30) that should be used only for test purposes - all calls from the implementation should provide the project directory. As it has a public visibility, then it can lead to confusions and wrong usages.  Making it deprecated is not possible as it is used across the whole project, thus this method should be removed.

dipak-pawar:

- Removes `Configuration.load()` which is used only for testing purpose.

Fixes #267

### DIFF
--- a/core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationUsingPropertyTest.java
+++ b/core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationUsingPropertyTest.java
@@ -15,6 +15,7 @@ import org.junit.experimental.categories.Category;
 import org.junit.rules.TemporaryFolder;
 
 import static java.util.Arrays.asList;
+import static org.arquillian.smart.testing.Constants.CURRENT_DIR;
 import static org.arquillian.smart.testing.RunMode.ORDERING;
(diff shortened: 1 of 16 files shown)


core/src/test/java/org/arquillian/smart/testing/configuration/ConfigurationUsingPropertyTest.java
```diff
@@ -176,7 +176,7 @@ public void should_load_configuration_with_defaults_and_with_specified_system_pr
         expectedConfiguration.setScm(scm);
 
         // when
-        final Configuration actualConfiguration = ConfigurationLoader.load();
+        final Configuration actualConfiguration = ConfigurationLoader.load(Paths.get("").toFile());
```
MatousJobanek: in these test classes, when it is used several times, wouldn't it be better to call `Paths.get("").toFile()` only once and assign it to some global variable such as `currentDir`...?

PR #303, opened 2018-01-08
fix(#294): setting setDebugLoggerLevel and useDefaultDistribution for embeddedMaven

### LINKED ISSUE #294: Change test-bed core to use latest Embedded Maven features
##### Issue Overview

In the Embedded Maven 3.0.0, there has been introduced two improvements.
* `setDebugLoggerLevel()` method that sets the invoker log level to debug. As a result, it prints the whole command that is used for Maven invocation.  [see the announcement](http://arquillian.org/blog/2017/11/13/resolver-3-0-0/). This can be used instead of [this workaround](https://github.com/arquillian/smart-testing/blob/master/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java#L73).
* there is hard-coded default Maven binary version `3.3.9` see the [doc](https://github.com/shrinkwrap/resolver#default-maven-binary). This can replace the [workaround](https://github.com/arquillian/smart-testing/blob/master/.travis.yml#L39) that we have for Travis. To use this automatically downloaded and extracted default version, it will be necessary to fix this issue first #293

dipak-pawar:

- Setting setDebugLoggerLevel embeddedMaven
- Setting useDefaultDistribution for embeddedMaven in case if you want to use maven default version.
- Updated .travis.yml 

Fixes #294

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,6 @@ before_install:
     fi'
 
 before_script:
-  - mkdir -p /home/travis/.arquillian/mvn
-  - wget https://archive.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.tar.gz -P /home/travis/.arquillian/resolver/maven/download
-  - tar -xf /home/travis/.arquillian/resolver/maven/download/apache-maven-3.3.9-bin.tar.gz -C ~/.arquillian/mvn
   # Cleanup PR doc previews - until we figure out how to have a webhook on closing the PR
   - 'if [ "${BRANCH}" == "master" ] && [ $GENERATE_DOC -eq 0 ]; then
         cd gh-pages;
@@ -60,9 +57,6 @@ before_script:
         cd ..;
       fi
   '
-
-env:
-  - TEST_BED_M2_HOME=/home/travis/.arquillian/mvn/apache-maven-3.3.9
 
 script:
   - ./mvnw clean verify -P travis -DJAVA_OPTS="-XX:-UseLoopPredicate"
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
@@ -69,8 +69,7 @@ public class ProjectBuilder {
 
         buildConfigurator.enableDebugOptions();
         setCustomMavenInstallation(embeddedMaven);
-
-        System.out.println("$ mvn " + Arrays.toString(goals).replaceAll("[\\[|\\]|,]", "") + " " + printSystemProperties());
+        embeddedMaven.setDebugLoggerLevel();
 
         final BuiltProject build = embeddedMaven
                     .setShowVersion(true)
@@ -127,10 +126,7 @@ public class ProjectBuilder {
         if (buildConfigurator.getMavenVersion() != null && !buildConfigurator.getMavenVersion().isEmpty()) {
             setMavenVersion(embeddedMaven);
         } else {
-            final String mvnInstallation = System.getenv("TEST_BED_M2_HOME");
-            if (mvnInstallation != null) {
-                embeddedMaven.useInstallation(new File(mvnInstallation));
-            }
+            embeddedMaven.useDefaultDistribution();
         }
     }
 


MatousJobanek:
Please fix the `TestNgProviderInfo` so it's not using the `StringUtils` class

.travis.yml
```diff
@@ -61,9 +58,6 @@ before_script:
       fi
   '
 
-env:
-  - TEST_BED_M2_HOME=/home/travis/.arquillian/mvn/apache-maven-3.3.9
```
bartoszmajsak: This was set because travis had older version of maven we couldn't use. It seems that since then it's using 3.5.0 on trusty image so that's a good find.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
```diff
@@ -127,10 +126,7 @@ private void setCustomMavenInstallation(PomEquippedEmbeddedMaven embeddedMaven)
         if (buildConfigurator.getMavenVersion() != null && !buildConfigurator.getMavenVersion().isEmpty()) {
             setMavenVersion(embeddedMaven);
         } else {
-            final String mvnInstallation = System.getenv("TEST_BED_M2_HOME");
-            if (mvnInstallation != null) {
-                embeddedMaven.useInstallation(new File(mvnInstallation));
-            }
+            embeddedMaven.useDefaultDistribution();
```
bartoszmajsak: I would still leave an option to change distribution - this will give us an easy way to test against different versions of maven in e.g. matrix jobs.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
```diff
@@ -67,11 +67,10 @@ private BuiltProject executeGoals(String... goals) {
             : Paths.get(rootPath, POM_XML);
         final PomEquippedEmbeddedMaven embeddedMaven = EmbeddedMaven.forProject(pomFile.toString());
 
+        embeddedMaven.setDebugLoggerLevel();
```
bartoszmajsak: What is the benefit of having it enabled by default? It slows down the build doing quite a bit of I/O in the background (even though not visible unless an embedded build failed or specifically defined through test bed config).

MatousJobanek:
So yeah, there has been an update of the apache-commons library and as part of this update the class has been moved to a different package, so it should be enough to just change the import.

MatousJobanek:
So yeah, there has been an update of the apache-commons library and as part of this update the class has been moved to a different package, so it should be enough to just change the import.

bartoszmajsak:
Do we really need to rely on 3rd party library for such a basic thing though?

.travis.yml
MatousJobanek: Yeah, but the SWR is now using 3.5.2, that is not installed on the travis. Using the `embeddedMaven.useDefaultDistribution();` will guarantee that we are using the latest one SWR is compatible with

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
MatousJobanek: the debug log level is for MavenInvoker, not for the Maven build itself. 
The benefit is that it prints the whole command that is used for invoking the build. If you run it, there is almost nothing else printed so no slowdown is expected.
see: https://github.com/shrinkwrap/resolver#maven-build-invocation-command--logging

.travis.yml
bartoszmajsak: So we cannot simply use anything lower than `3.5.x`? We might want to explore some different strategy to handle multiple mvn versions build in the future then.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
bartoszmajsak: Thanks for the reference. Naming got me confused. If so I'm totally fine with that.

.travis.yml
MatousJobanek: if your comment https://github.com/arquillian/smart-testing/pull/303#discussion_r162564945 is addressed, then sure, you can. 
see these lines - there is set a different Maven version https://github.com/arquillian/smart-testing/blob/master/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java#L127-L129

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
MatousJobanek: Good idea with it - I didn't realize some matrix jobs. Currently, different Maven versions are set using the API: https://github.com/arquillian/smart-testing/blob/master/functional-tests/test-bed/src/test/java/org/arquillian/smart/testing/ftest/configuration/DebugModeAndMaven325FunctionalTest.java#L61
On the other hand - these corner cases and supports for older versions should be tested in every build and having another job could just slow down the test suite execution. That's why we have this particular test - to verify backward compatibility.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
MatousJobanek: Yeah, the name can be confusing. I've explained the behavior in the Javadoc of this method.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
bartoszmajsak: I think having both options - global and per-test are valuable. Do we agree on leaving them then?

MatousJobanek:
> Do we really need to rely on 3rd party library for such a basic thing though?

because when I was writing it (copying from the surefire code) I didn't care so I used the library as it was on the classpath. So yeah, probably worth rewriting it - the code won't be so difficult :-)

bartoszmajsak:
> > Do we really need to rely on 3rd party library for such a basic thing though?

> because when I was writing it (copying from the surefire code) I didn't care so I used the library as it was on the classpath. So yeah, probably worth rewriting it - the code won't be so difficult :-)

Missing threads here... anyway - let's open an issue and work on it as a separated tasks if changing the import fixes this one.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/project/ProjectBuilder.java
MatousJobanek: Yeah, I'm fine with some system property (providing the required Maven version) that would allow us to test it in some matrix jobs.
so +1

MatousJobanek:
> Missing threads here... anyway - let's open an issue and work on it as a separated tasks if changing the import fixes this one.

it's up to @dipak-pawar 
I've just noticed that I'll need to release another version of SWR. The updates of Maven library brought some other 3rd party libs that are not reallocated... My bad. I'll do it today and ping @dipak-pawar when it's done.

dipak-pawar:
@MatousJobanek I will add it(`isNumeric`) as a part of this PR. No need to create separate issue for it.

bartoszmajsak:
The idea was more to have traceable chunks of changes, but sure, that works too, as it was revealed in this PR.

bartoszmajsak:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project arquillian-testng-core: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test failed: There was an error in the forked process
[ERROR] java.lang.NullPointerException
[ERROR] 	at org.arquillian.smart.testing.surefire.provider.info.TestNgProviderInfo.isNumeric(TestNgProviderInfo.java:41)
[ERROR] 	at org.arquillian.smart.testing.surefire.provider.info.TestNgProviderInfo.convertProviderParameters(TestNgProviderInfo.java:34)
[ERROR] 	at org.arquillian.smart.testing.surefire.provider.SurefireProviderFactory.createInstance(SurefireProviderFactory.java:37)
[ERROR] 	at org.arquillian.smart.testing.surefire.provider.SmartTestingSurefireProvider.invoke(SmartTestingSurefireProvider.java:63)
[ERROR] 	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:290)
[ERROR] 	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:242)
[ERROR] 	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:121)
[ERROR] 
```

Let's fix it by adding unit tests for this function @dipak-pawar.

bartoszmajsak:
@dipak-pawar any idea what that could be? https://travis-ci.org/arquillian/smart-testing/builds/331710958#L2842

dipak-pawar:
@bartoszmajsak Not sure about it. I tried it locally as well. Couldn't reproduce it

PR #307, opened 2018-01-10
feat(#301): Add support for Tag and Tags annotation from JUnit 5

### LINKED ISSUE #301: Support @Tag annotation from JUnit5 in category strategy
##### Issue Overview

Support `@Tag` annotation from JUnit5 in category strategy (http://junit.org/junit5/docs/current/user-guide/#writing-tests-tagging-and-filtering)

##### Expected Behaviour

Support `@Tag` annotation to categorize tests

##### Current Behaviour

Only JUnit4 annotations are supported.

lordofthejars:

Add support for `Tag` and `Tags` annotation from JUnit 5

- Add scanning support for `@Tag` and `@Tags` annotations
- Add support for meta annotations annotated with `@Tag`.

Fixes #301


lordofthejars:
Before merging I'll update the docs.

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
```diff
@@ -0,0 +1,127 @@
...
+    private final Logger logger = Log.getLogger();
+    private final CategorizedConfiguration strategyConfig;
+    private final List<String> specifiedCategories;
+
+    TagsParser(CategorizedConfiguration strategyConfig) {
+        this.strategyConfig = strategyConfig;
+        specifiedCategories = Arrays.stream(strategyConfig.getCategories())
+            .map(this::changeIfNonCaseSensitive)
+            .collect(Collectors.toList());
+    }
```
MatousJobanek: How about creating a new abstract class `Parser` with the common logic??

MatousJobanek: The method `hasCorrectTags` is pretty similar to `hasCorrectCategories` - that could be inside of the abstract class as well.

strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/project/FastTaggedClass.java
```diff
@@ -0,0 +1,5 @@
+package org.arquillian.smart.testing.strategies.categorized.project;
```
MatousJobanek: it would be nice to create sub-packages for tagged classes and categorized classes, otherwise, it's quite difficult to orientate there.

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
```diff
@@ -0,0 +1,107 @@
...
+
+    TagsParser(CategorizedConfiguration strategyConfig) {
+        super(strategyConfig);
+    }
+
+    @Override
+    protected List<String> findCategories(Class<?> clazz) {
+
+        return Arrays.stream(clazz.getAnnotations())
+            .map(annotation -> this.findJUnit5TagAnnotation(annotation))
```
dipak-pawar: This can be replaced with method reference. `this::findJUnit5TagAnnotation`

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.arquillian.smart.testing.logger.Log;
+import org.arquillian.smart.testing.logger.Logger;
+
+class TagsParser extends AbstractParser {
+
+    public static final String TAG = "org.junit.jupiter.api.Tag";
+    public static final String TAGS = "org.junit.jupiter.api.Tags";
```
dipak-pawar: Can't we make `TAG` and `TAGS` private?

strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/project/categories/FirstCategorizedClass.java
```diff
@@ -1,4 +1,4 @@
-package org.arquillian.smart.testing.strategies.categorized.project;
+package org.arquillian.smart.testing.strategies.categorized.project.categories;
 
 import org.arquillian.smart.testing.strategies.categorized.project.categories.FirstCategory;
```
dipak-pawar: no need to import

strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/project/categories/FirstAndSecondCategorizedClass.java
```diff
@@ -1,4 +1,4 @@
-package org.arquillian.smart.testing.strategies.categorized.project;
+package org.arquillian.smart.testing.strategies.categorized.project.categories;
 
 import org.arquillian.smart.testing.strategies.categorized.project.categories.FirstCategory;
 import org.arquillian.smart.testing.strategies.categorized.project.categories.SecondCategory;
```
dipak-pawar: no need to import line 3 and 4

strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/project/categories/ThirdCategorizedClass.java
```diff
@@ -1,4 +1,4 @@
-package org.arquillian.smart.testing.strategies.categorized.project;
+package org.arquillian.smart.testing.strategies.categorized.project.categories;
 
 import org.arquillian.smart.testing.strategies.categorized.project.categories.ThirdCategory;
```
dipak-pawar: no need to import

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
```diff
@@ -0,0 +1,107 @@
...
+        return specifiedCategories.contains(category);
+    }
+
+    private Stream<String> retrieveTagFromAnnotation(Annotation tagsAnnotation) {
+        final List<String> tags = new ArrayList<>();
+        if (TAG.equals(tagsAnnotation.annotationType().getName())) {
+            tags.add(getTagName(tagsAnnotation));
+        } else {
+            tags.addAll(getMultipleTagNames(tagsAnnotation));
+        }
```
dipak-pawar: I can see `getTagName` and `getMultipleTagNames` returning null. I think we should check null value before adding it to list.

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
```diff
@@ -0,0 +1,107 @@
...
+
+        } catch (Exception e) {
+            logger.warn("Something wrong happened when the annotation [%s] was being parsed: %s", tagAnnotation,
+                e.getMessage());
+        }
+        return null;
+    }
+
+    private Annotation findJUnit5TagAnnotation(Annotation annotation) {
+        String fgn = annotation.annotationType().getName();
```
dipak-pawar: Can we have better name for `fgn`

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
lordofthejars: yeah they are public because they were needed in previous design and I miss to change. Ty

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
lordofthejars: Only `getTagName` can return null although is unlikely that happens because of the `if` protection, it is better to protect for just in case.

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private List<String> getMultipleTagNames(Annotation tagsAnnotation) {
+        final List<String> tags = new ArrayList<>();
+        try {
+            Method valueMethod = tagsAnnotation.getClass().getMethod("value");
+            final Annotation[] tagAnnotations = (Annotation[]) valueMethod.invoke(tagsAnnotation);
+            tags.addAll(
+                Arrays.stream(tagAnnotations)
+                .filter(Objects::nonNull)
+                .map(this::getTagName)
```
dipak-pawar: We have to filter it on `null` as `getTagName` may return null to be on safe side

lordofthejars: It is filtered in previous `.filter` call

lordofthejars:
@dipak-pawar I am not sure that adds any benefit adding a ftest for this use case. Notice that the change is internal, the logic is the same,the only difference is that now we are taking `@Categories` or `@Tags` but then the logic is exactly the same. And this extracting logic is fully covered with unit tests.

In my opinion adding a new ftest means adding extra 15 seconds on build which are not necessary at all for this case.

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
dipak-pawar: I can see `getTagName` is returning null due to some exceptions. So if it returns null how previous `.filter` will help in this scenario.

lordofthejars: right sorry I was thinking in the other instance

PR #308, opened 2018-01-16
feat(#305): added support for test method selection

### LINKED ISSUE #305: Support selection of test methods
##### Issue Overview

Currently, Smart Testing selection is working only at the test class level. It should also support the selection of test methods by saying which test methods should be executed.
Change ST API (mainly `TestSelection` class) and inner logic to support this feature.
To figure out how to tell surefire provider which test methods can be executed and which not we can check how it is done in the case of [Running a Set of Methods in a Single Test Class](http://maven.apache.org/surefire/maven-surefire-plugin/examples/single-test.html)

MatousJobanek:

This PR brings a support for selecting particular test methods (not only the whole test classes).
The solution is using `TestListResolver` - this class is used by the providers to check and filter any method pattern specified by a user. Unfortunately, it is an immutable object set as a private field inside of another immutable object, so the only way to make it working is using a Java reflection.

This PR contains only a support in Core, next PR introducing a resolution of categories on test methods will show and test a real usage.

- new variable `testMethodNames` in `TestSelection` containing a list of test method names that should be selected. if the list is empty, then the whole class should be executed
- in `TestSelection` class has been also renamed variable `types` to `appliedStrategies`
- in `SmartTestingSurefireProvider` class, there has been introduced `SmartTestingInvoker` that invokes Smart Testing API - this change is because of unit tests - to mock SP API
- renamed  `bootParams` to `providerParameters` to reflect the type of this object

Fixes #305

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/TestSelection.java
+++ b/core/src/main/java/org/arquillian/smart/testing/TestSelection.java
@@ -4,8 +4,8 @@ import java.lang.reflect.Array;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.LinkedHashSet;
-import java.util.List;
 import java.util.Objects;
 
 import static java.lang.String.format;
@@ -18,7 +18,7 @@ public class TestSelection {
 
     private final Collection<String> appliedStrategies;
 
-    private final List<String> testMethodNames;
+    private final Collection<String> testMethodNames;
 
     public static final TestSelection NOT_MATCHED = new TestSelection("");
 
@@ -27,9 +27,13 @@ public class TestSelection {
     }
 
     public TestSelection(String className, String ... appliedStrategies) {
+        this(className, new ArrayList<>(), appliedStrategies);
+    }
+
+    public TestSelection(String className, Collection<String> testMethodNames, String ... appliedStrategies) {
         this.className = className;
         this.appliedStrategies = new LinkedHashSet<>(asList(appliedStrategies));
-        this.testMethodNames = new ArrayList<>();
+        this.testMethodNames = testMethodNames;
     }
 
     public String getClassName() {
@@ -37,7 +41,11 @@ public class TestSelection {
     }
 
     public Collection<String> getAppliedStrategies() {
-        return appliedStrategies; // TODO should we return clone to avoid manipulation?
+        return Collections.unmodifiableCollection(appliedStrategies);
+    }
+
+    public Collection<String> getTestMethodNames() {
+        return Collections.unmodifiableCollection(testMethodNames);
     }
 
     @Override
@@ -80,9 +88,5 @@ public class TestSelection {
         arraycopy(first, 0, result, 0, first.length);
         arraycopy(second, 0, result, first.length, second.length);
         return result;
-    }
-
-    public List<String> getTestMethodNames() {
-        return testMethodNames;
     }
 }
--- a/core/src/main/java/org/arquillian/smart/testing/TestSelection.java
+++ b/core/src/main/java/org/arquillian/smart/testing/TestSelection.java
@@ -2,8 +2,10 @@ package org.arquillian.smart.testing;
 
 import java.lang.reflect.Array;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Objects;
 
 import static java.lang.String.format;
@@ -14,23 +16,28 @@ public class TestSelection {
 
     private final String className;
 
-    private final Collection<String> types; // TODO or strategy instead (misleading name)
+    private final Collection<String> appliedStrategies;
 
-    public TestSelection(Path location, String... type) {
-        this(new ClassNameExtractor().extractFullyQualifiedName(location), type);
+    private final List<String> testMethodNames;
+
+    public static final TestSelection NOT_MATCHED = new TestSelection("");
+
+    public TestSelection(Path location, String... appliedStrategies) {
+        this(new ClassNameExtractor().extractFullyQualifiedName(location), appliedStrategies);
     }
 
-    public TestSelection(String className, String ... type) {
+    public TestSelection(String className, String ... appliedStrategies) {
         this.className = className;
-        this.types = new LinkedHashSet<>(asList(type));
+        this.appliedStrategies = new LinkedHashSet<>(asList(appliedStrategies));
+        this.testMethodNames = new ArrayList<>();
     }
 
     public String getClassName() {
         return className;
     }
 
-    public Collection<String> getTypes() {
-        return types; // TODO should we return clone to avoid manipulation?
+    public Collection<String> getAppliedStrategies() {
+        return appliedStrategies; // TODO should we return clone to avoid manipulation?
     }
 
     @Override
@@ -52,7 +59,7 @@ public class TestSelection {
 
     @Override
     public String toString() {
-        return "TestSelection{" + "className='" + className + '\'' + ", types=" + types + '}';
+        return "TestSelection{" + "className='" + className + '\'' + ", appliedStrategies=" + appliedStrategies + '}';
     }
 
     public TestSelection merge(TestSelection other) {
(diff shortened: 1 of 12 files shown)


bartoszmajsak:
Why not?

lordofthejars:
Because Matous told me that this change can break che integration, so I'd
like to have a latest version before this merge (thinking in my talk for
checonf :D)

MatousJobanek:
> Because Matous told me that this change can break che integration

@lordofthejars not this, there has been a change in ST API one release ago ;-). This won't affect anything

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
```diff
@@ -14,23 +16,28 @@
...
     private final String className;
 
-    private final Collection<String> types; // TODO or strategy instead (misleading name)
+    private final Collection<String> appliedStrategies;
 
-    public TestSelection(Path location, String... type) {
-        this(new ClassNameExtractor().extractFullyQualifiedName(location), type);
+    private final List<String> testMethodNames;
+
+    public static final TestSelection NOT_MATCHED = new TestSelection("");
```
dipak-pawar: I can't see, where is it used?

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
```diff
@@ -14,23 +16,28 @@
...
+    public TestSelection(Path location, String... appliedStrategies) {
+        this(new ClassNameExtractor().extractFullyQualifiedName(location), appliedStrategies);
     }
 
-    public TestSelection(String className, String ... type) {
+    public TestSelection(String className, String ... appliedStrategies) {
         this.className = className;
-        this.types = new LinkedHashSet<>(asList(type));
+        this.appliedStrategies = new LinkedHashSet<>(asList(appliedStrategies));
+        this.testMethodNames = new ArrayList<>();
```
dipak-pawar: Can you point me the place where we are updating this list of `testMethodNames`?

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
MatousJobanek: In a real implementation - check this PR: https://github.com/arquillian/smart-testing/pull/309
https://github.com/arquillian/smart-testing/pull/309/commits/e6314b83a7a4756d23b566994d9f99edc02934d9#diff-b179c0d6894a22602721a54a8efc1d54R33

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
MatousJobanek: https://github.com/arquillian/smart-testing/pull/309/commits/e6314b83a7a4756d23b566994d9f99edc02934d9#diff-b179c0d6894a22602721a54a8efc1d54R49

dipak-pawar: I got confused as it is part of another PR. Thanks

MatousJobanek: Sure, I wanted to keep it separated - the implementation of the core from the code of other strategies, otherwise, it would be one big confusing PR mixing several things together.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SecurityUtils.java
```diff
@@ -81,4 +82,27 @@
             }
         }
     }
+
+    public static Field getField(final Class<?> source, final String name) {
+        Field declaredAccessibleField = AccessController.doPrivileged(new PrivilegedAction<Field>() {
```
dipak-pawar: We can return directly here instead of assigning it here and returning on next statement. Also we can use lambda here

MatousJobanek: Yeah, copy-paste from the arquillian-core

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
bartoszmajsak: According to IDEA nowhere. Probably some left-over @MatousJobanek?

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
bartoszmajsak: Again, IDEA is your friend. It's used in the tests. https://github.com/arquillian/smart-testing/blob/468c81b33b72f78eabe053b9f6fb6d68bb3916eb/surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/SmartTestingProviderTest.java#L142

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
```diff
@@ -14,23 +16,28 @@
...
     }
 
     public String getClassName() {
         return className;
     }
 
-    public Collection<String> getTypes() {
-        return types; // TODO should we return clone to avoid manipulation?
+    public Collection<String> getAppliedStrategies() {
+        return appliedStrategies; // TODO should we return clone to avoid manipulation?
```
bartoszmajsak: So what do we do? (`TODO`)

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
```diff
@@ -74,4 +81,8 @@ public TestSelection merge(TestSelection other) {
         arraycopy(second, 0, result, first.length, second.length);
         return result;
     }
+
+    public List<String> getTestMethodNames() {
+        return testMethodNames;
```
bartoszmajsak: Same here, should this be available for modification outside of this class?

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
dipak-pawar: that's correct, but I didn't found real logic to filter testMethods

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
MatousJobanek: Yeah, sorry for that - I should have put it into the next PR - where it is really used. I just introduced it along with the other changes, so using the command `git add core/*` I put it into this PR...

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
bartoszmajsak: I missed followup conversation in the meantime while reviewing PRs, so my comment can be ignored :)

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
```diff
@@ -129,4 +169,26 @@ private boolean isGetConsoleLoggerMethod(Method method) {
...
+        TestListResolver originalResolver = providerParameters.getTestRequest().getTestListResolver();
+        TestListResolver selectedResolver = new TestListResolver(selectedTestMethods);
+
+        TestListResolver newTestListResolver =
+            TestListResolver.newTestListResolver(
+                selectedResolver.getIncludedPatterns(),
+                originalResolver.getExcludedPatterns());
+
+        try {
+            Field requestedTestsField = SecurityUtils.getField(TestRequest.class, "requestedTests");
```
bartoszmajsak: I understand this is the only way to bring this functionality to our tool. I'm a bit concerned - we should be careful using reflection as this can bite us when Surefire evolves as well as when finally adding JDK 9 support.

@MatousJobanek would it make sense to at least open JIRA in Surefire project asking about a possibility to opening this API?

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
MatousJobanek: I haven't decided yet :-D. 
To be honest, I don't see any potential use case when it could cause some problems. We don't expose this instance of `TestSelection` class to anyone, we just take it from the strategies and merge them internally. 
Correct me if I'm wrong.

surefire-provider/src/main/java/org/arquillian/smart/testing/surefire/provider/SmartTestingSurefireProvider.java
MatousJobanek: Yeah JDK 9 is my nightmare.

I'll open an issue for it in the Surefire project

core/src/main/java/org/arquillian/smart/testing/TestSelection.java
MatousJobanek: WDYT?

PR #309, opened 2018-01-16
 feat(304): categorized strategy is resolving also categories specified on test methods

### LINKED ISSUE #304: Categorized - support categorization for test methods
##### Issue Overview

Currently, only the categories of the whole test classes are retrieved. The goal of this task is to categorize also the test methods.
By default, this feature should be enabled, but the configuration should support disabling categorization of the test methods and keep only for test classes.

MatousJobanek:

This is a specific implementation of test method selection for the `categorized` strategy. It is built on top of the implementation proposed in this PR #308 

The resolution works for both JUnit 4 categories and JUnit 5 tags.

- added `smart.testing.categorized.methods` config parameter to disable/enable method recognition (default is `true`)
- changed `AbstractParser` to get categories/tags annotated on test methods; added abstract method `isTestMethod` to check which method is a test method.
- test-bed - slightly modified test result parser to get results on the test method level

#### Missing:
- [x] documentation

Fixes #304

### DIFF
--- a/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/testresults/TestResult.java
+++ b/functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/testresults/TestResult.java
@@ -44,7 +44,24 @@ public class TestResult {
             throw new IllegalArgumentException(
                 "Too many arguments passed to create TestResult. Expected to get status code and name, got [" + Arrays.toString(testResultParts) + "]");
         }
-        return new TestResult(testResultParts[1], "*", Status.from(testResultParts[0]));
+
+        String className;
+        String methodName;
+
+        if (expectedTestContainsSpecificMethodDefinition(testResultParts[1])) {
+            String[] classWithMethod = testResultParts[1].split("#");
+            className = classWithMethod[0];
+            methodName = classWithMethod[1];
+
+        } else {
+            className = testResultParts[1];
+            methodName = "*";
+        }
+        return new TestResult(className, methodName, Status.from(testResultParts[0]));
+    }
+
+    private static boolean expectedTestContainsSpecificMethodDefinition(String expectedTest) {
+        return expectedTest.contains("#");
     }
 
     @Override
--- a/strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
+++ b/strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
@@ -6,7 +6,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.arquillian.smart.testing.logger.Log;
@@ -23,9 +22,9 @@ class TagsParser extends AbstractParser {
     }
 
     @Override
-    protected List<String> findCategories(Class<?> clazz) {
+    protected List<String> findCategories(Annotation[] annotations) {
 
-        return Arrays.stream(clazz.getAnnotations())
+        return Arrays.stream(annotations)
             .map(this::findJUnit5TagAnnotation)
             .filter(Objects::nonNull)
             .flatMap(this::retrieveTagFromAnnotation)
@@ -106,6 +105,12 @@ class TagsParser extends AbstractParser {
 
         return annotation;
 
+    }
+
+    @Override
+    protected boolean isTestMethod(Method method) {
+        return Arrays.stream(method.getAnnotations())
+            .anyMatch(annotation -> annotation.annotationType().getName().equals("org.junit.jupiter.api.Test"));
     }
 }
 
--- a/strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/CategorizedTestsDetectorTest.java
+++ b/strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/CategorizedTestsDetectorTest.java
@@ -2,42 +2,47 @@ package org.arquillian.smart.testing.strategies.categorized;
 
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashSet;
 import java.util.List;
 import org.arquillian.smart.testing.TestSelection;
-import org.arquillian.smart.testing.api.SmartTesting;
 import org.arquillian.smart.testing.configuration.Configuration;
 import org.arquillian.smart.testing.configuration.ConfigurationLoader;
-import org.arquillian.smart.testing.strategies.categorized.project.tags.FastTaggedClass;
 import org.arquillian.smart.testing.strategies.categorized.project.categories.FirstAndSecondCategorizedClass;
-import org.arquillian.smart.testing.strategies.categorized.project.tags.FirstAndSecondTaggedClass;
 import org.arquillian.smart.testing.strategies.categorized.project.categories.FirstCategorizedClass;
-import org.arquillian.smart.testing.strategies.categorized.project.tags.FirstTaggedClass;
-import org.arquillian.smart.testing.strategies.categorized.project.categories.NonCategorizedClass;
-import org.arquillian.smart.testing.strategies.categorized.project.tags.NonTaggedClass;
-import org.arquillian.smart.testing.strategies.categorized.project.categories.ThirdCategorizedClass;
-import org.arquillian.smart.testing.strategies.categorized.project.tags.ThirdTaggedClass;
 import org.arquillian.smart.testing.strategies.categorized.project.categories.FirstCategory;
+import org.arquillian.smart.testing.strategies.categorized.project.categories.NonCategorizedClass;
 import org.arquillian.smart.testing.strategies.categorized.project.categories.SecondCategory;
+import org.arquillian.smart.testing.strategies.categorized.project.categories.ThirdCategorizedClass;
+import org.arquillian.smart.testing.strategies.categorized.project.categories.WithCategorizedMethodsClass;
+import org.arquillian.smart.testing.strategies.categorized.project.tags.FastTaggedClass;
+import org.arquillian.smart.testing.strategies.categorized.project.tags.FirstAndSecondTaggedClass;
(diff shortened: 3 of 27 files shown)


strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
```diff
@@ -107,5 +106,11 @@ private Annotation findJUnit5TagAnnotation(Annotation annotation) {
         return annotation;
 
     }
+
+    @Override
+    protected boolean isTestMethod(Method method) {
+        return Arrays.stream(method.getAnnotations())
+            .anyMatch(annotation -> annotation.annotationType().getName().equals("org.junit.jupiter.api.Test"));
```
lordofthejars: I usually refer to do `"org.junit.....".equals(instance)` to avoid NPE

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/testresults/TestResult.java
```diff
@@ -44,7 +44,12 @@ public static TestResult from(String ... testResultParts) {
             throw new IllegalArgumentException(
                 "Too many arguments passed to create TestResult. Expected to get status code and name, got [" + Arrays.toString(testResultParts) + "]");
         }
-        return new TestResult(testResultParts[1], "*", Status.from(testResultParts[0]));
+        String[] classWithMethod = testResultParts[1].split("#");
+        String methodName = "*";
+        if (classWithMethod.length == 2) {
+            methodName = classWithMethod[1];
+        }
+        return new TestResult(classWithMethod[0], methodName, Status.from(testResultParts[0]));
```
lordofthejars: Could we create a constant to know exactly what is 0 and 1?

strategies/categorized/src/main/java/org/arquillian/smart/testing/strategies/categorized/TagsParser.java
MatousJobanek: Yeah, in the cases when the NPE could occur, I use the same format. In this case, I don't expect that the name of any annotation type could be null :-)
But I can change it if you want...

lordofthejars: No worries, in fact now that you mention I think that in the annotation is impossible to set a null.

functional-tests/test-bed/src/main/java/org/arquillian/smart/testing/ftest/testbed/testresults/TestResult.java
MatousJobanek: constants? Do you mean something like:
```java
    private static final int POSITION_OF_TEST_CLASS_NAME = 0;
    private static final int POSITION_OF_TEST_METHOD_NAME = 1;

    public static TestResult from(String ... testResultParts) {
        if (testResultParts.length > 2) {
            throw new IllegalArgumentException(
                "Too many arguments passed to create TestResult. Expected to get status code and name, got [" + Arrays.toString(testResultParts) + "]");
        }
        String[] classWithMethod = testResultParts[1].split("#");
        String methodName = "*";
        if (classWithMethod.length == 2) {
            methodName = classWithMethod[POSITION_OF_TEST_METHOD_NAME];
        }
        return new TestResult(classWithMethod[POSITION_OF_TEST_CLASS_NAME], methodName, Status.from(testResultParts[0]));
    }
```
This doesn't seem to be very meaningfull from my point of view - this also brings a complication as there is another array `testResultParts`...

To say what should be inside of the array, I named it: `classWithMethod`
in addition, I assign the `classWithMethod[1]` to `methodName` which says what is expected there. 
I could also assign the first variable: 
```java
        String className = classWithMethod[0];
        if (classWithMethod.length == 2) {
            methodName = classWithMethod[1];
        }
        return new TestResult(className, methodName, Status.from(testResultParts[0]));
```
which is slightly redudant, but I can do it

lordofthejars: I think the first approach is good  setting the const as static ints with a name, but maybe creating a method for the if might be good and help to understand the method so instead of `clasWithMethod.length==2` do something like `if(classContainsMethod(classWithMethod))`

MatousJobanek: using the constants for the positions was irony... I would need to name it something like:
`POSITION_OF_TEST_CLASS_NAME_IN_CLASS_WITH_METHOD_ARRAY` because there is another array `testResultParts` which is something that I don't really like...

Does this make more sense to you?
```java
    public static TestResult from(String ... testResultParts) {
        if (testResultParts.length > 2) {
            throw new IllegalArgumentException(
                "Too many arguments passed to create TestResult. Expected to get status code and name, got [" + Arrays.toString(testResultParts) + "]");
        }

        String className;
        String methodName;

        if (expectedTestContainsSpecificMethodDefinition(testResultParts[1])) {
            String[] classWithMethod = testResultParts[1].split("#");
            className = classWithMethod[0];
            methodName = classWithMethod[1];

        } else {
            className = testResultParts[1];
            methodName = "*";
        }
        return new TestResult(className, methodName, Status.from(testResultParts[0]));
    }

    private static boolean expectedTestContainsSpecificMethodDefinition(String expectedTest) {
        return expectedTest.contains("#");
    }
```

lordofthejars: ah yeah now better I guess

MatousJobanek: Done

strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/CategorizedTestsDetectorTest.java
```diff
@@ -167,7 +182,41 @@ public void should_return_classes_without_first_category_when_reversed_is_used()
...
         // then
-        assertThat(SmartTesting.getClasses(new HashSet<>(testSelection)))
-            .containsExactlyInAnyOrder(ThirdCategorizedClass.class, NonCategorizedClass.class);
+        TestSelection testSelectionWithMethods =
+            getTestSelectionWithMethods(WithCategorizedMethodsClass.class, "testWithSecondCategory",
+                "testWithThirdCategory", "testWithoutCategory");
+        verifyContent(testSelection, testSelectionWithMethods, ThirdCategorizedClass.class, NonCategorizedClass.class);
+    }
+
+    private void verifyContent(Collection<TestSelection> actualTestSelections, TestSelection expectedWithMethodsSelection,
```
hemanik: Just a suggestion, isn't this also a good candidate for the custom assertions?

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/SmartTestingProviderTest.java
```diff
@@ -123,12 +135,67 @@ public void test_when_invoke_is_called_with_whole_set_of_classes() throws Except
...
+
+        // when
+        provider.invoke(null);
+
+        // then
+        softly.assertThat(testRequest.getTestListResolver())
+            .hasMethodPatterns(true)
+            .hasIncludedMethodPatterns(true)
+            .hasExcludedMethodPatterns(true)
+            .includedPatterns()
```
hemanik: How about renaming `includedPatterns()` to something like `hasIncludedPatterns ()` ?

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/SmartTestingProviderTest.java
```diff
@@ -123,12 +135,67 @@ public void test_when_invoke_is_called_with_whole_set_of_classes() throws Except
...
+            .containsExactlyInAnyOrder(
+                new ResolvedTest(ATest.class.getName(), "firstMethod", false),
+                new ResolvedTest(ATest.class.getName(), "secondMethod", false));
+
+        softly.assertThat(testRequest.getTestListResolver())
+            .excludedPatterns()
```
hemanik: How about renaming `excludedPatterns()` to something like `hasExcludedPatterns ()` ?

strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/CategorizedTestsDetectorTest.java
MatousJobanek: I was thinking about it, unfortunately, this is a verification of a collection with elements of a specific type. Collections are quite tricky - in this case as you can just provide that the assertion is for a collection, but you cannot specify the type of the elements. As AssertJ provides an assertion for a collection already, so it could be quite confusing...
On the other hand, why not - it will be used also in the failed strategy, so it makes sense. So I'll create `TestSelectionCollectionAssert` for it.

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/SmartTestingProviderTest.java
MatousJobanek: `includedPatterns()` doesn't verify anything, it just returns an instance of an asserstion for the collection: `Assertions.assertThat(actual.getExcludedPatterns());`

surefire-provider/src/test/java/org/arquillian/smart/testing/surefire/provider/SmartTestingProviderTest.java
MatousJobanek: same as `includedPatterns()`

strategies/categorized/src/test/java/org/arquillian/smart/testing/strategies/categorized/CategorizedTestsDetectorTest.java
MatousJobanek: Done :-)

PR #312, opened 2018-01-31
feat(#306): failed strategy selects the particular failed test methods

### LINKED ISSUE #306: Failed - support execution of the specific test methods that failed
##### Issue Overview

Currently, the `failed` strategy executes the whole test class even in the cases when there is only one test method that failed. Provide an opportunity to execute only the specific test methods, not the whole test class.
By default, this feature should be enabled and disabled via a configuration parameter.

- [x] change junit-test-result-parser to parse the result on the test method level (if necessary)
- [x] modify logic of the failed strategy accordingly

MatousJobanek:

Currently, the failed strategy selects the whole test classes no matter how many of the test methods failed in the previous run. This PR brings the support for selecting particular test methods that failed. By default, it is enabled - can be disabled using config property. 

- To reuse the `TestSelectionCollectionAssert` I've moved it to the `core` module and packing it inside of the test-jar

Fixes #306

### DIFF
--- a/strategies/categorized/pom.xml
+++ b/strategies/categorized/pom.xml
@@ -31,6 +31,13 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.arquillian.smart.testing</groupId>
+      <artifactId>core</artifactId>
+      <version>${project.version}</version>
+      <classifier>tests</classifier>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/strategies/categorized/pom.xml
+++ b/strategies/categorized/pom.xml
@@ -34,7 +34,6 @@
     <dependency>
       <groupId>org.arquillian.smart.testing</groupId>
       <artifactId>core</artifactId>
-      <version>${project.version}</version>
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
--- a/strategies/failed/pom.xml
+++ b/strategies/failed/pom.xml
@@ -36,6 +36,13 @@
       <artifactId>junit-test-result-parser</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.arquillian.smart.testing</groupId>
+      <artifactId>core</artifactId>
+      <version>${project.version}</version>
+      <classifier>tests</classifier>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/strategies/failed/pom.xml
+++ b/strategies/failed/pom.xml
@@ -39,7 +39,6 @@
     <dependency>
       <groupId>org.arquillian.smart.testing</groupId>
       <artifactId>core</artifactId>
-      <version>${project.version}</version>
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
--- a/strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/TestResultsFilter.java
+++ b/strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/TestResultsFilter.java
@@ -9,7 +9,7 @@ import java.util.stream.Collectors;
 import org.arquillian.smart.testing.TestSelection;
 import org.arquillian.smart.testing.spi.TestResult;
 
-public class TestResultsFilter {
+class TestResultsFilter {
 
     static Collection<TestSelection> getFailedTests(FailedConfiguration strategyConfig, Set<TestResult> testResults) {
 
--- a/strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/TestResultsFilter.java
+++ b/strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/TestResultsFilter.java
@@ -1,0 +1,34 @@
+package org.arquillian.smart.testing.strategies.failed;
+
+import java.util.Arrays;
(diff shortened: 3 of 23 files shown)


strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/TestResultsFilter.java
```diff
@@ -0,0 +1,34 @@
...
+
+    static Collection<TestSelection> getFailedTests(FailedConfiguration strategyConfig, Set<TestResult> testResults) {
+
+        return testResults
+            .stream()
+            .filter(TestResult::isFailing)
+            .map(testResult -> createTestSelection(strategyConfig, testResult))
+            .collect(
+                Collectors.toMap(TestSelection::getClassName, Function.identity(), TestSelection::merge,
+                    LinkedHashMap::new))
```
lordofthejars: Why are you creating a `Map` and then calling `values`method? Probably there is a reason but cannot see now.

strategies/categorized/pom.xml
```diff
@@ -31,6 +31,13 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.arquillian.smart.testing</groupId>
+      <artifactId>core</artifactId>
+      <version>${project.version}</version>
+      <classifier>tests</classifier>
+      <scope>test</scope>
+    </dependency>
```
dipak-pawar: Can you move this in `dependencyManagement` of parent pom.xml and use it like 
```xml
<dependency>
      <artifactId>core</artifactId>
      <classifier>tests</classifier>
      <scope>test</scope>
    </dependency>
```

strategies/failed/pom.xml
```diff
@@ -36,6 +36,13 @@
       <artifactId>junit-test-result-parser</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.arquillian.smart.testing</groupId>
+      <artifactId>core</artifactId>
+      <version>${project.version}</version>
+      <classifier>tests</classifier>
+      <scope>test</scope>
+    </dependency>
```
dipak-pawar: same as above

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/TestResultsFilter.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.arquillian.smart.testing.TestSelection;
+import org.arquillian.smart.testing.spi.TestResult;
+
+public class TestResultsFilter {
```
dipak-pawar: This can be package-private. We can make it public when needed

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/TestResultsFilter.java
MatousJobanek: the thing is that I have a list of TestSelections for the same class, but containing different test methods:
```java
TestSelection(TestClass.class, firstTestMethod, "failed")
TestSelection(TestClass.class, secondTestMethod, "failed")
TestSelection(AnotherTestClass.class, firstTestMethod, "failed")
TestSelection(AnotherTestClass.class, secondTestMethod, "failed")
```
So I need to group it according to the same test class and then merge it using the `TestSelection::merge` method. If there is some way to make it better without creating a map, then just tell it and I'll change it :-)

lordofthejars: ah right I see now what you mean, I think you can use `groupBy` which does not save from calling `.values()` but maybe it is more clear. http://www.baeldung.com/java-groupingby-collector

strategies/categorized/pom.xml
MatousJobanek: I'm not really a fan of having everything inside of the parent pom file - it's getting to be bigger and bigger. 
In addition, in the parent pom file, it is a part of the `dependencyManagement` section which takes care of the versions - to use the correct one. But in this case, there is nothing to manage - it is `${project.version}`. So putting 6 lines with another dependency inside of the parent pom file (which is quite long already) just because of saving 2 lines in these pom files (that is really not necessary) doesn't seem to be worth doing for me. Just my $0.02. WDYT @lordofthejars ?

strategies/failed/src/main/java/org/arquillian/smart/testing/strategies/failed/TestResultsFilter.java
MatousJobanek: good catch 👍

strategies/categorized/pom.xml
lordofthejars: @MatousJobanek I agree with you, but if you check other Arquillian projects (some of them not all) we are doing how @dipak-pawar mentions. For me both approaches are ok, what I would say is let's see how we are using in Smart Testing and apply the same logic.

MatousJobanek: > what I would say is let's see how we are using in Smart Testing and apply the same logic.

And this is exactly what I don't like and I wanted to start making more clear (mainly in this case of the jar with testing bits). Anyway, I'll move it there and create another issue to remove those dependencies from the parent pom file that don't make sense to have them there.

PR #319, opened 2018-02-16
fix(#318): the variable containing map of strategy configs is dumped as well

### LINKED ISSUE #318: Categorized strategy properties are not being recognized when using a configuration file
##### Issue Overview

Properties from configuration file are not getting mapped into the `categorized` strategy configuration.

##### Expected Behaviour

When running smart testing with the `categorized` strategy and a configuration file, smart testing should configure the `categorized` strategy from configuration file correctly. 

##### Current Behaviour

When running smart testing with `categorized` strategy and a configuration file, properties set in that file are not being recognized by the execution itself. According to the `report.xml` smart testing sees only, that the categorized strategy has been applied, but all the properties are ignored.
```xml
 <executionConfiguration>
        <usageMode>selecting</usageMode>
        <strategies>
            <strategy>categorized</strategy>
        </strategies>
        <properties>
            <property name="smart.testing" value="categorized"/>
        </properties>
    </executionConfiguration
```
Running smart testing with properties set as command line arguments are recognized properly.

##### Steps To Reproduce
1. Clone this repository https://github.com/spriadka/testsuite/tree/7.1.x-categories
2. Set up the machine according to the README
3. Create `smart-testing.yml` following:
``` yaml
strategies: categorized
debug: true
strategiesConfiguration:
  categorized:
    categories: Elytron
```
4. Execute following command 
```bash
mvn -Pbasic,standalone -Djboss.dist=<your_jboss_dist> -Darq.extension.webdriver.firefox_binary=<your_ff_binary_path> -Dtest="ElytronPolicyTestCase*" test
```
5. See the `basic/target/.smart-testing/report.xml`

MatousJobanek:

The variable containing a map of raw strategy configs is dumped into the precalculated config as well to make it available for later loading directly from the strategy implementations.

- Renamed `strategiesConfig` to `rawStrategyConfigurations`
- created getter to add the `rawStrategyConfigurations` variable to the dumped file
- fixed test-bed implementation to allow adding strategy configurations
- changed the logic that loads strategy configurations from the file - added a detection when the strategy config is not a map, but a list - eg. in this case:
```yaml
{autocorrect: false, customProviders: [], customStrategies: [], debug: false, disable: false,
  strategies: [categorized], strategiesConfiguration: [categorized: {
      caseSensitive: false, categories: [LoaderCategory, serviceCategory], matchAll: false,
      methods: false, reversed: false}]}
```

Fixes #318

### DIFF
--- a/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
+++ b/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
@@ -113,7 +113,7 @@ public class ConfigurationLoader {
                 configuration.setRawStrategyConfigurations((Map<String, Object>) strategiesConfiguration);
 
             } else if (strategiesConfiguration instanceof List) {
-                HashMap<String, Object> configMap = new HashMap<>();
+                Map<String, Object> configMap = new HashMap<>();
                 ((List) strategiesConfiguration).stream()
                     .filter(item -> item instanceof Map)
                     .forEach(config -> configMap.putAll((Map) config));
--- a/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
+++ b/core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
@@ -3,8 +3,9 @@ package org.arquillian.smart.testing.configuration;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
-import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import org.arquillian.smart.testing.hub.storage.local.LocalStorage;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/arquillian/smart/testing/configuration/ConfigurationLoader.java
```diff
@@ -108,7 +109,16 @@ private static Configuration loadAsConfiguration(Map<String, Object> yamlConfigu
 
         final Configuration configuration = mapToObject(Configuration.class, yamlConfiguration);
         if (strategiesConfiguration != null) {
-            configuration.setStrategiesConfig((Map<String, Object>) strategiesConfiguration);
+            if (strategiesConfiguration instanceof Map) {
+                configuration.setRawStrategyConfigurations((Map<String, Object>) strategiesConfiguration);
+
+            } else if (strategiesConfiguration instanceof List) {
+                HashMap<String, Object> configMap = new HashMap<>();
```
lordofthejars: `Map<String, String> configMap`

MatousJobanek: no, it's `HashMap<String, Object>` see the parameter in `Configuration` class

lordofthejars: Sorry I meant `Map<String, Object>`using the interface not the implementation.

MatousJobanek: Ok, yeah, sure :-)

PR #321, opened 2018-02-20
fix(#320): scans classes defined as annotations attributes

### LINKED ISSUE #320: In affected strategy classes as in annotation attribute is not resolved
##### Issue Overview

In affected strategy classes as in annotation, attribute is not resolved.

For example:

```java
@Import(SimpleImportsClass.class)
public class AnnotationAtClassLevelWithClassValue {
}
```

`SimpleImportsClass` class is not resolved.

##### Expected Behaviour

`SimpleImportsClass` needs to be resolved as well as dependency.

##### Current Behaviour

It is ignored.

lordofthejars:

Scans annotation attributes as well searching for classes.

- Adds a member visitor to find classes defined in annotation fields

Fixes #320

### DIFF
--- a/strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ast/JavaAssistClass.java
+++ b/strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ast/JavaAssistClass.java
@@ -163,71 +163,20 @@ public class JavaAssistClass extends AbstractJavaClass {
 
                 final Set<String> memberNames = each.getMemberNames();
 
-                for (String memberName : memberNames) {
-                    addSubtypes(imports, each, memberName);
+                if (memberNames != null) {
+                    for (String memberName : memberNames) {
+                        imports.addAll(addSubtypes(each, memberName));
+                    }
                 }
             }
         }
     }
 
-    private void addSubtypes(Collection<String> imports, Annotation each, String memberName) {
-        each.getMemberValue(memberName).accept(new MemberValueVisitor() {
-            @Override
-            public void visitAnnotationMemberValue(AnnotationMemberValue node) {
-            }
+    private Collection<String> addSubtypes(Annotation each, String memberName) {
+        final ClassExtractorAnnotationMemberValue classExtractorAnnotationMemberValue  = new ClassExtractorAnnotationMemberValue();
+        each.getMemberValue(memberName).accept(classExtractorAnnotationMemberValue);
 
-            @Override
-            public void visitArrayMemberValue(ArrayMemberValue node) {
-                final String clazz = node.getType().toString();
-                imports.add(clazz.substring(0, clazz.lastIndexOf(".class")));
-            }
-
-            @Override
-            public void visitBooleanMemberValue(BooleanMemberValue node) {
-            }
-
-            @Override
-            public void visitByteMemberValue(ByteMemberValue node) {
-            }
-
-            @Override
-            public void visitCharMemberValue(CharMemberValue node) {
-            }
-
-            @Override
-            public void visitDoubleMemberValue(DoubleMemberValue node) {
-            }
-
-            @Override
-            public void visitEnumMemberValue(EnumMemberValue node) {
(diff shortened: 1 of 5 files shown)


strategies/affected/src/main/java/org/arquillian/smart/testing/strategies/affected/ast/JavaAssistClass.java
```diff
@@ -146,10 +160,76 @@ private void addAnnotations(Collection<String> imports, AnnotationsAttribute ann
...
+
+                for (String memberName : memberNames) {
+                    addSubtypes(imports, each, memberName);
+                }
             }
         }
     }
 
+    private void addSubtypes(Collection<String> imports, Annotation each, String memberName) {
+        each.getMemberValue(memberName).accept(new MemberValueVisitor() {
```
bartoszmajsak: Can we extract this to its own class?

bartoszmajsak:
Build is failing... but... it's "only" [`NullPointerException`](https://travis-ci.org/arquillian/smart-testing/builds/343947864#L2553-L2554) :)

lordofthejars:
About functional test I feel confident that with a unit test is enough in this case since it is a corner case, really specific to one class with a clear signature. I think that for this concrete case unit test is the perfect test scope.

lordofthejars:
About NPE I need to sort out my head and start using some of the community time on Smart Testing and enable it in the project as well so before pushing instead of freezing the computer we just run the tests it makes sense.

bartoszmajsak:
> About functional test I feel confident that with a unit test is enough in this case since it is a corner case, really specific to one class with a clear signature. I think that for this concrete case unit test is the perfect test scope. Also because I think that ftest will increase an overhead that it is not worth for this scoped problem.

I agree to some extent, especially that this is now very optimistic check as we don't really know what particular annotation means. We might need some more real use case coverage once we go deeper in the understanding of what impact / relation between test and class in the annotation is.

bartoszmajsak:
> About NPE I need to sort out my head and start using some of the community time on Smart Testing and enable it in the project as well so before pushing instead of freezing the computer we just run the tests it makes sense.

Not sure if that's already a problem - doesn't take long on my machine, but I can understand it's heavyweight thing around functional tests. In this case it's only unit tests which failed, so maybe you can have a local pre-commit running only surefire tests instead? (or excluding ftest module). But sure, by all means, we can try smart test this project :)

Now we have ftest failing though :)

bartoszmajsak:
I think it might have sth to do with Travis. Works on my machine, but I noticed that build is taking way longer now compared to what I remember (13 minutes vs 7 minutes I recall from the past). We might want to review our functional tests to see if we can optimize something there. Ideally I would like to have the time cut by half :)

bartoszmajsak:
@lordofthejars it seems that Spacelift is failing on Travis for the first test. Do we have another concurrency problem? https://travis-ci.org/arquillian/smart-testing/builds/344038179#L2867

Any ideas @MatousJobanek?

bartoszmajsak:
One solution here would be to get Maven downloaded before the build and use it for all the tests:

https://github.com/arquillian/smart-testing/blob/dd0cd89e09192ab98cf7cfbfe9a5a5edebad48ef/.travis.yml#L38-L40
https://github.com/arquillian/smart-testing/blob/dd0cd89e09192ab98cf7cfbfe9a5a5edebad48ef/.travis.yml#L64-L65

MatousJobanek:
unfortunately, it seems that it is another concurrency problem - the marker file solution doesn't work for all cases obviously. I'm afraid that as it is being executed in separated java processes, it would be hard to solve this problem so there are no possible gaps when some concurrency problem could occur. 
With this, there is probably no other easy solution than bringing back the downloading of the Maven binaries before the build...

lordofthejars:
@bartoszmajsak I tried this PR in Spring Boot project and now it works :)

MatousJobanek:
another solution could be to introduce marker file for downloading the binaries as this relies on thread synchronization.

PR #327, opened 2018-02-22
fix(#326): releasing also test jars

### LINKED ISSUE #326: Release test jars
##### Issue Overview

Because of this line https://github.com/arquillian/smart-testing/blob/master/pom.xml#L261 Smart Testing is released without test jars. Because the `core` modules exposes test jar that is used in other modules, then the build of ST fails on the specific version (`0.0.9`) if it is built with a clean local repository.

##### Steps To Reproduce
1. checkout 0.0.9 tag
2. build it with a clean local repository (without installing ST there)

MatousJobanek:
Fixes #326

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -271,13 +271,7 @@
         <surefire.fork.count>2</surefire.fork.count>
         <surefire.thread.count>2</surefire.thread.count>
       </properties>
-    </profile>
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
Thank you @MatousJobanek! \o/

